PR #4, opened 2020-07-03
deftech expression

DanBurton:
Not sure why the change to manual-racket.js; I might be using a newer version of scribble or something.

The only real change here is adding a `@deftech{expression}` to fix the broken link


jeapostrophe:
Thanks, I originally had it that way, but I think that I want this section to link to 1.3 for expressions and maybe statements too. On the other hand... it is a bit gross to have the semantics and the syntax so linked, so I might use a different phrase than 'expression'.

PR #9, opened 2020-07-07
More warnings on inexhaustive patterns

DanBurton:
Code is untested. I did the dumb thing and inserted a bunch of `error` with `-- XXX`, so now the code is more explicit about when it can fail, at least. Future work is to use actual error monads or something of the sort.

I do want to highlight `CompiledSolRec` as these change highlight typical JSON parsing in Haskell.

I went with non-invasive changes for now, but future work also likely includes things like preferring Text over String, records over tuples, etc.

### DIFF
--- a/hs/package.yaml
+++ b/hs/package.yaml
@@ -44,6 +44,10 @@ dependencies:
 ghc-options:
 - -Wall
 - -Werror
+- -Wcompat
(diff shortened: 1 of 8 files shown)


DanBurton:
Passes `make clean build run` on the nim example. (caveat: I skipped emit_teal).

PR #11, opened 2020-07-24
hs code formatted w/ ormolu

DanBurton:
Not a real PR, just a demo of what the code looks like when run thru [`ormolu`](https://github.com/tweag/ormolu#readme).

Be warned that `stack build ormolu` takes a while to build because it uses `ghc-lib-parser` which is huge.

Ormolu has no config.

Overall it looks like the main thing it does compared to our style is it pushes more things onto a new line

* where bindings are pushed to the line after `where`
* multiline bindings are pushed to the line after `=`
* case branches are pushed to the line after `of`

Also

* it uses comma suffix style, rather than "Haskelly" comma prefix style.
* it lumps all imports together and alphabetizes

I have no strong feelings about any of these style choices, I just like to have an autoformatter that keeps the code looking consistent so that I don't have to think about it.

### DIFF
--- a/hs/src/Reach/VerifySMT.hs
+++ b/hs/src/Reach/VerifySMT.hs
@@ -1,29 +1,31 @@
 module Reach.VerifySMT where
+
 -- XXX clean up function names and params to not refer to z3
 
-import Data.Maybe (catMaybes)
-import qualified Data.Set as S
-import qualified Data.Map as M
-import Data.Char (isDigit)
-import Text.Read (readMaybe)
+--- Maybe use Language.SMTLib2 in future
+
+import Control.Loop
 import Control.Monad
 import Control.Monad.Extra
-import SimpleSMT hiding (not) --- Maybe use Language.SMTLib2 in future
-import System.IO
-import System.Exit
-import Data.Text.Prettyprint.Doc
-import Data.Digest.CRC32
 import qualified Data.ByteString.Char8 as BS
+import Data.Char (isDigit)
+import Data.Digest.CRC32
 import Data.IORef
-import Control.Loop
+import qualified Data.Map as M
+import Data.Maybe (catMaybes)
+import qualified Data.Set as S
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Text.IO as TIO
-
+import Data.Text.Prettyprint.Doc
 import Reach.AST
 import Reach.EmbeddedFiles
-import Reach.Util
 import Reach.Pretty ()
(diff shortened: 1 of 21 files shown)


hs/src/Reach/VerifySMT.hs
```diff
@@ -1,29 +1,31 @@
 module Reach.VerifySMT where
+
 -- XXX clean up function names and params to not refer to z3
 
-import Data.Maybe (catMaybes)
-import qualified Data.Set as S
-import qualified Data.Map as M
-import Data.Char (isDigit)
-import Text.Read (readMaybe)
+--- Maybe use Language.SMTLib2 in future
```
DanBurton, line 5: It doesn't let you comment on import statements; it moved this to the top of the import block

hs/test/Main.hs
```diff
@@ -185,19 +199,21 @@ goldenTests mkTest ext subdir = do
 
 main :: IO ()
 main = do
-  parseTests <- testsFor (Proxy @ParseErr) parseErrExample ".rsh" "parse-errors"
-  compileTests <- testsFor (Proxy @CompileErr) compileErrExample ".rsh" "compile-errors"
+  parseTests <- testsFor (Proxy :: Proxy ParseErr) parseErrExample ".rsh" "parse-errors"
+  compileTests <- testsFor (Proxy :: Proxy CompileErr) compileErrExample ".rsh" "compile-errors"
```
DanBurton, line 203: This was the only part of the code I had to change to accommodate ormolu. It apparently doesn't understand type application syntax using `@`.

hs/test/Main.hs
```diff
@@ -158,8 +167,13 @@ testExamplesCover p sources = testSpec label t where
...
 
-testsFor :: ReachErr err =>
-  proxy err -> (FilePath -> TestTree) -> String -> FilePath -> IO TestTree
+testsFor ::
+  ReachErr err =>
+  proxy err ->
+  (FilePath -> TestTree) ->
+  String ->
+  FilePath ->
+  IO TestTree
```
DanBurton, line 176: It splits long type signatures out onto one arg type per line

hs/src/Reach/VerifySMT.hs
jeapostrophe: Maybe if the comment were on the same line?

hs/test/Main.hs
jeapostrophe: Not so bad.

PR #12, opened 2020-07-24
NOMERGE sample output with cvc4

DanBurton:
Ran the whole test suite w/ cvc4. It's hitting "resource vanished" which suggests that something we're doing is killing it. But it does get thru some verifications just fine.


jeapostrophe:
One theory of crashing might be that CVC4 doesn't support the push/pop-multiple checking thing that we do? I don't know.

PR #13, opened 2020-07-25
hs - autoformatted w/ leading commas

DanBurton:
Found a fork that does leading commas


jeapostrophe:
Alright. Let's do it.

PR #25, opened 2020-11-19
Make "Incompatible States" Error Message More Readable

chrisnevers:
## Ticket(s)
[Make "incompatible states" error message more human readable](https://trello.com/c/eu4QIRZE)

## Purpose
Updated the error message for incompatible states to list off the issues between the actual and expected state of the application in a human readable way.


hs/test-examples/nl-eval-errors/Err_Eval_IncompatibleStates.txt
```diff
@@ -1,3 +1,4 @@
 reachc: error: ./Err_Eval_IncompatibleStates.rsh:6:28:after expr stmt semicolon: incompatible states:
-  After First Message: False vs True
-  Participant Definitions: fromList [] vs fromList [("A",DLVar ./Err_Eval_IncompatibleStates.rsh:5:6:dot "A" Address 0)]
+Issues:
+  * Expected a publication to have been made by this point.
+  * Expected "A" to be a participant but there are none.
```
jeapostrophe, line 4: I will be easy to modify this to be what we want, but I think I need to explain what these `pdvs` even are...

A participant appears in `pdvs` if it has been either `Participant.set()` or if it has `publish`ed. So, a good way to "explain" this is to say something like that, "Expected 'A' to have published a message or been set, but it hasn't."

hs/src/Reach/Eval.hs
```diff
@@ -159,31 +159,53 @@ didYouMean invalidStr validOptions maxClosest = case validOptions of
...
+  1 -> sing
+  _ -> plur
 
 showStateDiff :: SLState -> SLState -> String
 showStateDiff x y =
-  showDiff x y st_mode "Mode" showVS
-    <> showDiff x y st_live "Live" showVS
-    <> showDiff x y st_after_first "After First Message" showVS
-    <> showDiff x y st_pdvs "Participant Definitions" showVS
+  "\nIssues:"
```
jeapostrophe, line 177: I'm not in love with "issues"... I want to say something like "Program branch effects incompatible, because:", but I feel like that's too "programmy"

jeapostrophe (changes requested):

hs/src/Reach/Eval.hs
```diff
@@ -159,31 +159,53 @@ didYouMean invalidStr validOptions maxClosest = case validOptions of
...
+  <> showDiff x y st_mode (\ xMode yMode ->
+    unwords ["Expected to be in ", show yMode, ", but in ", show xMode <> "."])
+  <> showDiff x y st_live (\ xLive yLive ->
+      case (xLive, yLive) of
+        (False, True) -> "Expected there to be live state."
+        (True, False) -> "Expected there to be no live state."
+        _             -> fail "Compiler error"
```
DanBurton, line 184:
```
λ> fail "Compiler error" :: String
""
```

This is probably not what you meant. Use `expect_throw` instead.

hs/src/Reach/Eval.hs
```diff
@@ -159,31 +159,53 @@ didYouMean invalidStr validOptions maxClosest = case validOptions of
...
+    )
+  <> showDiff x y st_after_first (\ xAfter yAfter ->
+      case (xAfter, yAfter) of
+        (False, True) -> "Expected a publication to have been made by this point."
+        (True, False) -> "Expected no publication to have been made by this point."
+        _             -> fail "Compiler error"
```
DanBurton, line 190: Also here

hs/test-examples/nl-eval-errors/Err_Eval_IncompatibleStates.txt
chrisnevers: Ah okay, thanks. Is it correct to say that the participant has to have an associated address to be in `pdvs`? In this example, `A` has `publish`ed before `transfer`ing and it's not in `pdvs`

hs/src/Reach/Eval.hs
jeapostrophe: `impossible`

hs/test-examples/nl-eval-errors/Err_Eval_IncompatibleStates.txt
jeapostrophe: Yes, being in `pdvs` /means/ the participant has an address (i.e. has acted publicly or been assigned)

hs/src/Reach/Eval.hs
chrisnevers: `The expected state of the program varies between branches, because`?

jeapostrophe: Sure

jeapostrophe (approved):

PR #26, opened 2020-11-20
Add Lookup Context To `env_lookup`

chrisnevers:
# Ticket(s)
[Add some more context to env_lookup to see why it is being looked up](https://trello.com/c/uhxLtO7y)

# Purpose
* Add `LookupCtx` datatype
* Update `env_lookup` and `Err_Eval_UnboundId` to accept `LookupCtx` param.

There are two use cases for `env_lookup`:
* When a user wants to reference an id. e.g. `switch (y) { ... }`, `3 + a`.
* When the compiler wants to ensure an id exists. e.g. `main`. In the future, we might only care about a default export and this won't be needed.

I looked into abstracting object property lookups into this function, but I think it would be too much overhaul compared to the added benefits.

### DIFF
--- a/examples/abstract-simul/index.txt
+++ b/examples/abstract-simul/index.txt
@@ -1,1 +1,1 @@
-reachc: error: <top level>: Invalid unbound identifier: main. Did you mean: ["rps","rental"]
+reachc: error: <top level>: Expected the following identifier to be declared: "main"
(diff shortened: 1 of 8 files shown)


jeapostrophe (approved):
Can you make a new card for going back for the object stuff at some point in the future?

chrisnevers:
@jeapostrophe will do

PR #28, opened 2020-11-27
Match expression

chrisnevers:
## Ticket(s)
[switch expression form](https://trello.com/c/xmEkJTwE)

## Purpose
Add a `match` property on `Data` objects. It will essentially desugar
```javascript
// mx :: Maybe(UInt)
mx.match({
  None: ( ) => 0,
  Some: (i) => i,
})
```
to
```javascript
((data_id, cases_id) => {
  switch (data_id) {
    None: return cases_id.None(),
    Some: return cases_id.Some(data_id),
})(mx, { None: () => 0, Some: (i) => i })
```
All the evaluation and exhaustiveness checks are done by the existing switch statement logic.


hs/src/Reach/AST.hs
```diff
@@ -240,6 +240,21 @@ data SLVal
...
+  SLV_Bytes a _ -> Just a
+  SLV_Array a _ _ -> Just a
+  SLV_Tuple a _ -> Just a
+  SLV_Object a _ _ -> Just a
+  SLV_Clo a _ _ _ _ -> Just a
+  SLV_Data a _ _ _ -> Just a
+  SLV_DLVar (DLVar a _ _ _) -> Just a
+  SLV_Participant a _ _ _ -> Just a
+  _ -> Nothing
+
```
jeapostrophe, line 257: I actually recently added a type class for this: `SrcLocOf`

jeapostrophe (changes requested):
Try a hand at writing documentation, or would you like me to do it?

hs/src/Reach/AST.hs
chrisnevers: I saw that, but was concerned about the type signature since a lot of `SLVal`s don't have `SrcLoc`. What are your thoughts about extending the signature to include a default param? I could leave it to the `SLVal` implementation to return an empty position, but I could provide a 'closer' location as a default. 

Also, I'll try adding some documentation for `match`

jeapostrophe: I think it would be to return `srcloc_builtin` and check for it in this call site... even though that's generally bad, I think it's okay in this case.

jeapostrophe (approved):

PR #31, opened 2020-12-07
Have `connectAccount` Remember Which Stdlib it Came From

chrisnevers:
Prevent users from passing different standard libraries (than the one they used to connect the account) to the backend. 

## Breaking Changes

* The `stdlib` argument is no longer needed for backend initialization.
    ```javascript
    backend.Alice(stdlib, ctc, {});
    ```
    is now
    ```javascript
    backend.Alice(ctc, {});
    ```

* Standard libraries must now be treated as default imports.
    ```javascript
    import * as ETH_stdlib from '@reach-sh/stdlib/ETH.mjs'
    ```
    is now
    ```javascript
    import ETH_stdlib from '@reach-sh/stdlib/ETH.mjs'
    ```
    This change is needed because each standard library module is now self referential, so the standard library is now an object that is exported by default, with a `this` getter. Now, instead of `export function`, new stdlib functions will be exported like this:
    ```javascript
    function x () {}
    
    stdlib = {
        ...,
        // Add it to list of exports at end of file
        x 
    };
    export default stdlib;
    ```


js/ts/ALGO.ts
```diff
@@ -621,7 +621,8 @@ const doQuery = async (dhead:string, query: ApiCall<any>): Promise<any> => {
   return txn;
 };
 
-export const connectAccount = async (networkAccount: NetworkAccount) => {
+const connectAccount = async (networkAccount: NetworkAccount) => {
+  const stdlibT: any = stdlib;
```
DanBurton, line 625: What does the T suffix mean?

examples/tut-7/build/index.main.mjs
```diff
@@ -1,3514 +0,0 @@
-// Automatically generated with Reach 0.1.2
```
jeapostrophe, line 1: Should not be deleted

examples/tut-7/build/index.main.sol
```diff
@@ -1,271 +0,0 @@
-// Automatically generated with Reach 0.1.2
```
jeapostrophe, line 1: Should not be deleted

js/ts/ALGO.ts
chrisnevers: To avoid the same name, I just appended a T for this. Which is probably a pretty poor choice. Open to name suggestions

jeapostrophe (changes requested):

PR #37, opened 2020-12-31
Extend Standard Library

chrisnevers:
* Add an `Either` data type and some functions to the standard library. 
    * Either
      * either
      * isLeft
      * isRight
      * fromLeft
      * fromRight
    * Array
      * min
      * max
      * all
      * any
      * or
      * and
      * sum
      * includes
      * indexOf
      * findIndex
      * count
    * compose (function composition)
* Update `evalAsEnv` to handle `Array/Data/Tuple` that come from the interaction interface. 
* Fix an issue with the optimizer where it wasn't optimizing the variable that `switch` destructures.

I had some difficulty finding things that were useful to the language as it is, so if I missed anything let me know.


jeapostrophe (approved):

jeapostrophe:
This is great. It might be nice to have a few test cases, but I'm not really doubting that it works :)

PR #38, opened 2021-01-06
hs: update language-javascript

DanBurton:


DanBurton:
There's currently something wrong with the way I turn `JSConciseBody` into `JSStatement` and/or `[JSStatement]` somewhere. It is causing a TypeMeets_Mismatch on `nim/index-abstract.rsh`, and some sort of impossible situation where it can't find an identifier on `timeoutception/index.txt`.

hs/src/Reach/JSUtil.hs
```diff
@@ -55,6 +58,21 @@ jsa_flatten a = concatMap f a
     f (JSArrayComma _) = []
     f (JSArrayElement e) = [e]
 
+jsConciseBodyToBlock :: JSConciseBody -> JSBlock
+jsConciseBodyToBlock = \case
+  JSConciseFunctionBody b -> b
+  JSConciseExpressionBody e -> JSBlock JSNoAnnot [JSReturn JSNoAnnot (Just e) JSSemiAuto] JSNoAnnot
+
+jsConciseBodyToStmts :: JSConciseBody -> [JSStatement]
+jsConciseBodyToStmts = \case
```
jeapostrophe, line 67: I think I'd make this use `jsConciseBodyToBlock`, then pull them out

hs/src/Reach/JSUtil.hs
```diff
@@ -55,6 +58,21 @@ jsa_flatten a = concatMap f a
...
+  JSConciseFunctionBody (JSBlock _ stmts _) -> stmts
+  JSConciseExpressionBody e -> [JSReturn JSNoAnnot (Just e) JSSemiAuto]
+
+jsConciseBodyToStmt :: JSConciseBody -> JSStatement
+jsConciseBodyToStmt = \case
+  JSConciseFunctionBody (JSBlock l stmts r) -> JSStatementBlock l stmts r JSSemiAuto
+  JSConciseExpressionBody e -> JSStatementBlock JSNoAnnot [JSReturn JSNoAnnot (Just e) JSSemiAuto] JSNoAnnot JSSemiAuto
```
jeapostrophe, line 74: Ditto

jeapostrophe (commented):
My guess is that the `JSMethodCall` case isn't behaving the same way with these new `JSConciseBody` things. It is important for it *not* to be a return when the RHS is a method call.

PR #40, opened 2021-01-14
sltype, typechecks and is maybe not wrong anymore

DanBurton:
dt2st is always safe

st2dt crashes if you give it something that is not a dt

All failing test cases are currently caused by st2dt crashing

Changed `DLType` to `SLType` in the following cases:

* funFold
* primOpType
* isFirstOrder
* SLVal constructors
* Some Eval Err_* constructors
* checkAndConvert & friends
* TypeEnv
* typeSubst
* InteractEnv

There may be something weird going on with:

* CollectsTypes
* Type.hs: checkAndConvert: typeSubst is now unused; that's definitely wrong
* SMT.hs: defineIE

### DIFF
--- a/hs/src/Reach/AST/Base.hs
+++ b/hs/src/Reach/AST/Base.hs
@@ -166,19 +166,10 @@ instance Monoid SecurityLevel where
 --- Static Language
 type SLVar = String
 
-data DLType
-  = T_Null
-  | T_Bool
-  | T_UInt
-  | T_Bytes Integer
-  | T_Digest
-  | T_Address
-  | T_Array DLType Integer
-  | T_Tuple [DLType]
-  | T_Object (M.Map SLVar DLType)
-  | T_Data (M.Map SLVar DLType)
-  deriving (Eq, Generic, NFData, Ord)
-
+-- SL types are a superset of DL types.
+-- We copy/paste constructors instead of using `ST_Val DLType`
+--  because some things can exist in SL that do not exist in DL,
+--  such as an object whose fields are functions.
 data SLType
   = ST_Null
   | ST_Bool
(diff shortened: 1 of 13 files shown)


hs/src/Reach/Type.hs
```diff
@@ -247,29 +249,30 @@ typeChecks tint@(mcfs, _) at env ts vs =
...
+      dargs <- typeChecks tint at env (map st2dt dom) args
+      return (st2dt rng, dargs)
+    ST_Forall var ft -> do
       var_ref <- newSTRef Nothing
       let env' = M.insert var var_ref env
-      (vrng, dargs) <- checkAndConvert_i tint at env' ft args
-      rng <- typeSubst at env' vrng
+      (rng, dargs) <- checkAndConvert_i tint at env' ft args
+      -- XXX happens prior to this point now?
+      -- rng <- typeSubst at env' vrng
```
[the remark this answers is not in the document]
jeapostrophe: So, if the types preclude that... then this suggests there might be another place where a `st2dt` would error because of the type variable not having been substituted

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1050,12 +1050,12 @@ _verify_smt mc vst smt lp = do
...
   let defineIE who (v, it) =
         case it of
-          T_Fun {} -> mempty
+          -- ST_Fun {} -> mempty
           _ ->
             pathAddUnbound_v ctxt Nothing at (smtInteract ctxt who v) it O_Interact
   let definePIE (who, InteractEnv iem) = do
         pathAddUnbound_v ctxt Nothing at (smtAddress who) T_Address O_BuiltIn
-        mapM_ (defineIE who) $ M.toList iem
+        mapM_ (defineIE who) $ M.toList $ M.map st2dt iem
```
jeapostrophe, line 1058: This is what I was mentioning before... no `SLType` should be exposed past `Eval.hs`

hs/src/Reach/AST/Base.hs
```diff
@@ -219,36 +223,37 @@ hasFun = funFold z k fun
...
+  show ST_Address = "Address"
+  show (ST_Array ty i) = "Array(" <> show ty <> ", " <> show i <> ")"
+  show (ST_Tuple tys) = "Tuple(" <> showTys tys <> ")"
+  show (ST_Object tyMap) = "Object({" <> showTyMap tyMap <> "})"
+  show (ST_Data tyMap) = "Object({" <> showTyMap tyMap <> "})"
+  show (ST_Fun tys ty) = "Fun([" <> showTys tys <> "], " <> show ty <> ")"
+  show (ST_Forall x t) = "Forall(" <> show x <> ", " <> show t <> ")"
+  show (ST_Var x) = show x
+  show (ST_Type ty) = "Type(" <> show ty <> ")"
+
```
jeapostrophe, line 256: I believe that all of the `SLType` stuff can go in `SL.hs`

hs/src/Reach/AST/DLBase.hs
```diff
@@ -45,6 +69,45 @@ data DLVar = DLVar SrcLoc String DLType Int
...
+  ST_Address -> T_Address
+  ST_Array ty i -> T_Array (st2dt ty) i
+  ST_Tuple tys -> T_Tuple (map st2dt tys)
+  ST_Object tyMap -> T_Object (M.map st2dt tyMap)
+  ST_Data tyMap -> T_Data (M.map st2dt tyMap)
+  -- XXX consider using Maybe so that callers have to handle the error case
+  t@(ST_Fun {}) -> error $ "ST_Fun not a dt: " <> show t
+  t@(ST_Forall {}) -> error $ "ST_Forall not a dt: " <> show t
+  t@(ST_Var {}) -> error $ "ST_Var not a dt: " <> show t
+  t@(ST_Type {}) -> error $ "ST_Type not a dt: " <> show t
```
jeapostrophe, line 90: I try not to use `error`, because I want us to think that `impossible` means that some earlier part of the compiler has not done its job to enforce some invariant. In this case, I believe (?) that `Type.hs` is responsible for this kind of logic and either it would enforce it (so this is `impossible`) or it is a consumer of this, so it needs it to be a `Maybe` so that it can blame the user

DanBurton:
a version of this has been merged

PR #41, opened 2021-01-18
0944 / Tutorial 6: client-py

mattaudesse:
Some remaining things that could be improved:
- ~Replace `sleep` on [line 45](https://github.com/reach-sh/reach-lang/compare/master...mattaudesse:rpc-prime?expand=1#diff-4b2c1ce062487372814816ee188003e7a63df48fd088e6bab9dc94ccc137d467R45) with something more robust (ideally from a drop-in package and not handcrafted)~
- Exceptions raised in child Alice/Bob threads currently don't bubble up immediately or reliably
- ~The selection of `client-py` (as opposed to `client-js`, or `client-N`) in `docker-compose.yml` is currently hard-coded - how do we want this to work?~
- ~Bob seems to win way too often when I run this... perhaps there's a better source of randomness than `random.randint`~
- I'm not sure if I think `requests` ~and `concurrent.futures`~ are the best choices here (although they're probably the most obvious answers a Python programmer would offer if asked)


examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,108 @@
+import json
+import os
+import random
+import requests
+from concurrent.futures import ALL_COMPLETED, ThreadPoolExecutor, wait
+from time               import sleep
+
+
+def log(t):
+    print(t, flush=True)
```
jeapostrophe, line 10: Just use `print`... I don't want to have people think that logging fancy is needed in the example

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def mk_rpc(proto, host, port):
+    def rpc(m, *args):
```
jeapostrophe, line 14: wait goes here

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,108 @@
...
+
+    return rpc, rpc_callbacks
+
+
+def main():
+    log('I am the client')
+    sleep(3)  # TODO wait until server becomes available to fulfill requests
+    host = os.environ['REACH_RPC_SERVER']
+    port = os.environ['REACH_RPC_PORT']
+
```
jeapostrophe, line 48: Can you move this into `mk_rpc`... maybe use them if no args are provided?

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,108 @@
...
+    rpc, rpc_callbacks = mk_rpc('http', host, port)
+
+    starting_balance = rpc('/stdlib/parseCurrency', 10)
+    acc_alice        = rpc('/stdlib/newTestAccount', starting_balance)
+    acc_bob          = rpc('/stdlib/newTestAccount', starting_balance)
+
+    fmt          = lambda x: rpc('/stdlib/formatCurrency', x, 4)
+    get_balance  = lambda w: fmt(rpc('/stdlib/balanceOf', w))
```
jeapostrophe, line 56: Would "normal Python people" write these as functions some other way?

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,108 @@
...
+
+    acceptWager = lambda amt: log('Bob accepts the wager of %s' % fmt(amt))
+
+    player = lambda who: \
+        {'stdlib.hasRandom': True,
+         'getHand':       lambda:   getHand(who),
+         'informTimeout': lambda:   log('%s observed a timeout' % who),
+         'seeOutcome':    lambda n: log('%s saw outcome %s'
+            % (who, OUTCOME[rpc('/stdlib/bigNumberToNumber', n)])) # noqa
+         }
```
jeapostrophe, line 80: Ditto with those

jeapostrophe (changes requested):
Re: Randomness --- Don't worry about it

Re: Hardcode in compose --- Don't worry about it

examples/tut-6-rpc/client-py/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM       python:3.8-alpine
+
+WORKDIR    /app
+
+# Install and cache `pip` dependencies
+RUN        python3.8 -m venv venv
+COPY       pip-deps.txt /app
+RUN        source /app/venv/bin/activate \
+             && pip install --upgrade pip \
+             && pip install -r pip-deps.txt
```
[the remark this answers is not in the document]
mattaudesse: Sorry, I don't understand - include what directly?

examples/tut-6-rpc/client-py/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM       python:3.8-alpine
+
+WORKDIR    /app
+
+# Install and cache `pip` dependencies
+RUN        python3.8 -m venv venv
```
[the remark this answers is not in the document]
mattaudesse: `venv` is less about Python version (although it sandboxes that too) and more about keeping third-party packages installed via `pip` isolated from the host system's package manager-installed versions. It's generally considered good practice to sandbox these and completely bypass dependency hell.

If you look at `pip-deps.txt` you'll see that versions are also pinned there, whereas they wouldn't be on the host system unless we did something else special.

For our containerized purposes some people might say "it's Docker, so it doesn't matter"... I figure it's easy to do it the clean way so that's generally my default. Would you rather ditch using `venv`s?

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,108 @@
+import json
+import os
+import random
+import requests
+from concurrent.futures import ALL_COMPLETED, ThreadPoolExecutor, wait
```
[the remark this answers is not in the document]
mattaudesse: I'll play with it and see if I can make it look less gross but `threading` is even clunkier in some ways.

examples/tut-6-rpc/client-py/index.py
mattaudesse: `flush=True` is necessary unless you want all output from the Python client to be buffered and spat to console in one go, unfortunately. Raw `print` also prevents interleaving output with the server's logging.

examples/tut-6-rpc/client-py/index.py
mattaudesse: This "isn't Pythonic", no... They want you to use several lines of the form:
```python
def fmt(x):
    return rpc('/stdlib/formatCurrency', x, 4)

# "PEP8" has different rules about how many lines must separate function definitions based on scope

def whatever(foo):
    pass
```

Better to appeal to familiarity than concision?

examples/tut-6-rpc/client-py/index.py
mattaudesse: Oh, maybe I misunderstood you - are you asking me to replace `log` with `print` + inline the extra `flush=True` everywhere?

examples/tut-6-rpc/client-py/Dockerfile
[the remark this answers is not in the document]
mattaudesse: I'm guessing my changes in 67e372a are what we're talking about - please correct me if not.

examples/tut-6-rpc/client-py/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY       pip-deps.txt /app
+RUN        source /app/venv/bin/activate \
+             && pip install --upgrade pip \
+             && pip install -r pip-deps.txt
+
+COPY       tox.ini index.py /app/
+RUN        /app/venv/bin/flake8 --config /app/tox.ini /app/index.py
```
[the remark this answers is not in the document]
mattaudesse: Dropped in 67e372a.

examples/tut-6-rpc/client-py/Dockerfile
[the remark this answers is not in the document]
jeapostrophe: Yes <3

examples/tut-6-rpc/client-py/Dockerfile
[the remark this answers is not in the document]
jeapostrophe: It's fine

examples/tut-6-rpc/client-py/index.py
jeapostrophe: I'm just saying to just replace `s/log/print` and don't worry about the downsides... because the point of this code is to be beautiful for someone to say, "Wow, look how simple using Reach is from Python". Every little bit of friction counts

examples/tut-6-rpc/client-py/index.py
jeapostrophe: Yea, I think we should go with familiarity. I don't want Python people to think they NEED to use `lambda`. Similarly, is it the case that the callbacks would "just work" on an object?

examples/tut-6-rpc/client-py/index.py
mattaudesse: Moved in a98cd5e

examples/tut-6-rpc/client-py/index.py
mattaudesse: De-logged in 3cd5ef1.

examples/tut-6-rpc/client-py/index.py
mattaudesse: Done in 2d9ee6d.

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,108 @@
...
+from time               import sleep
```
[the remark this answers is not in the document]
mattaudesse: Done in 2d9ee6d.

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,130 @@
...
+                                   'on host {} to start accepting connections.'
+                                   .format(port, host)) from ex
+
+
+def mk_rpc(proto='http',
+           host=os.environ['REACH_RPC_SERVER'],
+           port=os.environ['REACH_RPC_PORT']):
+
+    def rpc(m, *args):
+        wait_for_port(port, host)
```
jeapostrophe, line 39: I'd put think when `mk_rpc` is called so it only happens once. We'll assume that once it works once, all subsequent failures should be exposed

examples/tut-6-rpc/client-py/index.py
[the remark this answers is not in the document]
jeapostrophe: The thing that looked appealing about it to me was that it is just `import threading` and then `threading.Thread(blah).start` then `.join`. Yours is intimidating to my eye.

examples/tut-6-rpc/client-py/index.py
jeapostrophe: What I mean by an object is... see the next comment

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,130 @@
...
+         'getHand':       lambda:   getHand(who),
+         'informTimeout': lambda:   print('%s observed a timeout' % who),
+         'seeOutcome':    lambda n: print('%s saw outcome %s'
+            % (who, OUTCOME[rpc('/stdlib/bigNumberToNumber', n)]))
+         }
+
+    play_alice = lambda: rpc_callbacks(
+        '/backend/Alice',
+        ctc_alice,
+        dict(wager=rpc('/stdlib/parseCurrency', 5), **player('Alice')))
```
jeapostrophe, line 107: Right here you are passing a dictionary to `rpc_callback`. Would it more "Pythonic" to pass an object where the interaction interface methods are methods of the object? Is the code that you already have in `rpc_callbacks` compatible with it being an object?

examples/tut-6-rpc/client-py/index.py
```diff
@@ -0,0 +1,130 @@
...
+    play_bob = lambda: rpc_callbacks(
+        '/backend/Bob',
+        ctc_bob,
+        dict(acceptWager=acceptWager, **player('Bob')))
+
+    with ThreadPoolExecutor() as ex:
+        t_alice = ex.submit(play_alice)
+        t_bob   = ex.submit(play_bob)
+
+        wait([t_alice, t_bob], return_when=ALL_COMPLETED)
```
jeapostrophe, line 118: I am not an average Python programmer, but when I read this, I imagine that an average Python programmer thinks they need to understand more than

```
alice = threading.Thread(target=play_alice)
bob = threading.Thread(target=play_bob)
alice.start()
bob.start()
alice.join()
bob.join()
```

It would be nice if you could put the `start` in the constructor... but maybe I am trying to over-optimize when I really don't know what Pythonistas expect

jeapostrophe (changes requested):

examples/tut-6-rpc/client-py/index.py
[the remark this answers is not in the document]
mattaudesse: Replaced in 9ab5bef.

examples/tut-6-rpc/client-py/index.py
mattaudesse: The `with` pattern has special significance and you see it used + promoted all over the place in Python but concurrency/parallelism have always been very weak points, so most people avoid the standard library's answer to these (using 3rd party packages instead) to build anything beyond small stuff (which is why it's hard to choose one standard library API as opposed to another).

I swapped out the threading code in 9ab5bef to match yours, and maybe that's easier for beginners to read anyway, but unfortunately you can't chain `start` or put it in the constructor; their API forces you to call methods of the `Thread` instances separately.

examples/tut-6-rpc/client-py/index.py
mattaudesse: Moved `wait_for_port` in bbeabd7.

examples/tut-6-rpc/client-py/index.py
mattaudesse: I'll leave this unresolved in case you'd still like to consider the object route (Slack convo provides more detail), but I've replaced the `lambda`s in 4183771.

jeapostrophe (approved):

PR #42, opened 2021-01-19
Reformat SMT Errors as Javascript

chrisnevers:
Reformat the SMT errors as Javascript by:
* Additionally storing `DLExpr` in binding environment.
* When there's a failure: expand the expression & count how many times each variable is used
* For each var:
    * If var usage > 1: add an assignment statement to the theorem formalization
    * If var usage = 1: inline the variable's value into the expression
* The inlining is done during printing as the DL language does not handle recursive expressions


examples/overview/index-error.txt
```diff
@@ -7,59 +7,13 @@ Verification failed:
...
+  // Violation witness
+  const interact_Alice_request = 1;
+  //    ^ from interaction at ./index-error.rsh:4:12:application
 
-  This could be violated if...
-    "interact_Alice_request" = 1
-      (bound at: ./index-error.rsh:4:12:application)
-      (because: interaction)
+  // Theorem formalization
+  assert(0 == ((((0 + 0) + v4) + 0) - (interact_Alice_request - 1)));
```
jeapostrophe, line 15: This `v4` didn't get substituted

hs/src/Reach/Verify/SMT.hs
```diff
@@ -311,6 +317,130 @@ seVars se =
 set_to_seq :: S.Set a -> Seq.Seq a
 set_to_seq = Seq.fromList . S.toList
 
+type BindingEnv = M.Map String (Maybe DLVar, SrcLoc, BindingOrigin, Maybe SExpr, Maybe DLExpr)
+
+-- Log every occurence of dlvars so we know what is used how many times
+dlvOccurs :: [Int] -> BindingEnv -> DLExpr -> [Int]
```
jeapostrophe, line 323: This function might be what `Counts.hs` is about

jeapostrophe (changes requested):
I believe that the documentation for overview and the tutorial show the error message. They should be updated to use the correct text range and explain it. I believe that now it says something like "There's a complicated message", but now it should not be so scary to really explain it.

hs/src/Reach/Verify/SMT.hs
chrisnevers: After looking at it, I don't think it's appropriate. After counting a variable, I need to expand it into the value it's associated with in `bindings` and process that value as well.

jeapostrophe: Ok!

jeapostrophe (approved):
This is good to go. If you want, I can update the docs, or you can. Just let me know

jeapostrophe:
Oh, duh, you got that while I was typing :)

jeapostrophe:
Here's one las thing... would it be easy for you to change it so that this:

```
  const v38 = (v3 + (4 - ((v3 + 1) % 3))) % 3;
  const v51 = (v38 == 0) ? [0, 1] : ((v38 == 1) ? [1, 1] : [2, 0]);
  assert(0 == (((interact_Alice_wager + interact_Alice_wager) - (v51[0] * interact_Alice_wager)) - (v51[1] * interact_Alice_wager)));
```

was:

```
  const v38 = (v3 + (4 - ((v3 + 1) % 3))) % 3;
 //      ^ would be 0
  const v51 = (v38 == 0) ? [0, 1] : ((v38 == 1) ? [1, 1] : [2, 0]);
 //       ^ would be [0, 1]
  assert(0 == (((interact_Alice_wager + interact_Alice_wager) - (v51[0] * interact_Alice_wager)) - (v51[1] * interact_Alice_wager)));
```

You should be able to get these "would be" values by querying the model.

PR #44, opened 2021-01-21
0944 / Tutorial 6: client-go

mattaudesse:


examples/tut-6-rpc/client-go/index.go
```diff
@@ -0,0 +1,214 @@
...
+
+
+func mkRpc() (func(string, ...interface{}) interface{},
+              func(string, contract, map[string]interface{})) {
+
+  proto := "http"
+  host  := os.Getenv("REACH_RPC_SERVER")
+  port  := os.Getenv("REACH_RPC_PORT")
+
+  // TODO wait for server to become available
```
jeapostrophe, line 44: It is very annoying that we always need this. I wish there were a good way to build it into Docker

examples/tut-6-rpc/client-go/index.go
```diff
@@ -0,0 +1,214 @@
...
+
+func dieIf(err error) {
+  if err != nil {
+    log.Fatal(err)
+    os.Exit(1)
+  }
+}
+
+
+func please(a interface{}, err error) interface{} {
```
jeapostrophe, line 31: What's the story with this?

examples/tut-6-rpc/client-go/index.go
```diff
@@ -0,0 +1,214 @@
...
+
+          ans = nil
+          for _, a := range cb.Call(vargs) {
+            ans = a.Interface()
+            break
+          }
+
+          p = rpc("/kont", p["kid"], ans).(map[string]interface{})
+
+        // Respond with simple field value
```
jeapostrophe, line 111: I don't understand this... weren't these things transmitted in `vals`? What error did you have before this... it seems like a problem somewhere else

examples/tut-6-rpc/client-go/index.go
```diff
@@ -0,0 +1,214 @@
...
+      fmt.Printf("%s played %s\n", who, HAND[hand])
+      return hand
+    }
+
+    informTimeout := func() {
+      fmt.Printf("%s observed a timeout\n", who)
+    }
+
+    seeOutcome := func(n bigNumber) {
+      o := int(rpc("/stdlib/bigNumberToNumber", n).(float64))
```
jeapostrophe, line 168: Why `float64`?

mattaudesse: I'm guessing the type is inferred somewhere in the HTTP library's guts - there's a runtime exception if I don't capture it this way and cast to `int`.

examples/tut-6-rpc/client-go/index.go
mattaudesse: It's apparently commonplace to repeat the code you see in `please` and `dieIf` everywhere it's possible to receive a failure condition (so... constantly), so I made these helpers and was able to clean things up considerably before whittling down+consolidating things to the point where they are now.

I figured you'd ask. Want me to inline them instead?

examples/tut-6-rpc/client-go/index.go
mattaudesse: Would you rather me investigate a generic Docker solution than redo it in each client language?

examples/tut-6-rpc/client-go/index.go
mattaudesse: This is one of the things I was alluding to when I said I was still poking at it - I think you're right and I can remove this part.

Trying to explain why it's here wouldn't make much sense without context that's stale now :-)

examples/tut-6-rpc/client-go/index.go
jeapostrophe: Naw, it's fine to do it in each one, because when deployed, they won't necessarily be in Docker

examples/tut-6-rpc/client-go/index.go
jeapostrophe: Naw... that's nuts though.

examples/tut-6-rpc/client-go/index.go
jeapostrophe: :)

examples/tut-6-rpc/client-go/index.go
jeapostrophe: Ah

examples/tut-6-rpc/client-go/index.go
mattaudesse: Rejiggered in d2164e4.

jeapostrophe:
Beautiful, I think we just need the `waitPort`

mattaudesse:
> Beautiful, I think we just need the `waitPort`

I've made an update in 6d0c6b1 that's intended to match the Python client's behavior.

jeapostrophe:
Awesome... please rebase and merge <3 and have a good weekend!

PR #45, opened 2021-01-26
reach: use reach-network for docker stuff

DanBurton:
See: https://trello.com/c/SjbwvzNU/879-improve-our-docker-network-hygiene

This is a fairly simple change, and it does the thing it's supposed to do.

However, there's an awkward catch, which is that sometimes if you've already done `reach run` for a project, using this new `reach` script for `reach run` may fail for dumb docker caching reasons. The workaround is to run `reach down` first, which will disassociate the already-built containers from their old docker network.

I'm opening this as a PR as a place to discuss how we want to roll this out to our users. Just tell them to `reach down` if they encounter the known problem? The known problem looks like this:

```
Successfully tagged reachsh/reach-app-tut-6-index:latest
docker-compose -f "docker-compose.yml.index" run --rm reach-app-tut-6-index-${REACH_CONNECTOR_MODE} index
Starting tut-6_ethereum-devnet_1 ... error

ERROR: for tut-6_ethereum-devnet_1  Cannot start service ethereum-devnet: network 0b986a7b7b552a3b8e83a12d99633e438742b6eba1d3874d39f6f6f17b8c6537 not found

ERROR: for ethereum-devnet  Cannot start service ethereum-devnet: network 0b986a7b7b552a3b8e83a12d99633e438742b6eba1d3874d39f6f6f17b8c6537 not found
ERROR: Encountered errors while bringing up the project.
make[1]: *** [run-target] Error 1
make: *** [run] Error 2
```

"cannot start service", "network not found"

### DIFF
--- a/reach
+++ b/reach
@@ -223,11 +223,17 @@ ensure_connector_mode () {
   export REACH_CONNECTOR_MODE
 }
 
+ensure_docker_reach_network () {
+  docker network inspect reach-network >/dev/null 2>&1 || \
+    (echo creating reach-network && docker network create --driver bridge reach-network)
+}
+
 do_whoami () {
   docker info --format '{{.ID}}' 2>/dev/null
 }
 
 do_compile () {
+	ensure_docker_reach_network
     HS=${HERE}/hs
 
     # Note: shellcheck says splatting is dangerous,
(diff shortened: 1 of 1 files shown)


jeapostrophe:
I don't see this attempt to do `reach down`... is that because `reach` already did that?

What if you caught this error and either gave a message "You should run `reach down`", or better yet, you try to do it (once) and then try again?

DanBurton:
I took a slightly different approach. Trying to catch the error after the fact was daunting, so I instead attempted to detect the possibility of the error beforehand and do some pre-emptive cleanup. This seems to work as desired.

DanBurton:
merged a rebased version of this

PR #47, opened 2021-02-01
js: reorg by docker image

DanBurton:
This doesn't necessarily need review; I'm just pushing to a branch first to make sure that CI passes. It's mostly just boring reorganization of stuff, so that the `js` folder now has one sub-folder for each Docker image.

* The `js-deps` and `stdlib` images are not actually published, but they are used by the other two images, `runner` and `react-runner`.
* The `stdlib` image will continue to be used by `js-reach-stdlib`, which means you need to build the `stdlib` image locally (since it is no longer published to dockerhub) if you want to publish changes to the node library. Since I'm usually the one that builds & pushes to dockerhub, as well as publishes to node, there is no change in my workflow.

We should start thinking about whether we want to change the way we publish js libraries. I'll bring this up in slack.


jeapostrophe (approved):
!

PR #50, opened 2021-02-05
RPC server: abstract from tut-6-rpc + stdlib-ify

mattaudesse:


examples/tut-6-rpc/Makefile
```diff
@@ -9,7 +9,7 @@ server/build/%.main.mjs: server/%.rsh
 
 .PHONY: build
 build: server/build/index.main.mjs
-	cd server    && docker build -f Dockerfile --tag=reachsh/reach-app-tut-6-rpc-server:latest    .
+	@# NB: we presume that a reachsh/rpc-server:latest image is already available
```
jeapostrophe, line 12: Yea, so just delete this note.

js/rpc-server/Dockerfile
```diff
@@ -0,0 +1,30 @@
+ARG NODE_VERSION
```
jeapostrophe, line 1: I think this Docker image should either be based on `runner` or actually `runner`

examples/tut-6-rpc/docker-compose.yml
```diff
@@ -1,6 +1,9 @@
 version: '3.4'
 x-server-base: &server-base
-  image: reachsh/reach-app-tut-6-rpc-server:latest
+  image: reachsh/rpc-server:latest
+  volumes:
+    - ./server/build:/app/build
```
jeapostrophe, line 5: This might be a temporary thing, but I think that there should be a custom image generated by `./reach server` that includes this copying into the image rather than mounting it at runtime

js/rpc-server/Dockerfile
mattaudesse: How do we feel about [this](https://github.com/reach-sh/reach-lang/pull/50/commits/abe562ef465a4ed232654d21ebf37f7aefc2379d)?

jeapostrophe: Amen

PR #53, opened 2021-02-10
Signed Integers

chrisnevers:
Add a signed `Int` type to the standard library along with basic operations


docs-src/ref-programs.scrbl
```diff
@@ -1996,3 +1996,39 @@ There are convenience methods defined for comparing fixed point numbers:
...
+
+@index{imod} @reachin{imod(x, y)} finds the remainder of dividing the @reachin{Int} @tt{x} by the @reachin{Int} @tt{y}.
+
+@index{ilt} @reachin{ilt(x, y)} determines whether @tt{x} is less than @tt{y}.
+
+@index{ile} @reachin{ile(x, y)} determines whether @tt{x} is less than or equal to @tt{y}.
+
+@index{igt} @reachin{igt(x, y)} determines whether @tt{x} is greather than @tt{y}.
+
+@index{ige} @reachin{ige(x, y)} determines whether @tt{x} is greater than or equal to @tt{y}.
```
jeapostrophe, line 2034: `ieq`, even though `eq` works just fine

docs-src/ref-programs.scrbl
```diff
@@ -1996,3 +1996,39 @@ There are convenience methods defined for comparing fixed point numbers:
...
+@subsubsection{@tt{Signed Integers}}
+
+The standard library provides abstractions for dealing with signed integers. The following definitions
+are used to represent @reachin{Int}s:
+
+@(mint-define! '("Int") '("Positive") '("Negative"))
+@reach{
+  const Int = { sign: bool, i: UInt };
+  const Positive = true;
+  const Negative = false;  }
```
jeapostrophe, line 2009: `Pos` and `Neg` abbreviations?

jeapostrophe: Maybe hack the compiler so that unary plus/minus are the constants... maybe that's a bad idea?

docs-src/ref-programs.scrbl
```diff
@@ -1996,3 +1996,39 @@ There are convenience methods defined for comparing fixed point numbers:
 @index{fxeq} @reachin{fxeq(x, y)} tests whether @tt{x} is equal to @tt{y}.
 
 @index{fxne} @reachin{fxne(x, y)} tests whether @tt{x} is not equal to @tt{y}.
+
+@subsubsection{@tt{Signed Integers}}
+
+The standard library provides abstractions for dealing with signed integers. The following definitions
```
jeapostrophe, line 2002: Add a `margin-note` about how this is necessary because some Reach target platforms don't support native integers.

jeapostrophe: Add a new card about compiling this efficiently somehow when they do

hs/rsh/stdlib.rsh
```diff
@@ -334,3 +334,36 @@ export const pow = (base, power, precision) =>
...
+  } else {
+    const [ max, min ] =
+      (x.i > y.i) ? [ x, y ] : [ y, x ];
+    return int(max.sign, max.i - min.i);
+  }
+}
+
+export const isub = (x, y) => iadd (x, int(!y.sign, y.i));
+export const imul = (x, y) => int(x.sign == y.sign, x.i * y.i);
+export const idiv = (x, y) => int(x.sign == y.sign, x.i / y.i);
```
jeapostrophe, line 368: Does fixed point math support negative numbers?

docs-src/ref-programs.scrbl
chrisnevers: That would be nice, I'll look into that

hs/rsh/stdlib.rsh
chrisnevers: Besides `fxsqrt` not supporting negative numbers, I don't see why it couldn't be extended to handle negatives. Do you think it's worth having an unsigned  and signed `FixedPoint` or just amend the current definition to include a sign?

jeapostrophe: I think it would be painful to have `ufx` and `sfx`, but I might be naive. It seems like it would be REALLY painful to WANT signed fixed points and not have them.

docs-src/ref-programs.scrbl
chrisnevers: I made it so `FixedPoint` are signed and are a similar shape to `Int`, `{ sign: bool, i: a }`. Now, unary operations will work on both types. Additionally, the unary ops will cast values of type `UInt` to `Int` before applying the operation. The values can now be desugared from:

```javascript
4     // : UInt
+4    // : Int
-4    // : Int
4.0   // : FixedPoint
-4.0  // : FixedPoint
+4.0  // : FixedPoint
```

jeapostrophe (changes requested):
This is awesome. The docs do need to talk about the interpretation of literals though, I didn't miss that did I?

jeapostrophe (approved):

PR #54, opened 2021-02-11
RPC client-js: abstract from tut-6-rpc (now with less Docker)

mattaudesse:


examples/tut-6-rpc/Makefile
```diff
@@ -7,8 +7,14 @@ clean:
 server/build/%.main.mjs: server/%.rsh
 	$(REACH) compile $^ main
 
+# TODO: ditch this once `@reach-sh/rpc-client` package is available on NPM
+.PHONY: client-lib-js
+client-lib-js:
+	cd ../../js/rpc-client && npm pack
+	mv ../../js/rpc-client/reach-sh-rpc-client-0.0.1.tgz ./client-js
```
jeapostrophe, line 14: `cp -fr`?

js/rpc-client/index.mjs
```diff
@@ -0,0 +1,84 @@
...
+      throw new Error(`Neither \`opts.${k}\` nor ${defaults[k]} environment`
+                    + ` variable are configured!`);
+    }
+  });
+
+  const call = bent(`https://${o.host}:${o.port}`, `POST`, `json`, 200, {
+    'X-API-Key': o.apiKey,
+  });
+
+  const rpc = mkRpc(call);
```
jeapostrophe, line 77: My taste is to put these things just inside this function and use nature scope and to use `mkRPC` (basically, I think that the letters in `rpc` should all have the same case and then obey the camel rules.)

jeapostrophe (changes requested):

examples/tut-6-rpc/Makefile
mattaudesse: Hmmm... Maybe I'm missing something obvious, but why is that preferable?

jeapostrophe: No `npm` :)

mattaudesse: Ah, gotcha - thanks.

I'd actually done this without `npm pack` to begin with but I thought it kind of defeated the purpose based on what you said before WRT this being representative of how end users should build their own programs.

I'll tweak this next.

js/rpc-client/index.mjs
mattaudesse: How's c267002 ?

jeapostrophe: parfait

examples/tut-6-rpc/Makefile
```diff
@@ -7,8 +7,14 @@ clean:
 server/build/%.main.mjs: server/%.rsh
 	$(REACH) compile $^ main
 
+# TODO: ditch this once `@reach-sh/rpc-client` package is available on NPM
```
DanBurton, line 10: My understanding is that we won't actually ditch this approach, because we always want CI to run this example against the in-repo version of the library

examples/tut-6-rpc/Makefile
mattaudesse: How about 213d510 ?

jeapostrophe: Coooooool

jeapostrophe:
Push it!

PR #55, opened 2021-02-11
RPC server: use `stdlib.debug` instead of `console.log`

mattaudesse:


examples/tut-6-rpc/docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
     depends_on:
       - algorand-postgres-db
     environment:
-      - REACH_DEBUG
+      - REACH_DEBUG=1
```
jeapostrophe, line 15: We actually want this to be what it was before, because it means that `REACH_DEBUG` from the user will be inherited

jeapostrophe (changes requested):
Tiny change

examples/tut-6-rpc/docker-compose.yml
mattaudesse: Ah, okay - thanks.

mattaudesse: Fixed in d9a6aa6

jeapostrophe (approved):

PR #56, opened 2021-02-13
0944/RPC: loosen newTestAccount stdlib type sig + fix subtly-broken test suite + fix docs discrepancy

mattaudesse:
This pull request corresponds to the
> Push `bigNumberify` into real `stdlib` to widen interface for `stdlib` functions

line item in Trello, and was originally going to include all the other components which need similar treatment, but while working on it I discovered some issues in our current test suite which probably ought to have other eyeballs on before I move forward (in case I'm missing some context).

Likewise with my docs type signature fix; I'm not sure if those functions were labelled as returning `Promise<void>` deliberately for some reason, but AFAICT each of the backend providers' code disagree and this was probably just a mistake.

The test suite issues were subtle and easy to miss but basically boil down to timing gotchas. Before I added the extra `awaits` and rethought the `describe: wait` stanza I was getting failures because the suite presently `expects` very specific block numbers instead of using more dynamic+permissive logic. In my experience, you can make reliable assertions about "not before", and perhaps "only after", but rarely "exactly at".

I'm not in love with how it works now but it seems to mostly stick with the original intention this way... Any thoughts?

### DIFF
--- a/docs-src/ref-backends.scrbl
+++ b/docs-src/ref-backends.scrbl
@@ -263,7 +263,7 @@ this is the current block number, represented as a @litchar{BigNumber}.
 
 @(mint-define! '("waitUntilTime"))
 @js{
- waitUntilTime(time, onProgress) => Promise<void>
+ waitUntilTime(time, onProgress) => Promise<time>
 }
 
 Returns a Promise that will only be resolved after the specified consensus network @tech{time}.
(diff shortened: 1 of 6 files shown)


jeapostrophe (approved):
This looks great. Are there any other places where `export` takes in `BigNumber`? Those should be changed too.

mattaudesse:
> This looks great. Are there any other places where `export` takes in `BigNumber`? Those should be changed too.

Yup, that's what I was saying in the description - I'll knock those out next.

PR #58, opened 2021-02-16
Fix SMT Not Showing Every Error

chrisnevers:
The SMT solver should fail if the negation of a claim (excluding `possible`) is `satisfiable`. There was an issue where subsequent false claims were not causing errors because the solver would incorrectly determine the assertion was `unsatisfiable`. 

## Issue

After a false `assert` was verified (`check_m`), `smt_e` would also assert the expression as well (`assert_m`). 

https://github.com/reach-sh/reach-lang/blob/4a0f9a126ef2f273d2a38bee3c2315f9000cbf3f/hs/src/Reach/Verify/SMT.hs#L853-L858

This is an issue because the claim may have to be negated if it is not `possible`.

https://github.com/reach-sh/reach-lang/blob/4a0f9a126ef2f273d2a38bee3c2315f9000cbf3f/hs/src/Reach/Verify/SMT.hs#L647-L649

This causes the next check to fail since it will try to verify `(assert false)`, as well.

From what I can tell, the `assert_m` in  `check_m <> assert_m` is redundant anyway, since it is performed correctly in `verify1`, so I removed it.

### DIFF
--- a/examples/tut-4-attack/index-bad.txt
+++ b/examples/tut-4-attack/index-bad.txt
@@ -8,9 +8,9 @@ Verification failed:
   at ./index-bad.rsh:45:11:application
 
   // Violation witness
-  const interact_Alice_wager = 1;
+  const interact_Alice_wager = 2;
   //    ^ from interaction at ./index-bad.rsh:14:12:application
-  const v4 = 0;
+  const v4 = 2;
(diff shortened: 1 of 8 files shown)


jeapostrophe (approved):
Awesome, thank you, that was easy to fix, although probably hard to discover :)

PR #60, opened 2021-02-19
Store Binding Information in DLVar

chrisnevers:
Update `DLVar` to remove the `String` field that stored a description of the variable's context and replace it with binding origin information, `Just (SrcLoc, SLVar)`, related to the source program. Update **SMT** and **Knowledge** to use this information when displaying errors.


examples/overview/build/index.main.mjs
```diff
@@ -5,6 +5,7 @@ export const _version = '0.1.2';
```
jeapostrophe, line 5: I try not to commit these files if nothing interesting is different, because it creates noise in the history

examples/tut-7/build/index.main.mjs
```diff
@@ -77,26 +78,26 @@ export async function Alice(ctc, interact) {
...
       const v194 = txn3.value;
       const v199 = txn3.time;
       const v193 = txn3.from;
       
       const v195 = stdlib.eq(v194, stdlib.checkedBigNumberify('reach standard library:decimal', stdlib.UInt_max, 0));
       stdlib.assert(v195, {
-        at: 'reach standard library:67:7:dot',
-        fs: ['at ./index.rsh:51:41:application call to [unknown function] (defined at: reach standard library:66:8:function exp)'],
+        at: 'reach standard library:66:7:dot',
+        fs: ['at ./index.rsh:51:41:application call to [unknown function] (defined at: reach standard library:65:8:function exp)'],
```
jeapostrophe, line 93: Can the labels on closures get infected too and maybe this `unknown function` will go away?

hs/src/Reach/Eval/Core.hs
```diff
@@ -2144,37 +2146,28 @@ evalPropertyPair fenv = \case
     --- FIXME support these
     expect_ $ Err_Obj_IllegalMethodDefinition p
 
-evalId_ :: String -> SLVar -> App SLSSVal
-evalId_ lab x = do
+evalId_ :: SLVar -> App SLSSVal
+evalId_ x = do
   sco <- e_sco <$> ask
-  let c = ("consensus environment", return $ sco_cenv sco)
```
jeapostrophe, line 2150: You need to rebase

hs/src/Reach/Eval/Core.hs
```diff
@@ -2491,7 +2484,12 @@ evalDeclLHS rhs_lvl lhs_env v = \case
   JSIdentifier a x -> do
     locAtf (srcloc_jsa "id" a) $ do
       at_ <- withAt id
-      env_insert x (SLSSVal at_ rhs_lvl v) lhs_env
+      let bindingOrigin = Just (at_, x)
+      let v' =
```
jeapostrophe, line 2488: Can infect be abstracted and used here?

hs/src/Reach/Eval/Module.hs
```diff
@@ -41,7 +41,7 @@ evalImportClause env im =
       (at', ns) <- withAt $ flip parseIdent ji
       return $ M.singleton ns $ (SLSSVal at' Public $ SLV_Object at' (Just $ "module " <> ns) env)
     JSImportClauseNamed (JSImportsNamed _ iscl _) ->
-      evalImExportSpecifiers (LC_RefFrom "module import") env go iscl
+      evalImExportSpecifiers LC_RefFrom env go iscl
```
jeapostrophe, line 44: I don't understand why this is removed... because we do want to continue to get this nice error... but maybe I'll see something good when I get to the examples

hs/src/Reach/Linearize.hs
```diff
@@ -368,7 +368,7 @@ df_con = \case
           case r of
             Nothing -> return $ Nothing
             Just _ -> do
-              dv <- allocVar $ DLVar at (show fv) (fluidVarType fv)
+              dv <- allocVar $ DLVar at (Just (srcloc_builtin, show fv)) (fluidVarType fv)
```
jeapostrophe, line 371: I think it should be `Nothing`, because the user never wrote `show fv`... but then again, we want to know that `v8` is the balance... so maybe it could be `balance()` or `lastConsensusTime()`?

hs/src/Reach/Verify/Knowledge.hs
```diff
@@ -101,12 +101,16 @@ displayPath who = \case
...
+        show $ pretty n <> "/" <> pretty i <> " (defined at " <> pretty at <> ")"
+      P_Var (DLVar at' Nothing _ i) ->
         show $ "v" <> pretty i <> " (defined at " <> pretty at' <> ")"
       ow -> sp ow
     publishInfo = \case
-      P_Var (DLVar at' _ _ i) ->
-        show $ "v" <> pretty i <> " was published at " <> pretty at'
+      P_Var (DLVar at' (Just (_, n))  _ i) ->
+        show $ pretty n <> "/" <> pretty i <> " was published at " <> pretty at'
+      P_Var (DLVar at' Nothing _ i) ->
```
jeapostrophe, line 112: This is begging for a function

hs/src/Reach/Verify/SMT.hs
```diff
@@ -231,9 +232,12 @@ smtVar :: DLVar -> App String
...
   pvs <- ctxt_primed_vars <$> ask
   let mp = case elem dv pvs of
-        True -> "p"
-        False -> ""
-  return $ "v" ++ show i ++ mp
+            True -> "p"
+            False -> ""
+  let name = "v" ++ show i ++ mp
+  v2dv <- ctxt_v_to_dv <$> ask
+  liftIO $ modifyIORef v2dv $ M.insert name dv
```
jeapostrophe, line 239: This map, `v2dv`, is going to have duplicates in it and you're going to arbitrarily select the last one that was seen. I think you should try to come up with an example where the same values gets multiple names and try to see if you can get bad output. This then is a place to insert that "most specific" logic we talked about.

hs/test-examples/nl-eval-errors/Err_Export_Missing.txt
```diff
@@ -1 +1 @@
-reachc: error: ./Err_Export_Missing.rsh:8:1:export: Invalid unbound identifier in module export: x. Did you mean: ["fx","eq","ge","gt","le"]
+reachc: error: ./Err_Export_Missing.rsh:8:1:export: Invalid unbound identifier: x. Did you mean: ["fx","eq","ge","gt","le"]
```
jeapostrophe, line 1: I think this is an example of a worse message

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
```diff
@@ -831,6 +831,8 @@ infectWithId_sv :: SrcLoc -> SLVar -> SLVal -> SLVal
 infectWithId_sv at v = \case
   SLV_Participant a who _ mdv ->
     SLV_Participant a who (Just v) mdv
+  SLV_Clo a Nothing e b c ->
```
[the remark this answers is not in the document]
chrisnevers: Do you mean like this:
```javascript
const x = () => 5;
const y = x;
y();
```
will show as y?

hs/src/Reach/Eval/Core.hs
```diff
@@ -1599,7 +1603,7 @@ evalPrim p sargs =
             (T_Array x_ty x_sz, T_Array y_ty y_sz) -> do
               meet_ty <- dtypeMeet (at, x_ty) (at, y_ty)
               let t = T_Array meet_ty (x_sz + y_sz)
-              let mkdv = DLVar at "array_concat" t
+              let mkdv = DLVar at Nothing t
```
DanBurton, line 1606: Isn't the idea to forget less stuff? Would it make sense to make this slot an `Either String Blah` instead of a `Maybe Blah` so that these informative strings can be preserved?

jeapostrophe: In practice, these strings don't saying anything more than the srclocs, and I haven't even found them useful as the compiler author

chrisnevers: From what I can tell, this field was not used for anything/anywhere

jeapostrophe (approved):
Sorry for the delay

PR #61, opened 2021-02-19
reach server

DanBurton:
* introduces `reach server` command
* changes tut-6-rpc to run using `reach server`
  * tut-6-rpc now runs tests for js, py, and go clients w/ ETH
  * tut-6-rpc tests do not work correctly w/ ALGO, created new issue: https://trello.com/c/dMjRVT7x/1036-make-tut-6-rpc-using-reach-server-work-w-algo


jeapostrophe (approved):
Noiiice

PR #66, opened 2021-02-26
Add `timeout` short hand for `parallel_reduce`

chrisnevers:
Add a shorthand in `parallel_reduce` for specifying:

```Javascript
.timeout(timeRemaining(), () => {
    race(...Participants).publish();
    return LHS; })
```

It can now be written simply as:

```Javascript
.time_remaining(timeRemaining())
```

### DIFF
--- a/docs-src/ref-programs.scrbl
+++ b/docs-src/ref-programs.scrbl
@@ -751,10 +751,13 @@ while the @reachin{.case} and @reachin{.timeout} components are like the corresp
 
 The @reachin{.case} component may be repeated many times, provided the @reachin{PART_EXPR}s each evaluate to a unique @tech{participant}, just like in a @reachin{fork} statement.
 
-In the case of absolute deadlines, there is a common pattern in the @reachin{TIMEOUT_BLOCK} to have participants
-@reachin{race} to @reachin{publish} and return the accumulator. There is a shorthand, @reachin{.time_remaining},
-available for this situation:
+@subsubsub*section{@tt{.time_remaining}}
 
+When dealing with absolute deadlines in @reachin{parallel_reduce}, there is a common pattern in the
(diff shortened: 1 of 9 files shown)


docs-src/ref-programs.scrbl
```diff
@@ -1902,7 +1921,26 @@ expression. For example:
     .invariant(...)
     .while( keepGoing() )
     .case(...)
-    .timeout( timeRemaining(), ...) }
+    .timeout( timeRemaining(), () => { ... }) }
+
+When dealing with absolute deadlines in @reachin{parallel_reduce}, there is a common pattern in the
```
jeapostrophe, line 1926: I think that the expansion should be explained only once and instead this block of docs should say... "this pattern is so common, it is abbreviate as `.time_remaining`", which would be a link to the further explanation. I am scared of copy & paste

jeapostrophe:
Bikeshed: maybe `timeoutWithDeadline` or something? I feel like we're not sufficiently committing to snakes or camels

jeapostrophe:
`timeoutWithDeadline` is long and ugly and doesn't really make sense

DanBurton:
My only bikeshed comment here is that if we are trying to "feel like JavaScript" then we should move away from snake_case names like `parallel_reduce` and `time_remaining` and instead use their camelCase counterparts.

But the general idea of having a contraction for this common pattern seems like a good one.

chrisnevers:
As far as case goes, I just used snake because thats the case of `parallel_reduce` and I thought it would be weird to mix them. And as far as naming goes, I'm open to suggestions. I'm not ecstatic over `timeRemaining` but it's hard to think of something terse and descriptive.

jeapostrophe:
I agree and think we should worry about getting right of snakes in the future.

jeapostrophe (approved):

PR #68, opened 2021-03-01
Map & Set Standard Library Functions

chrisnevers:
* Abstract `Array` stdlib functions that use `reduce` into `Foldable` functions. 
* Make `Array` and `Map` reduction functions point to `Foldable` functions.
* Create `Set` data type.
* Add `Set` functions to standard library.

### DIFF
--- a/hs/src/Reach/Eval/Core.hs
+++ b/hs/src/Reach/Eval/Core.hs
@@ -2166,6 +2166,8 @@ getKwdOrPrim ident env =
     -- Hack: Allow `default` to be used as object property name for `match`
     -- expr. Keywords are allowed as property names in JS anyway, *shrug*
     Just (SLSSVal _ _ (SLV_Kwd SLK_default)) -> Nothing
+    -- Allow `new` to be used as object property name for `Set`.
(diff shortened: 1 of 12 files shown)


hs/src/Reach/Eval/Core.hs
```diff
@@ -1013,14 +1006,23 @@ evalAsEnvM obj = case obj of
...
-        [ ("reduce", retV $ public $ SLV_Prim $ SLPrim_MapReduce) ]
+      M.fromList $
+        [ ("reduce", retV $ public $ SLV_Prim $ SLPrim_MapReduce) ] <> foldableValueEnv
   SLV_Map _ ->
     Just $
-      M.fromList
-        [ ("reduce", delayCall SLPrim_MapReduce) ]
+      M.fromList $
+        [ ("reduce", delayCall SLPrim_MapReduce) ] <> foldableObjectEnv
+  SLV_Prim SLPrim_Set ->
```
jeapostrophe, line 1015: I think that you could just define `Set` in the standard library as `const Set = () => { const s = new Map(Null); return { ..... };` and not modify Core.hs

jeapostrophe (changes requested):

PR #69, opened 2021-03-02
RPC server/stdlib: improve continuation-tracking

mattaudesse:
Completes the following line items in Trello #0944:
 + Use random continuation IDs in server
 + Use a better data-structure... (for continuation-tracking)
 + Add deletion functions... (for account/contract-tracking)
 + Make RPC server more robust...

Additionally:
 + Each of the JS/Golang/Python clients have been updated to use
   the new `forget` interface
 + The stdlib test suite has been updated to enhance readability and
   exercise the new continuation-tracking machinery

### DIFF
--- a/js/stdlib/ts/rpc_server.ts
+++ b/js/stdlib/ts/rpc_server.ts
@@ -30,11 +30,10 @@ const withApiKey = () => {
 const mkKont = () => {
   // TODO consider replacing stringly-typed exceptions with structured
   // descendants of `Error` base class
-  const COLLISION = 'Collision on continuation ID:';
   const UNTRACKED = 'Untracked continuation ID:';
-  const collision = (i: string) => `${COLLISION} ${i}`;
   const untracked = (i: string) => `${UNTRACKED} ${i}`;
   const k: any    = {};
+  let   i: number = 0;
 
   const mkWas = (m: string) => (e: Error): boolean =>
     !!(e.message
@@ -42,7 +41,6 @@ const mkKont = () => {
       .match(`^${m}$`));
 
   const was = {
-    collision: mkWas(COLLISION),
     untracked: mkWas(UNTRACKED),
   };
 
@@ -50,15 +48,14 @@ const mkKont = () => {
     throw new Error(e);
   };
 
-  const track = (a: any, bytes: number = 24) =>
-    new Promise<string>((res, rej) => randomBytes(bytes, (e, b) =>
-        e ? rej(e)
-          : res(b.toString('hex'))))
-      .then(i =>
-        k[i] !== undefined
-          ? Promise.reject(collision(i))
-          : (() => { k[i] = a; return i; })())
-      .catch(raise);
+  const track = async (a: any) => {
+    const rb = await randomBytes(24);
+    const id = `${i}_${rb.toString('hex')}`;
+    k[id]    = a;
+    i++;
+
+    return id;
+  };
 
   const id = (i: string) =>
     k[i] === undefined
@@ -77,9 +74,8 @@ const mkKont = () => {
     // Internals
     _: {
       k,
-      COLLISION,
+      i,
       UNTRACKED,
-      collision,
       untracked,
     },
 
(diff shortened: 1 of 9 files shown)


mattaudesse:
Here's an example of some stdlib test suite output with the extra formatting + `expect` helpers:

```
> @reach-sh/stdlib-test-index@ index /app
> node --experimental-modules --unhandled-rejections=strict index.mjs

(node:16) ExperimentalWarning: The ESM module loader is experimental.
The `ETH` stdlib
  exposes a `bigNumberToHex` function that
    correctly translates positive `BigNumber`s to hex
    correctly translates negative `BigNumber`s to hex

  exposes a `bigNumberToNumber` function that
    correctly translates from `bigNumber` to `Number`

  exposes a `stringToHex` function that
    works correctly with `bytes_eq`

  exposes a `bigNumberify` function that
    correctly translates integer inputs to their `BigNumber` equivalents
    correctly translates string inputs to their `BigNumber` equivalents

  exposes an `isBigNumber` function that
    returns `true` for `BigNumber` arguments
    returns `false` for non-`BigNumber` arguments

  exposes a `BigNumber` comparison function called
    `eq` that
      returns `true` when its arguments match
      returns `false` when provided mismatched arguments

    `ge` that
      returns `true` when its first argument is greater than or equal to its second
      returns `false` when its first argument is less than its second

    `gt` that
      returns `true` when its first argument is greater than its second
      returns `false` when its first argument is equal to or less than its second

    `le` that
      returns `true` when its first argument is lesser than or equal to its second
      returns `false` when its first argument is greater than its second

    `lt` that
      returns `true` when its first argument is lesser than its second
      returns `false` when its first argument is equal to or greater than its second

  exposes a `BigNumber` arithmetic function called
    `add` that sums its arguments
    `sub` that subtracts its second argument from its first
    `mod` that returns the remainder of its first argument divided by its second
    `mul` that returns the product of its arguments

  exposes a `newTestAccount` function which
    accepts numeric arguments of type
      `BigNumber`
      raw JavaScript `Number`

  exposes a `createAccount` function which
    constructs new accounts with zero balances

  exposes a `fundFromFaucet` function which
    can fund testnet accounts with `BigNumber` values
    can fund testnet accounts with JavaScript `Number` values

  exposes a `protect` function that
    converts nully things to Null
    converts bytesy things to Bytes
    converts numbery things to UInt
    handles Bool
    handles Address
    recurses into Tuples
    recurses into Objects
    recurses into Arrays
    recurses into Data

  wait
    prog0: 1
    prog0: 2
    prog1: 1
    prog1: 2
    prog2: 1
    prog2: 2
    prog2: 3
    prog2: 4
    prog2: 5
    prog2: 6

  exposes a `getDefaultAccount` function which
    represents a faucet when running against a devnet

  exposes a `newAccountFromSecret` function which
    begins life with a zero balance

  exposes a `connectAccount` function which connects when given
    a => a.networkAccount

  T_Address.canonicalize can handle multiple inputs
    T_address.canonicalize(accRelay) returns address
    T_address.canonicalize(accRelay.networkAccount) returns address
    T_address.canonicalize(accRelay.networkAccount.address) returns address

  exports
    should only export a few extra things compared to ALGO
    should export everything that ALGO does
    should only export a few extra things compared to FAKE
    should export everything that FAKE does

  currency conversions
    should convert to WEI correctly
    should convert from WEI correctly


 ---

The `ETH` RPC server
  exposes a `bigNumberToHex` function that
    correctly translates positive `BigNumber`s to hex
    correctly translates negative `BigNumber`s to hex

  exposes a `bigNumberToNumber` function that
    correctly translates from `bigNumber` to `Number`

  exposes a `stringToHex` function that
    works correctly with `bytes_eq`

  exposes a `bigNumberify` function that
    correctly translates integer inputs to their `BigNumber` equivalents
    correctly translates string inputs to their `BigNumber` equivalents

  exposes an `isBigNumber` function that
    returns `true` for `BigNumber` arguments
    returns `false` for non-`BigNumber` arguments

  exposes a `BigNumber` comparison function called
    `eq` that
      returns `true` when its arguments match
      returns `false` when provided mismatched arguments

    `ge` that
      returns `true` when its first argument is greater than or equal to its second
      returns `false` when its first argument is less than its second

    `gt` that
      returns `true` when its first argument is greater than its second
      returns `false` when its first argument is equal to or less than its second

    `le` that
      returns `true` when its first argument is lesser than or equal to its second
      returns `false` when its first argument is greater than its second

    `lt` that
      returns `true` when its first argument is lesser than its second
      returns `false` when its first argument is equal to or greater than its second

  exposes a `BigNumber` arithmetic function called
    `add` that sums its arguments
    `sub` that subtracts its second argument from its first
    `mod` that returns the remainder of its first argument divided by its second
    `mul` that returns the product of its arguments

  exposes a `newTestAccount` function which
    accepts numeric arguments of type
      `BigNumber`
      raw JavaScript `Number`

  exposes a `createAccount` function which
    constructs new accounts with zero balances

  exposes a `fundFromFaucet` function which
    can fund testnet accounts with `BigNumber` values
    can fund testnet accounts with JavaScript `Number` values

  exposes a `protect` function that
    converts nully things to Null
    converts bytesy things to Bytes
    converts numbery things to UInt
    handles Bool
    handles Address
    recurses into Tuples
    recurses into Objects
    recurses into Arrays
    recurses into Data

  wait
    prog0: 1
    prog0: 2
    prog1: 1
    prog1: 2
    prog2: 1
    prog2: 2
    prog2: 3
    prog2: 4
    prog2: 5
    prog2: 6

  exposes a `getDefaultAccount` function which
    represents a faucet when running against a devnet

  exposes a `newAccountFromSecret` function which
    begins life with a zero balance

  exposes a `connectAccount` function which connects when given
    a => a

  exposes a `mkKont` constructor which provides
    an `id` function that
      raises an exception for untracked IDs
      fetches the entry indexed by a given ID

    a `track` function that
      generates random IDs
        of length 48 (by default)
        which index its `a` argument in `k`
        which are guaranteed to be lowercase, non-0x hexadecimal

      raises an exception if collision occurs in `k` indices

    a `replace` function that
      swaps the entry indexed by a given ID and returns the same ID
      raises an exception for untracked IDs

    a `forget` function that
      clears a given index from `k`

    a `was` namespace which exposes
      a `collision` function that can correctly
        identify exceptions raised due to collision in `k`
        distinguish other exceptions

      an `untracked` function that can correctly
        identify exceptions raised due to untracked IDs
        distinguish other exceptions
```

js/stdlib/ts/rpc_server.ts
```diff
@@ -26,84 +27,177 @@ const withApiKey = () => {
...
+    collision: mkWas(COLLISION),
+    untracked: mkWas(UNTRACKED),
+  };
+
+  const raise = (e: string) => {
+    throw new Error(e);
+  };
+
+  const track = (a: any, bytes: number = 24) =>
+    new Promise<string>((res, rej) => randomBytes(bytes, (e, b) =>
```
jeapostrophe, line 54: As mentioned in Slack, I think you should change this to `randomBytes() + "_" + id++` so there's no risk of collision, get rid of `collision` and `safely`. Also, I prefer `async` over explicitly making Promises. I am too weak of a programmer to hand Promises themselves

jeapostrophe (changes requested):

js/stdlib/ts/rpc_server.ts
mattaudesse: How does 5debabcc6d4d76c8313a3b7c37be85b0d9a2e4d7 look?

I've left the "untracked" bits as they were because I think it's important to capture those and do something deliberate at the HTTP layer.

jeapostrophe (commented):
Beautiful

PR #70, opened 2021-03-04
Update SMT To Account For Primed Vars

chrisnevers:
* Ensure that **SMT**'s `ctxt_bindingsr` has an updated `DLExpr` in its environment. 
* When displaying SMT errors and counting the occurrence of variables in `dlvOccurs`, assume that if a given variable is in `ctxt_primed_vars` then the primed var should be used over the original. This is a safe assumption because `ctxt_primed_vars` will only be populated when processing `continue` and not the _before_ of a loop.

### DIFF
--- a/hs/src/Reach/Verify/SMT.hs
+++ b/hs/src/Reach/Verify/SMT.hs
@@ -1020,6 +1020,8 @@ sub_loop_vars_de de = do
 
 smt_e :: SrcLoc -> Maybe DLVar -> DLExpr -> SMTComp
 smt_e at_dv mdv de = do
+  -- We could change `ctxt_loop_var_subst` to `Map DLVar DLVar`
+  -- by having jumps create new fresh variables then re-use `Subst` here
   de' <- sub_loop_vars_de de
   case de of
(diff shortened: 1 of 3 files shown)


hs/src/Reach/Verify/SMT.hs
```diff
@@ -971,25 +976,68 @@ smt_la at_de dla = do
       vv' <- smt_a at_de vv
       return $ smtApply (s ++ "_" ++ vn) [vv']
 
+sub_loop_vars_a :: M.Map DLVar DLArg -> DLArg -> DLArg
```
jeapostrophe, line 979: These functions might be `Subst`

jeapostrophe (commented):
If it is `Subst`, please use that instead of duplicating

hs/src/Reach/Verify/SMT.hs
chrisnevers: If I change `SubstEnv` from `Map DLVar DLVar` to `Map DLVar DLArg`, remove the instance`Subst DLVar` and tweak `DeJump` accordingly I can re-use it. Any concern with removing the `subst dlvar` instance?

PR #71, opened 2021-03-04
Default Function Arguments

chrisnevers:
* Add `verifyFormals` to confirm parameters with default arguments come after normal parameters. Used when evaluating arrow expressions.
* Add `instDefaultArgs` to utilize default args if necessary. Used when evaluating closure applications.


hs/src/Reach/Eval/Core.hs
```diff
@@ -2045,10 +2065,11 @@ evalApplyVals rator randvs =
     SLV_Clo clo_at mname formals (JSBlock body_a body _) (SLCloEnv clo_penvs clo_cenv) -> do
       ret <- ctxt_alloc
       let body_at = srcloc_jsa "block" body_a clo_at
+      let err = Err_Apply_ArgCount clo_at (length formals) (length randvs)
+      instArgs <- instDefaultArgs err formals randvs
```
jeapostrophe, line 2069: This function includes a call to `evalExpr`, which means that the way you have it not, it has access to scope (i.e. environment) of the caller and it can make changes to the state. I think this is dangerous.

Instead, I think it should be evaluated in the closure environment and with a pure version of whichever mode we're currently in (local or consensus)

Later, it would be good to be able to support things like `(x, y, z = (x + y)) => ....`, which would require more work (so make a note of that somewhere, either in the code & a non-example or in trello)

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
chrisnevers: That makes sense, is there any pattern already established for evaluating in a new environment? Should I just temporarily change `e_st.st_mode` and `e_sco`?

jeapostrophe: Use `locStMode` for changing the mode and `locSco`. You should be able to do `locStMode (pureVersionOf whatEverCurrentModeIs) $ locSco clo_sco $ instDefaultArgs .....` and then if you dig deeper and get the `(x + y)` example working, it will require some surgery in `evalDecls` (which I *think* should NOT apply to other places where `evalDecls` is used)

chrisnevers: I ended up not having to make any changes to `evalDecl` by inlining the process into `instDefaultArgs` as opposed to doing it sequentially

hs/test-examples/features/default_fn_arguments_dependant.rsh
```diff
@@ -0,0 +1,14 @@
...
+
+export const foo = (x, y, z = (x + y + a), t = z - 2, u = x + x) =>
+  x + y + z + t + u;
+
+export const main =
+  Reach.App(
+    {},
+    [Participant('A', {})],
+    (A) => {
+      assert(foo(1, 2) == 13);
```
jeapostrophe, line 13: Can we add a `const a = 10;` or something above this as a test that it isn't using the caller environment? (I believe the code is right, but I want that tested explicitly)

hs/src/Reach/Eval/Core.hs
```diff
@@ -2078,6 +2076,12 @@ evalApplyVals rator randvs =
                , sco_penvs = clo_penvs
                , sco_cenv = clo_cenv
                })
+      m <- readSt st_mode
+      arg_env <-
+        locStMode (pure_mode m) $
+          locSco clo_sco $
+            instDefaultArgs mempty err formals randvs
+      at <- withAt id
```
jeapostrophe, line 2084: Very nice!

jeapostrophe (changes requested):

jeapostrophe:
Otherwise, good to go!

PR #72, opened 2021-03-08
Rename Snake Case Words To Camel Case

chrisnevers:
* Add a deprecation warning to `parallel_reduce` that points to `parallelReduce`. 
* `fxpow_ratio` isn't exported from the stdlib but it has been renamed to `fxpowRatio`.
* Rename `.time_remaining` to `.timeRemaining`, I _think_ it's safe to just rename this one as well since it's a week old?
* Rename variable names in examples to camel case.


jeapostrophe (approved):
It would be nice to have an example in the tests that shows the snake case deprecation warning

PR #77, opened 2021-03-11
'Use Strict" Mode

chrisnevers:
`'use strict'` enables unused variables warnings for all subsequent declarations within the current scope. Since `'use strict'` is scope sensitive, you can do some cool things like use it in a single thunk etc.

## Implementation:
* Keep a global list of `(SrcLoc * SLVar)` in `App` to track variable declarations.
* Keep a bool indicative of `use_strict` in `SLScope`
* When a variable is declared via `evalDecl` add it to environment if we are in strict mode.
* When a variable is looked up, mark it as seen if we are in strict mode.
* After evaluation, list off any variables that were unused.
* Copy the current `sco_use_strict` to the `SLCloEnv` so closures use the mode from when they were declared as opposed to when they are inlined.
* Added a `trackVars` parameter to `evalDecl` because we should not treat the mutations before a `continue` as variable declarations.
 
## Importing A Non-Strict Lib Into A Strict File

No warnings are emitted when importing a non-strict library into a strict file:

#### dynamicLib.rsh
```javascript
'reach 0.1';

export function foo(a, b) {
  void a;
  return;
}
```

#### index.rsh
```javascript
'reach 0.1';
'use strict';

import { foo } from './dynamicLib.rsh';

export const main =
  Reach.App(
    {},
    [Participant('A', { x : Bool })],
    (A) => {
      void foo(1, 2);
    });
```

```terminal
  Verifying when ONLY "A" is honest
Checked 3 theorems; No failures!
```

## Importing A Strict Lib Into Non-Strict File

Warnings are emitted when importing a strict lib to a non-strict file:

#### staticLib.rsh
```javascript
'reach 0.1';

'use strict';

export function foo(x, y) {
  void x;
  return;
}

export const aOnly = () => {
  const x = 5;
}
```

#### index.rsh
```javascript
'reach 0.1';

import { foo, aOnly } from './staticLib.rsh';

export const main =
  Reach.App(
    {},
    [Participant('A', { x : Bool })],
    (A) => {
      A.only(aOnly);
      foo(1, 2);
    });
```

```terminal
WARNING: unused variable: y at /Users/chris/dev/reach/test/index.rsh:5:24:id
WARNING: unused variable: x at /Users/chris/dev/reach/test/staticLib.rsh:11:9:id
  Verifying when ONLY "A" is honest
Checked 3 theorems; No failures!
```

### DIFF
--- a/docs-src/ref-programs.scrbl
+++ b/docs-src/ref-programs.scrbl
@@ -2306,9 +2306,9 @@ In an application without any @tech{participant class}es, @reachin{Anybody} inst
 
 @index{'use strict'} @reachin{'use strict'} enables unused variables checks for all subsequent
 declarations within the current scope. If a variable is declared, but never used, there will
-be a warning emitted at compile time.
+be a error emitted at compile time.
 
-Strict mode will reject some code that is normally valid and limit how dynamic Reach's type system is.
+@deftech{strict mode} will reject some code that is normally valid and limit how dynamic Reach's type system is.
 For example, normally Reach will permit expressions like the following to be evaluated:
 
 @reach{
@@ -2318,12 +2318,27 @@ For example, normally Reach will permit expressions like the following to be eva
   void foo({ b: true });
   void foo(false); }
 
-Without @reachin{'use strict'}, Reach will not evaluate @reachin{o.b} when @reachin{o = false} and this code will compile
-successfully.
+Reach allows @reachin{o} to be either an object with a @reachin{b} field or @reachin{false} because it
+partially evaluates the program at compile time. So, without @reachin{'use strict'}, Reach will not evaluate
+@reachin{o.b} when @reachin{o = false} and this code will compile successfully.
 
-However, to determine variable usage in strict mode, Reach traverses every conditional branch,
-whether or not it is the chosen one. In doing so, Reach will
-raise an error when reviewing the true branch of the ternary:
+But, in @tech{strict mode}, Reach will ensure that this program treats @reachin{o} as
+having a single type and detect an error in the program as follows:
 
 @verbatim{
   reachc: error: Invalid field access. Expected object, got: Bool }
+
+The correct way to write a program like this in @tech{strict mode} is to use @reachin{Maybe}. Like this:
+
+@reach{
+  const MObj = Maybe(Object({ b : Bool }));
+
+  const foo = (mo) =>
+    mo.match({
+      None: (() => false),
+      Some: ((o) => o.b)
+    });
+
+  void foo(MObj.Some({ b : true }));
+  void foo(MObj.None()); }
+
--- a/docs-src/ref-programs.scrbl
+++ b/docs-src/ref-programs.scrbl
@@ -2297,3 +2297,33 @@ it does not matter who @reachin{publish}es, such as in a @reachin{timeout}.
 @reachin{Anybody} is strictly an abbreviation of a @reachin{race} involving all of the named participants of the application.
 In an application with a @tech{participant class}, this means any principal at all, because there is no restriction on which principals (i.e. addresses) may serve as a member of that class.
 In an application without any @tech{participant class}es, @reachin{Anybody} instead would mean only the actual previously-bound @tech{participant}s.
+
+@subsubsection{'use strict'}
+
+@(mint-define! '("'use strict'"))
+@reach{
+  'use strict'; }
+
+@index{'use strict'} @reachin{'use strict'} enables unused variables checks for all subsequent
+declarations within the current scope. If a variable is declared, but never used, there will
+be a warning emitted at compile time.
+
+Strict mode will reject some code that is normally valid and limit how dynamic Reach's type system is.
+For example, normally Reach will permit expressions like the following to be evaluated:
+
+@reach{
+  const foo = (o) =>
+    o ? o.b : false;
+
+  void foo({ b: true });
+  void foo(false); }
+
+Without @reachin{'use strict'}, Reach will not evaluate @reachin{o.b} when @reachin{o = false} and this code will compile
+successfully.
+
+However, to determine variable usage in strict mode, Reach traverses every conditional branch,
+whether or not it is the chosen one. In doing so, Reach will
+raise an error when reviewing the true branch of the ternary:
+
+@verbatim{
+  reachc: error: Invalid field access. Expected object, got: Bool }
--- a/hs/src/Reach/AST/SL.hs
+++ b/hs/src/Reach/AST/SL.hs
@@ -97,7 +97,7 @@ st2it t = case t of
 type SLPartEnvs = M.Map SLPart SLEnv
 
 data SLCloEnv
-  = SLCloEnv SLPartEnvs SLEnv
+  = SLCloEnv SLPartEnvs SLEnv Bool
   deriving (Eq, Generic)
 
 data SLVal
@@ -398,7 +398,7 @@ data SLSSVal = SLSSVal
   deriving (Eq, Generic)
 
 instance Pretty SLSSVal where
-  pretty (SLSSVal at level val) = viaShow at <> pretty (level, val) -- <> "@" <> pretty at
+  pretty (SLSSVal _ level val) = pretty (level, val) -- <> "@" <> pretty at
   -- TODO: incorporate srcloc in pretty?
 
 sss_restrict :: SecurityLevel -> SLSSVal -> SLSSVal
--- a/hs/src/Reach/AST/SL.hs
+++ b/hs/src/Reach/AST/SL.hs
@@ -16,7 +16,7 @@ import Reach.AST.DLBase
 import Reach.JSOrphans ()
 import Reach.Texty
 import Reach.Util
-import Reach.Deprecation (Deprecation)
+import Reach.Warning (Deprecation)
 
 -- SL types are a superset of DL types.
 -- We copy/paste constructors instead of using `ST_Val DLType`
@@ -398,7 +398,7 @@ data SLSSVal = SLSSVal
   deriving (Eq, Generic)
 
 instance Pretty SLSSVal where
-  pretty (SLSSVal _ level val) = pretty (level, val) -- <> "@" <> pretty at
+  pretty (SLSSVal at level val) = viaShow at <> pretty (level, val) -- <> "@" <> pretty at
   -- TODO: incorporate srcloc in pretty?
 
 sss_restrict :: SecurityLevel -> SLSSVal -> SLSSVal
--- a/hs/src/Reach/AST/SL.hs
+++ b/hs/src/Reach/AST/SL.hs
@@ -97,7 +97,11 @@ st2it t = case t of
 type SLPartEnvs = M.Map SLPart SLEnv
 
 data SLCloEnv
-  = SLCloEnv SLPartEnvs SLEnv Bool
+  = SLCloEnv {
+    clo_penvs :: SLPartEnvs,
+    clo_cenv :: SLEnv,
+    clo_use_strict :: Bool
+  }
   deriving (Eq, Generic)
 
 data SLVal
--- a/hs/src/Reach/Eval.hs
+++ b/hs/src/Reach/Eval.hs
@@ -200,6 +200,9 @@ compileBundle cns jsb main = do
   ms' <- readIORef me_ms
   final <- readIORef e_lifts
   unused_vars <- readIORef e_unused_variables
-  unless (null unused_vars) $
-    error $ show $ Err_Unused_Variables $ S.toList unused_vars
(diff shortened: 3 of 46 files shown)


docs-src/ref-programs.scrbl
```diff
@@ -2297,3 +2297,33 @@ it does not matter who @reachin{publish}es, such as in a @reachin{timeout}.
...
+
+@subsubsection{'use strict'}
+
+@(mint-define! '("'use strict'"))
+@reach{
+  'use strict'; }
+
+@index{'use strict'} @reachin{'use strict'} enables unused variables checks for all subsequent
+declarations within the current scope. If a variable is declared, but never used, there will
+be a warning emitted at compile time.
```
jeapostrophe, line 2309: Do you feel strongly about warning versus error? I assumed we'd do error.

docs-src/ref-programs.scrbl
```diff
@@ -2297,3 +2297,33 @@ it does not matter who @reachin{publish}es, such as in a @reachin{timeout}.
...
+
+Strict mode will reject some code that is normally valid and limit how dynamic Reach's type system is.
+For example, normally Reach will permit expressions like the following to be evaluated:
+
+@reach{
+  const foo = (o) =>
+    o ? o.b : false;
+
+  void foo({ b: true });
+  void foo(false); }
```
jeapostrophe, line 2319: "Reach allows `o` to be either an object with a `b` field or `false, because it partially evaluates the program at compile time."

docs-src/ref-programs.scrbl
```diff
@@ -2297,3 +2297,33 @@ it does not matter who @reachin{publish}es, such as in a @reachin{timeout}.
...
+
+Without @reachin{'use strict'}, Reach will not evaluate @reachin{o.b} when @reachin{o = false} and this code will compile
+successfully.
```
jeapostrophe, line 2322: "So, without..."

docs-src/ref-programs.scrbl
```diff
@@ -2297,3 +2297,33 @@ it does not matter who @reachin{publish}es, such as in a @reachin{timeout}.
...
+
+However, to determine variable usage in strict mode, Reach traverses every conditional branch,
+whether or not it is the chosen one. In doing so, Reach will
+raise an error when reviewing the true branch of the ternary:
```
jeapostrophe, line 2326: You have exposed the implementation reason for this behavior. In contrast, I'd like to think that `'use strict'` exists to "be strict" and we just discovered that it was easy to do this. So I'd say something like, "But, in `use strict` mode, Reach will ensure that this program treats `o` as having a single type, and detect an error in the program as follows:"

docs-src/ref-programs.scrbl
```diff
@@ -2297,3 +2297,33 @@ it does not matter who @reachin{publish}es, such as in a @reachin{timeout}.
...
+
+@verbatim{
+  reachc: error: Invalid field access. Expected object, got: Bool }
```
jeapostrophe, line 2329: "The correct way to write a program like this in strict mode, is to use `Maybe`. Like this: ...."

I think you should `@deftech{strict mode}`

hs/src/Reach/AST/SL.hs
```diff
@@ -97,7 +97,7 @@ st2it t = case t of
 type SLPartEnvs = M.Map SLPart SLEnv
 
 data SLCloEnv
-  = SLCloEnv SLPartEnvs SLEnv
+  = SLCloEnv SLPartEnvs SLEnv Bool
```
jeapostrophe, line 100: I think it is worth defining this as a record with names

hs/src/Reach/Eval.hs
```diff
@@ -115,13 +116,15 @@ compileDApp cns (SLV_Prim (SLPrim_App_Delay at opts part_ios top_formals top_s t
         env0 <- locAt slcpi_at $ env_insert "interact" iov top_env_wps
         return $ (slcpi_who, env0)
   penvs <- M.fromList <$> mapM make_penvp part_ios
+  use_strict <- liftIO . dupeIORef =<< asks (sco_use_strict . e_sco)
```
jeapostrophe, line 119: Can you put `dupeIORef` in `Reach.Util` and get rid of all the duplicate definitions of it?

hs/test-examples/features/use_strict.txt
```diff
@@ -0,0 +1,7 @@
+Verifying knowledge assertions
+Verifying for generic connector
+  Verifying when ALL participants are honest
+  Verifying when NO participants are honest
+  Verifying when ONLY "A" is honest
+Checked 5 theorems; No failures!
+WARNING: unused variable: x at ./use_strict.rsh:11:15:id
```
jeapostrophe, line 7: Woooaaaaah! But it *is* used! What's going on here?

hs/src/Reach/Eval/Core.hs
```diff
@@ -310,15 +321,45 @@ env_merge_ imode left righte = foldlM (env_insertp_ imode) left $ M.toList right
...
+trackVariable :: (SrcLoc, SLVar) -> App ()
+trackVariable el =
+  whenUsingStrict $ do
+    unused_vars <- asks e_unused_variables
+    unless (shouldNotTrackVariable el) $
+      liftIO $ modifyIORef unused_vars (el :)
+
+markVarUsed :: (SrcLoc, SLVar) -> App ()
+markVarUsed v =
+  whenUsingStrict $ do
```
jeapostrophe, line 347: I think this should be removed... we can always mark them used... it's generating the warning that we don't care about

hs/src/Reach/Eval/Core.hs
```diff
@@ -310,15 +321,45 @@ env_merge_ imode left righte = foldlM (env_insertp_ imode) left $ M.toList right
...
+    unused_vars <- asks e_unused_variables
+    liftIO $ modifyIORef unused_vars $ filter (v /=)
```
jeapostrophe, line 349: I think the data type would be more efficient if it were a set. You have this linear rather than logarithmic

hs/src/Reach/Eval/Core.hs
```diff
@@ -3275,6 +3323,11 @@ evalStmt = \case
   (s@(JSBreak a _ _) : _) -> illegal a s "break"
   (s@(JSLet a _ _) : _) -> illegal a s "let"
   (s@(JSClass a _ _ _ _ _ _) : _) -> illegal a s "class"
+  (JSExpressionStatement (JSStringLiteral a hs) sp) : ks
+    | trimQuotes hs == "use strict" -> do
+      use_strict <- asks $ sco_use_strict . e_sco
+      liftIO $ writeIORef use_strict True
```
jeapostrophe, line 3329: Hmm... this seems dangerous. From your docs, I assumed `use_strict` would be something that you would `local (\e -> e { sco = old sco { sco_use_strict = True } }` are the `evalStmt ks`. By making it a mutation... doesn't this mean that something like:

```
const f = (x, y) => y;
'use strict';
f(1, 2);
```

will error, because the body of the closure doesn't get evaluated until after the modification has happened?

hs/src/Reach/Eval/Core.hs
```diff
@@ -3613,6 +3668,13 @@ data MapEnv = MapEnv
   , me_ms :: IORef (M.Map DLMVar DLMapInfo)
   }
 
+dupeIORef :: IORef a -> IO (IORef a)
+dupeIORef r = newIORef =<< readIORef r
```
jeapostrophe, line 3672: This is the place that would go in `Reach.Util`

jeapostrophe (changes requested):

hs/test-examples/features/use_strict.txt
chrisnevers: This example is a bit contrived since I think `use strict` will almost always be used at the top level. But, it errors due to the scope sensitivity of `use strict`. `x` is published outside of the scope `use strict` was declared in (`A.only`)

hs/src/Reach/Eval/Core.hs
chrisnevers: In doing so, I believe this will cause inner scopes to not behave as I intended. 

```javascript
A.only(() => {
  'use strict'
  const x = 5;
})
A.publish(x);
```
I would expect `x` to be unused within the scope that `use strict` was declared in. This change will mark `x` as seen although I wanted to `use strict` for only inside the thunk.

hs/src/Reach/Eval/Core.hs
chrisnevers: That code segment will not error because the closure was declared in non-strict mode. When we create the `SLV_Clo` and convert sco to cloenv, we will copy what `use_strict` was set to at the time of declaration and use that when evaluating `SLV_Clo`.

docs-src/ref-programs.scrbl
chrisnevers: Personally, if my program will do what I want, despite a little bloat, I'd like it to compile 🙂 but I don't feel super strong about it, I can make it error if you want

jeapostrophe: I think I'd like error

hs/test-examples/features/use_strict.txt
jeapostrophe: I think that `'use strict'` is about monitoring the variables, not about monitoring the uses.

hs/src/Reach/Eval/Core.hs
jeapostrophe: Yea, I think we disagree about what it means

hs/src/Reach/Eval/Core.hs
jeapostrophe: How about this then:

```
switch (x) {
 case Var1: {
  'use strict';
  const y1 = 1;
  return y1 + 1; }
 case Var2: {
  const y2 = 2;
  return 2; }
}
```

I believe that `y2` will be protected when it should not, because each case shares the same scope

hs/src/Reach/Eval/Core.hs
jeapostrophe: My `IORef` idea would be even more efficient, because the environment would just also store a `Maybe (IORef Bool)` and then just set it to false and then the unused thing would run the ones set to true.

hs/src/Reach/Eval/Core.hs
chrisnevers: That example compiles fine, the strict mode currently follows "syntactic scope" if that makes sense as opposed to how our evaluation which might use the same env for different blocks/continuations. It would treat the two `case` blocks as adjacent scopes as opposed to one scope

hs/src/Reach/Eval/Core.hs
chrisnevers: Removed the `whenUsingStrict` check on `markVarUsed` so this now behaves as you intended

hs/src/Reach/Eval/Core.hs
jeapostrophe: I'm scared Chris. Mutation is scary. Does it not work if it is `local`?

jeapostrophe:
Does this program error?

`f.rsh`:
```
'reach 0.1';
'use strict';

const x = 1;
export const y = 2;
```

`g.rsh`:
```
'reach 0.1';
import * from 'f.rsh';

export const main = .... y ....;
```

because `x` is not used or exported in`f.rsh`?

Also, if I removed the reference to `y` in `g.rsh`, and made `g.rsh` in strict mode, would it error that `y` wasn't used?

chrisnevers:
@jeapostrophe `import * from 'f.rsh'` isn't proper syntax. But, using the follow syntaxes both give the `unused variable: x` errors:
* `import 'f.rsh'`
* `import * as f from 'f.rsh'`

because `x` is not used within `f`, nor exported. Anything that is exported from a module will be considered "used" from the viewpoint of that module.

# 

If you removed the reference to `y` in `g.rsh` and used strict mode it would give the following errors:
* `import 'f.rsh'` : unused var `x, y`
* `import * as f from 'f.rsh'`: unused var `f, x`

because anything imported into a module should be used

jeapostrophe:
Excellent

hs/src/Reach/Eval.hs
```diff
@@ -202,5 +200,6 @@ compileBundle cns jsb main = do
   ms' <- readIORef me_ms
   final <- readIORef e_lifts
   unused_vars <- readIORef e_unused_variables
-  forM_ unused_vars $  emitWarning . uncurry W_UnusedVariable
+  unless (null unused_vars) $
+    error $ show $ Err_Unused_Variables $ S.toList unused_vars
```
jeapostrophe, line 204: This should use `expect_throw`

jeapostrophe (approved):

PR #83, opened 2021-03-18
Update `fork` To Support Multiple Cases For Each Participant

chrisnevers:
If a participant has one case in a `fork`, essentially nothing changed. If a participant has multiple cases in a `fork`:

Generate a `Data` def representing the possible message types:

```javascript
const AMsgType = Data({ A0: a, A1: b, ... })
```

Update the `only` that processes `before`s to be:

```javascript
A.only(() => {
  const runA0 = () => {
    const res = before0();
    const msg = res.msg || null;
    const when = res.when || true;
    return { ... res, when, msg: AMsgType.A0(msg) };
  }
  // ...

  const case_res0 = runA0();
  const case_res1 = case_res0.when ? case_res0 : runA1();
  const res = case_resN_1.when ? case_resN_1 : runAN();

  const fork_msg = res.msg;
  const fork_when = res.when;
})
```

Updates `case A` of `race.pay` to match on `AMsgType` and pay the appropriate amount.

```javascript
race(A, B)
  .pay(fork_msg.match({
      A: (t) =>
        t.match({
          A0: () => payment,
          A1: () => payment * 2,
        }) }) )
```

Finally, when we select the appropriate body to run depending on who published first, match on `MsgType` to run the correct `after` thunk.

```javascript
switch (fork_msg) {
  case A: {
    assert(this == A);
    switch (fork_msg) {
      case A0: {
        commit();
        A.only(() => interact.show(1));
      }
      case A1: {
        commit();
        A.only(() => interact.show(2));
      } } } }
```


hs/src/Reach/Eval/Core.hs
```diff
@@ -3120,9 +3122,27 @@ data CompiledForkCase = CompiledForkCase
   , cfc_req_prop :: JSObjectProperty
   , cfc_switch_case :: JSSwitchParts
   , cfc_only :: JSStatement
+  , cfc_msg_type_def :: [JSStatement]
   }
 
-doFork :: [JSStatement] -> [(SrcLoc, (JSExpression, JSExpression, JSExpression, JSExpression))] -> Maybe (SrcLoc, [JSExpression]) -> App SLStmtRes
+forkCaseAt :: ForkCase -> SrcLoc
```
jeapostrophe, line 3128: This feels like you really wanted to define a record type with its own accessors

hs/src/Reach/Eval/Core.hs
```diff
@@ -3120,9 +3122,27 @@ data CompiledForkCase = CompiledForkCase
...
+forkCaseWho :: ForkCase -> JSExpression
+forkCaseWho (_, (who, _, _, _)) = who
+
+forkCaseSameParticipant :: ForkCase -> ForkCase -> Bool
+forkCaseSameParticipant l r = getWho l == getWho r
+  where
+    getWho e = jse_expect_id (forkCaseAt e) (forkCaseWho e)
+
+indexed :: [a] -> [(Int, a)]
+indexed = aux 0
```
jeapostrophe, line 3140: `zip [0..] l`?

chrisnevers: derp, I'll use that!

jeapostrophe (approved):

jeapostrophe:
Do the docs need to change?

chrisnevers:
@jeapostrophe yeah I just realized that, I'll make those changes now

jeapostrophe (approved):

PR #86, opened 2021-03-22
Use Strict Mode in Standard Library and Examples

chrisnevers:
Fixed an issue where a top-level `use strict` did not infect `main`. This issue was caused by returning `App_delay` when `main` is encountered in `evalLib`. I added another field to `Prim_AppDelay` to reflect whether the module is using strict mode.

Update the standard library and most examples to use strict mode. I did not use strict mode for the tutorials, examples that do not compile successfully or anything specifically relying on dynamic typing such as the trust fund workshop.

### DIFF
--- a/examples/overview/index.rsh
+++ b/examples/overview/index.rsh
@@ -1,4 +1,5 @@
 'reach 0.1';
+'use strict';
 
 export const main =
   Reach.App(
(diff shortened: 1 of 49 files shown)


examples/overview/index.rsh
```diff
@@ -1,4 +1,5 @@
 'reach 0.1';
+'use strict';
```
jeapostrophe, line 2: This program is referenced in `overview.scrbl`, so the line numbers need to be updated.

jeapostrophe (changes requested):

jeapostrophe:
Awesome!

jeapostrophe (approved):

PR #87, opened 2021-03-24
Export Values And Functions To Backend

chrisnevers:
Grab the export list when processing a module. Filter out anything that exports `Reach.App`. Export anything that is a `DLArg` or `DLLargeArg`. Pass a `[(SLVar, DLExportValue)]` throughout all the steps. Write a backend function that generates `getExports(stdlib) => { ... }`.

## To Do
* Check participant name collision with `getExports`
* Export functions from backend


js/stdlib/ts/ALGO.ts
```diff
@@ -1376,3 +1376,5 @@ export const verifyContract = async (ctcInfo: ContractInfo, backend: Backend): P
 
   return true;
 };
+
```
chrisnevers, line 1379: This is so the frontend can call 
```javascript
const stdlib = await loadStdlib();
const exports = getExports(stdlib.reachStdlib)
``` 
instead of passing an arbitrary ctc.

hs/src/Reach/AST/DK.hs
```diff
@@ -84,13 +84,16 @@ instance Pretty DKBlock where
   pretty (DKBlock _ _ k a) = prettyBlockP k a
 
 data DKProg
-  = DKProg SrcLoc DLOpts SLParts DLInit DKTail
+  = DKProg SrcLoc DLOpts SLParts DLInit [(SLVar, DLExportValue)] DKTail
```
jeapostrophe, line 87: I think you should make a type like `type DLinExports a = M.Map SLVar (DLinExportVal a)` where the `a` is the type of the function body, then make a family of those for `DLExports`, `DKExports`, and so on

hs/src/Reach/AST/DLBase.hs
```diff
@@ -258,6 +258,11 @@ data DLArgExpr
   | DLAE_Data (M.Map SLVar DLType) String DLArgExpr
   | DLAE_Struct [(SLVar, DLArgExpr)]
 
+data DLExportValue
+  = DLEV_Arg DLArg
+  | DLEV_LArg DLLargeArg
```
jeapostrophe, line 263: As discussed, we'll have `DLEV_Fun [DLVar] (DLinBlock a)` also

hs/src/Reach/Backend/JS.hs
```diff
@@ -709,7 +714,11 @@ jsPIProg cr (PLProg _ (PLOpts {}) dli (EPPs pm) _) = do
   partsp <- mapM (uncurry (jsPart dli)) $ M.toAscList pm
   cnpsp <- mapM (uncurry jsCnp) $ HM.toList cr
   let connsExp = jsConnsExp $ HM.keys cr
-  return $ vsep_with_blank $ preamble : emptyDoc : partsp ++ emptyDoc : cnpsp ++ [emptyDoc, connsExp, emptyDoc]
+  exportProps <- mapM (\ (k, v) -> do
+        vs <- jsExportValue v
+        return $ "    " <> pretty k <> " : " <> vs <> ",\n") dexports
+  let exportsp = "export const getExports = (stdlib) => {\n  return ({\n" <> hcat exportProps <> "  });\n};"
```
jeapostrophe, line 720: This is going to have to have its own "contract environment" like individual backends have

hs/src/Reach/Backend/JS.hs
```diff
@@ -698,8 +698,13 @@ jsConnsExp names = "export const _Connectors" <+> "=" <+> jsObject connMap <> se
   where
     connMap = M.fromList [(name, "_" <> pretty name) | name <- names]
 
+jsExportValue :: DLExportValue -> App Doc
+jsExportValue = \case
+  DLEV_Arg a -> jsArg a
+  DLEV_LArg a -> jsLargeArg a
```
jeapostrophe, line 704: Functions will have to `jsProtect` their arguments based on the types of the corresponding `DLVar`

hs/src/Reach/EraseLogic.hs
```diff
@@ -165,8 +165,8 @@ instance Erase LLStep where
       return $ LLS_ToConsensus at send' recv' mtime'
 
 instance Erase LLProg where
-  el (LLProg at llo ps dli s) =
-    LLProg at llo ps dli <$> el s
+  el (LLProg at llo ps dli dex s) =
+    LLProg at llo ps dli dex <$> el s
```
jeapostrophe, line 169: Erase logic should inspect the functions

hs/src/Reach/Eval.hs
```diff
@@ -143,21 +143,31 @@ compileDApp cns (SLV_Prim (SLPrim_App_Delay at opts part_ios top_formals top_s (
...
-compileDApp _ topv =
+     in DLProg at dlo sps dli dlExports final
+compileDApp _ _ topv =
   expect_t topv $ Err_Top_NotApp
 
+getExports :: SLLibs -> [(SLVar, SLSSVal)]
+getExports =
+  M.foldr' (\ lib exs ->
+    M.toList (M.filter (\case
+      SLSSVal { sss_val = SLV_Prim SLPrim_App_Delay {} } -> False
```
jeapostrophe, line 157: I think this should just use `slToDLExportVal`, which would be partial

hs/src/Reach/Eval/Core.hs
```diff
@@ -526,6 +527,61 @@ slToDL = \case
     yes = return . Just
     no = return Nothing
 
+getExportValArg :: DLExportValue -> Maybe DLArg
+getExportValArg (DLEV_Arg a) = Just a
+getExportValArg _ = Nothing
+
+slToDLExportVal :: SLVal -> App (Maybe DLExportValue)
+slToDLExportVal = \case
```
jeapostrophe, line 535: The code you have duplicates a lot of the existing `slToDl` code and should not.

The function version would need to actually evaluate and collect the `DLStmts`

hs/src/Reach/Linearize.hs
```diff
@@ -152,10 +152,10 @@ dk_ at = \case
   s Seq.:<| ks -> dk1 at ks s
 
 dekont :: DLProg -> IO DKProg
-dekont (DLProg at opts sps dli ss) = do
+dekont (DLProg at opts sps dli dex ss) = do
   let eRets = mempty
   flip runReaderT (DKEnv {..}) $
-    DKProg at opts sps dli <$> dk_ at ss
+    DKProg at opts sps dli dex <$> dk_ at ss
```
jeapostrophe, line 158: It will need to dekont the funs

hs/src/Reach/Linearize.hs
```diff
@@ -251,10 +251,10 @@ instance LiftCon DKTail where
     DK_Continue at asn -> return $ DK_Continue at asn
 
 liftcon :: DKProg -> IO DKProg
-liftcon (DKProg at opts sps dli k) = do
+liftcon (DKProg at opts sps dli dex k) = do
   let eLifts = Nothing
   flip runReaderT (LCEnv {..}) $
-    DKProg at opts sps dli <$> lc k
+    DKProg at opts sps dli dex <$> lc k
```
jeapostrophe, line 257: This too, although it would never apply, because they'd be in SLM_Module

hs/src/Reach/Linearize.hs
```diff
@@ -415,7 +415,7 @@ df_step = \case
   x -> df_com LLS_Com df_step x
 
 defluid :: DKProg -> IO LLProg
-defluid (DKProg at (DLOpts {..}) sps dli k) = do
+defluid (DKProg at (DLOpts {..}) sps dli dex k) = do
```
jeapostrophe, line 418: Ditto

hs/src/Reach/Optimize.hs
```diff
@@ -311,13 +311,13 @@ instance Optimize DLInit where
...
 instance Optimize LLProg where
-  opt (LLProg at opts ps dli s) = do
+  opt (LLProg at opts ps dli dex s) = do
     let SLParts m = ps
     let psl = M.keys m
     env0 <- liftIO $ mkEnv0 psl
     local (\_ -> env0) $
       focusa $
-        LLProg at opts ps <$> opt dli <*> opt s
+        LLProg at opts ps <$> opt dli <*> pure dex <*> opt s
```
jeapostrophe, line 320: We'd need to optimize them

hs/src/Reach/UnrollLoops.hs
```diff
@@ -179,8 +179,8 @@ instance Unroll LLStep where
       LLS_ToConsensus at send <$> ul recv <*> ul mtime
 
 instance Unroll LLProg where
-  ul (LLProg at opts ps dli s) =
-    LLProg at opts ps dli <$> ul s
+  ul (LLProg at opts ps dli dex s) =
```
jeapostrophe, line 182: And unroll them, because we're going to verify them and the verifier can't deal with loops

hs/src/Reach/Verify/Knowledge.hs
```diff
@@ -330,7 +330,7 @@ kgq_pie ctxt who (InteractEnv m) =
     >> (mapM_ (kgq_pie1 ctxt who) $ M.keys m)
 
 kgq_lp :: Maybe Handle -> VerifySt -> LLProg -> IO ()
-kgq_lp mh vst (LLProg _ (LLOpts {}) (SLParts psm) _dli s) = do
+kgq_lp mh vst (LLProg _ (LLOpts {}) (SLParts psm) _dli _ s) = do
```
jeapostrophe, line 333: It is safe to ignore because they can't receive secrets

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1444,7 +1444,7 @@ _verify_smt mc ctxt_vst smt lp = do
         case mc of
           Just c -> smt_lt at_de $ conCons c cn
           Nothing -> Atom $ smtConstant cn
-  let LLProg at (LLOpts {..}) (SLParts pies_m) (DLInit {..}) s = lp
+  let LLProg at (LLOpts {..}) (SLParts pies_m) (DLInit {..}) _ s = lp
```
jeapostrophe, line 1447: We need to verify the functions, because they might contain `assert`

js/stdlib/ts/ALGO.ts
jeapostrophe: I think the interface should be `getExport(stdlib)` and it will know to look at `stdlib.reachStdLib`

hs/src/Reach/AST/PL.hs
```diff
@@ -258,14 +258,16 @@ instance Pretty a => Pretty (CHandlers_ a) where
   pretty (CHandlers m) =
     render_obj m
 
+type CCExports a = DLinExports (DLinBlock a)
+
 data CPProg a
-  = CPProg SrcLoc (CHandlers_ a)
+  = CPProg SrcLoc (CCExports a) (CHandlers_ a)
```
jeapostrophe, line 264: This should not be part of `CPProg`, it should be part of `PLProg`

hs/src/Reach/AST/SL.hs
```diff
@@ -163,6 +167,7 @@ instance Pretty SLVal where
     SLV_Object _ (Just lab) _ -> pretty lab
     SLV_Object _ _ m -> render_obj m
     SLV_Clo {} -> "<closure>"
+    SLV_CloTyped {} -> "<closure_typed>"
```
jeapostrophe, line 170: This is a nit, but I'd go with `typed closure`

hs/src/Reach/AddCounts.hs
```diff
@@ -117,10 +118,20 @@ ac_top x = do
...
+
 add_counts :: PIProg -> IO PLProg
 add_counts (PLProg at plo dli _epps cp) = do
   let epps' = EPPs mempty -- XXX hack
-  let CPProg cat (CHandlers chs) = cp
+  let CPProg cat dex (CHandlers chs) = cp
   chs' <- mapM ac_top chs
-  let cp' = CPProg cat $ CHandlers chs'
+  dex' <- mapMapM ac_ev dex
+  let cp' = CPProg cat dex' $ CHandlers chs'
```
jeapostrophe, line 136: As mentioned, this is in the wrong spot

hs/src/Reach/AddCounts.hs
```diff
@@ -117,10 +118,20 @@ ac_top x = do
   e_cs <- newIORef $ mempty
   flip runReaderT (Env {..}) $ ac_ch x
 
+ac_ev :: DLinExportVal PILBlock -> IO (DLinExportVal PLBlock)
+ac_ev (DLEV_Arg a)   = return $ DLEV_Arg a
+ac_ev (DLEV_LArg a)  = return $ DLEV_LArg a
+ac_ev (DLEV_Fun a b) = do
```
jeapostrophe, line 124: I prefer using `\case` to defining the function in multiple equations

hs/src/Reach/Backend/JS.hs
```diff
@@ -698,8 +708,48 @@ jsConnsExp names = "export const _Connectors" <+> "=" <+> jsObject connMap <> se
   where
     connMap = M.fromList [(name, "_" <> pretty name) | name <- names]
 
+jsExportValue :: DLinExportVal PILBlock -> App Doc
+jsExportValue = \case
+  DLEV_Arg a  -> do
+    let t = argTypeOf a
+    a' <- jsArg a
+    jsProtect "null" t a'
```
jeapostrophe, line 716: You don't need to protect, because this is a positive position

hs/src/Reach/Backend/JS.hs
```diff
@@ -698,8 +708,48 @@ jsConnsExp names = "export const _Connectors" <+> "=" <+> jsObject connMap <> se
...
+  DLEV_LArg a -> do
+    let t = largeArgTypeOf a
+    a' <- jsLargeArg a
+    jsProtect "null" t a'
```
jeapostrophe, line 720: Ditto

hs/src/Reach/EPP.hs
```diff
@@ -559,8 +559,20 @@ be_s = \case
           return $ ET_ToConsensus at from_v prev last_time_mv which mfrom msg_vs out_vs amt_v time_v mtime' ok_l'
     return $ ok_l''m
 
+
+mkexports :: DLinExportVal LLBlock -> BApp (DLinExportVal PILBlock)
+mkexports (DLEV_Fun args (DLinBlock at sf ll a)) = do
```
jeapostrophe, line 564: Another thing that should be `\case`

hs/src/Reach/EPP.hs
```diff
@@ -559,8 +559,20 @@ be_s = \case
...
+  let process = \case
+        DT_Return rat -> return $ return $ DT_Return rat
+        DT_Com m k -> return $ return $ DT_Com m k
```
jeapostrophe, line 567: This is `dtReplace` I believe

hs/src/Reach/EraseLogic.hs
```diff
@@ -164,9 +164,20 @@ instance Erase LLStep where
...
 
+instance Erase (DLinExportVal LLBlock) where
+  el = \case
+    DLEV_Fun args body ->
+      DLEV_Fun <$> mapM el args <*> el body
+    DLEV_Arg a  -> return $ DLEV_Arg a
+    DLEV_LArg a -> return $ DLEV_LArg a
+
+instance Erase LLExports where
+  el dex = sequence $ M.map el dex
```
jeapostrophe, line 175: I think this is `mapM`

hs/src/Reach/EraseLogic.hs
```diff
@@ -164,9 +164,20 @@ instance Erase LLStep where
...
+
 instance Erase LLProg where
-  el (LLProg at llo ps dli s) =
-    LLProg at llo ps dli <$> el s
+  el (LLProg at llo ps dli dex s) = do
+    dex' <- el dex
+    LLProg at llo ps dli dex' <$> el s
```
jeapostrophe, line 180: `LLProg at llo ps dli <$> el dex <*> el s`

jeapostrophe: No `do`

hs/src/Reach/Eval/Core.hs
jeapostrophe: This is still true

hs/src/Reach/Eval/Core.hs
```diff
@@ -2270,88 +2353,93 @@ instDefaultArgs env err formals = \case
...
-                            Just ri -> M.singleton ri (rlifts, rda)
-                    return $ (retsm, rty)
-              (retsms, tys) <- unzip <$> mapM go rs
-              let retsm = mconcat retsms
-              r_ty <- locAt body_at $ typeMeets_d tys
-              let dv = DLVar body_at Nothing r_ty ret
-              saveLift $ DLS_Prompt body_at (Right (dv, retsm)) body_lifts
-              return $ SLAppRes clo_sco'' (lvl, (SLV_DLVar dv))
+    SLV_Clo clo_at sc -> evalApplyClosureVals clo_at sc randvs
+    SLV_CloTyped clo_at sc _ -> evalApplyClosureVals clo_at sc randvs
```
jeapostrophe, line 2434: I think that this should apply the refinements

jeapostrophe: If they are present and potentially check the types, although not call with `DLVar`s

hs/src/Reach/Linearize.hs
```diff
@@ -151,11 +151,21 @@ dk_ at = \case
...
-dekont (DLProg at opts sps dli ss) = do
+runDk :: ReaderT DKEnv m a -> m a
+runDk = do
   let eRets = mempty
-  flip runReaderT (DKEnv {..}) $
-    DKProg at opts sps dli <$> dk_ at ss
+  flip runReaderT (DKEnv {..})
+
+dk_export :: SrcLoc -> DLExportVal -> IO (DLinExportVal DKBlock)
+dk_export at (DLEV_Fun args body) =
```
jeapostrophe, line 160: `\case`

hs/src/Reach/Linearize.hs
```diff
@@ -414,8 +424,14 @@ df_step = \case
     return $ LLS_ToConsensus at send recv' mtime'
   x -> df_com LLS_Com df_step x
 
+
+df_export :: DLinExportVal DKBlock -> DFApp (DLinExportVal LLBlock)
+df_export (DLEV_Fun args body) = DLEV_Fun args <$> df_bl body
```
jeapostrophe, line 429: `\case`

hs/src/Reach/Optimize.hs
```diff
@@ -258,6 +259,14 @@ instance {-# OVERLAPPING #-} (Eq a, Sanitize a, Extract a) => Optimize (DLinBloc
   opt (DLinBlock at fs b a) =
     newScope $ DLinBlock at fs <$> opt b <*> opt a
 
+instance {-# OVERLAPPING #-} Optimize (DLinExportVal LLBlock) where
+  opt (DLEV_Arg a)   = DLEV_Arg   <$> opt a
```
jeapostrophe, line 263: `\case`

hs/src/Reach/Util.hs
```diff
@@ -87,3 +91,12 @@ dupeIORef r = newIORef =<< readIORef r
 
 mapWithKeyM :: (Ord k, Monad m) => (k -> a -> m b) -> M.Map k a -> m (M.Map k b)
 mapWithKeyM f m = M.fromList <$> (mapM (\(k, x) -> (,) k <$> f k x) $ M.toAscList m)
+
+mapValM :: (Traversable t1, Monad m) => (t2 -> m b) -> t1 (a, t2) -> m (t1 (a, b))
+mapValM f = mapM (\ (k, v) -> do { v' <- f v; return (k, v')})
+
+mapMapM :: (Ord k, Monad f) => (t -> f a) -> M.Map k t -> f (M.Map k a)
+mapMapM f m = M.fromList <$> mapValM f (M.toList m)
```
jeapostrophe, line 99: I think this is just `mapM`... why do you need more?

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
chrisnevers: I tried to refactor but I think the way the data is structured makes coalescing these functions much more complex to understand and work with . 

`DLExportVal` has the structure `DLArg | DLLargeArg` where `DLLargeArgs` contain `DLArgs`. Meanwhile, `DLArgExpr` has a flattened structure `DLArg | ...DLLargeArg`, where the constructors analogous to large args contain `DLArgExpr`. The result was just really messy and full of conditionals on whether it's producing an `ExportVal` or `DLArgExpr`

hs/src/Reach/Util.hs
```diff
@@ -87,3 +90,9 @@ dupeIORef r = newIORef =<< readIORef r
 
 mapWithKeyM :: (Ord k, Monad m) => (k -> a -> m b) -> M.Map k a -> m (M.Map k b)
 mapWithKeyM f m = M.fromList <$> (mapM (\(k, x) -> (,) k <$> f k x) $ M.toAscList m)
+
+mapValM :: (Traversable t1, Monad m) => (t2 -> m b) -> t1 (a, t2) -> m (t1 (a, b))
+mapValM f = mapM (\ (k, v) -> do { v' <- f v; return (k, v')})
```
jeapostrophe, line 95: Is this used... isn't it also just `mapM` at the uses?

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1429,6 +1431,14 @@ _smtDefineTypes smt ts = do
   mapM_ type_name ts
   readIORef tmr
 
+smt_export :: DLinExportVal LLBlock -> App SExpr
+smt_export = \case
+  DLEV_Arg at a  -> smt_a at a
+  DLEV_LArg at a -> smt_la at a
+  DLEV_Fun at args body -> do
+    forM_ args $ flip (pathAddUnbound at) O_Export . Just
+    smt_block body
```
jeapostrophe, line 1440: As mentioned, this will fail if there are loops, because you haven't unrolled them

hs/src/Reach/Eval/Core.hs
jeapostrophe: Right now, there's

```
slToDL :: SLVal -> App (Maybe DLArgExpr)
```

I think there needs to be

```
newThing :: SLVal -> App (Maybe CoolNewThing)
cntToDL :: CoolNewThing -> App (Maybe DLArgExpr)
cntToEV :: CoolNewThing -> App (Maybe DLExportVal)
```

And maybe `CoolNewThing` just IS `DLArgExpr` and and `cntToDL` is really just `compileDLArgExpr`

jeapostrophe: Bottom line: We can't have this duplication.

jeapostrophe: If you want, I can experiment once you prototype

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
```diff
@@ -480,91 +480,67 @@ typeMeets_d = \case
...
+  DLV_Data at t s mdv -> DLEV_LArg at . DLLA_Data t s <$> rec mdv
+  DLV_Struct at mdvs -> return $ DLEV_LArg at $ DLLA_Struct $ mapMaybe recAssoc mdvs
+  where
+    rec = maybe Nothing getDLVArg
+    recs = mapMaybe rec
+    recAssoc (k, v) = (,) k <$> rec v
+
+
+getDLVArg :: DLValue -> Maybe DLArg
+getDLVArg (DLV_Arg _ a) = Just a
```
jeapostrophe, line 518: `\case`

hs/src/Reach/AST/DL.hs
```diff
@@ -183,6 +183,15 @@ data DLBlock
 instance Pretty DLBlock where
   pretty (DLBlock _at _ ss da) = prettyBlock (render_dls ss) da
 
+data DLValue
```
jeapostrophe, line 186: I think this and `DLArgExpr` are only used by `Core`, so maybe it should be in `Eval/Types`?

hs/src/Reach/EraseLogic.hs
```diff
@@ -171,6 +171,10 @@ instance Erase (DLinExportVal LLBlock) where
     DLEV_Arg at a  -> return $ DLEV_Arg at a
     DLEV_LArg at a -> return $ DLEV_LArg at a
 
+instance Erase (DLExportinBlock LLVar) where
+  el = \case
+    DLExportinBlock s r -> DLExportinBlock s <$> el r
```
jeapostrophe, line 176: Isn't this `s` a tail, so it needs to be analyzed?

hs/src/Reach/Eval/Core.hs
```diff
@@ -492,31 +492,75 @@ dlvToDL = \case
...
-  DLV_Fun at args b -> return $ DLEV_Fun at args b
-  DLV_Array at dt mdvs -> return $ DLEV_LArg at $ DLLA_Array dt (recs mdvs)
-  DLV_Tuple at mdvs -> return $ DLEV_LArg at $ DLLA_Tuple (recs mdvs)
-  DLV_Obj at menv ->
-    return $ DLEV_LArg at $ DLLA_Obj $ M.fromList $ mapMaybe recAssoc $ M.toList menv
-  DLV_Data at t s mdv -> DLEV_LArg at . DLLA_Data t s <$> rec mdv
-  DLV_Struct at mdvs -> return $ DLEV_LArg at $ DLLA_Struct $ mapMaybe recAssoc mdvs
+  DLV_Arg at a -> return $ DLExportBlock Seq.empty $ DLEV_Arg at a
+  DLV_Fun at a b -> return $ DLExportBlock Seq.empty $ DLEV_Fun at a b
+  DLV_Array at dt mdvs -> locAt at $ do
```
jeapostrophe, line 505: This function now seems extremely similar to the `dlvToDL` composed with `compileDLArgExpr` (or whatever its called)

hs/src/Reach/Optimize.hs
```diff
@@ -264,6 +264,10 @@ instance {-# OVERLAPPING #-} Optimize a => Optimize (DLinExportVal a) where
     DLEV_LArg at a  -> DLEV_LArg at <$> opt a
     DLEV_Fun at a b -> DLEV_Fun at a <$> opt b
 
+instance {-# OVERLAPPING #-} (Eq a, Sanitize a, Extract a) => Optimize (DLExportinBlock a) where
+  opt = \case
+    DLExportinBlock b r -> DLExportinBlock <$> opt b <*> opt r
```
jeapostrophe, line 269: These are going to need new scopes so we don't rely on the evaluation order being anything in particular. If instead we went for there being one block, then it would not need to be separate scopes for each one

hs/src/Reach/Optimize.hs
```diff
@@ -264,6 +264,10 @@ instance {-# OVERLAPPING #-} Optimize a => Optimize (DLinExportVal a) where
...
+
 instance {-# OVERLAPPING #-} Optimize LLExports where
   opt = mapM opt
```
jeapostrophe, line 272: The separate scope would only go here if we did the one block method

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
```diff
@@ -485,82 +485,30 @@ slToDL v = slToDLV v <&> maybe Nothing dlvToDL
...
-    argEnv <- mapM collectRet $ M.fromList dev
-    let ty = T_Struct $ M.toList $ M.map argTypeOf argEnv
-    let le = DLE_LArg at $ DLLA_Obj argEnv
-    (ev, assign) <- mkAssign at ty le
-    let assigns = collectStmts $ map snd $ dev
-    let block = assigns Seq.|> assign
-    return $ DLExportBlock block ev
+  DLV_Arg at a   -> retMt $ DLEV_Arg at a
+  DLV_Fun at a b -> retMt $ DLEV_Fun at a b
+  ow ->
```
jeapostrophe, line 504: That's pretttttty

examples/exports/index.rsh
```diff
@@ -0,0 +1,20 @@
...
+  a: 5,
+  b: {
+    c: true,
+    d: [1, 2, [true, false]]
+  }
+}
+
+export const add1_impl = (x) => x + 1;
+
+export const add1= is(add1_impl, Fun([UInt], UInt));
```
jeapostrophe, line 15: Can you add an example that use `Refine` to impose constraints on both the input and the output and that uses `assert` inside?

examples/exports/index.mjs
```diff
@@ -0,0 +1,31 @@
...
+
+  const ctcAlice = alice.deploy(backend);
+  const ctcBob = bob.attach(backend, ctcAlice.getInfo());
+
+  console.log('Exports:');
+  const exports = backend.getExports(stdlib);
+  const a = exports.a;
+  console.log(`a: ${JSON.stringify(a)}`);
+  const o = exports.o;
+  console.log(`o: ${JSON.stringify(o)}`);
```
jeapostrophe, line 19: I think you can do `===` and check that it looks like what we know its supposed to be so this is really a test case

jeapostrophe (changes requested):

examples/exports/index.rsh
chrisnevers: added an example function that uses refine and assert `subMul2`

examples/exports/index.mjs
chrisnevers: I added an equality check for the exports/result of functions.

examples/exports/index.mjs
```diff
@@ -1,6 +1,18 @@
...
+  if (a instanceof Date && b instanceof Date)
+    return a.getTime() === b.getTime();
+  if (!a || !b || (typeof a !== 'object' && typeof b !== 'object'))
+    return a === b;
+  if (a.prototype !== b.prototype) return false;
+  let keys = Object.keys(a);
+  if (keys.length !== Object.keys(b).length) return false;
+  return keys.every(k => equals(a[k], b[k]));
+};
+
```
jeapostrophe, line 15: I am a monster that just compares the strings generated by `JSON.stringify`

examples/exports/index.rsh
```diff
@@ -10,11 +10,29 @@ export const o = {
...
+  (([x, y]) => x < y),
+  (([x, y], z) => x + y < z)
+);
+
+export const sumMul2 = is(sumMul2_impl, sumMul2_ty);
 
 export const main = Reach.App(
-  {}, [Participant('Alice', {}), Participant('Bob', {})], (Alice, Bob) => {
-    assert(add1(1) == 2);
+  { verifyPerConnector: true },
```
jeapostrophe, line 33: Why per?

hs/src/Reach/Eval/Core.hs
```diff
@@ -2413,9 +2413,11 @@ evalApplyVals rator randvs =
     SLV_CloTyped clo_at sc tf -> do
       at <- withAt id
       let isDLVar = \case { SLV_DLVar _ -> True; _ -> False } . snd
-      unless (all isDLVar randvs) $
-        void $ assertRefinedArgs randvs at tf
-      evalApplyClosureVals clo_at sc randvs
+      let ct = if all isDLVar randvs then CT_Assume True else CT_Assert
+      (dom_tupv, _) <- assertRefinedArgs ct randvs at tf
+      res@(SLAppRes _ (_, ret_v)) <- evalApplyClosureVals clo_at sc randvs
+      forM_ (stf_post tf) $ flip (applyRefinement ct) [dom_tupv, ret_v]
```
jeapostrophe, line 2419: This post is always `CT_Assert`

hs/src/Reach/Eval/Core.hs
```diff
@@ -2413,9 +2413,11 @@ evalApplyVals rator randvs =
     SLV_CloTyped clo_at sc tf -> do
       at <- withAt id
       let isDLVar = \case { SLV_DLVar _ -> True; _ -> False } . snd
-      unless (all isDLVar randvs) $
-        void $ assertRefinedArgs randvs at tf
-      evalApplyClosureVals clo_at sc randvs
+      let ct = if all isDLVar randvs then CT_Assume True else CT_Assert
```
jeapostrophe, line 2416: I don't get why this isn't always `CT_Assume True`?

chrisnevers: Typed closures can be applied within a Reach program, so it should assert the static arguments satisfy the pre condition instead of assuming

examples/exports/index.rsh
chrisnevers: residual from playing around, Ill remove

hs/src/Reach/Eval/Core.hs
jeapostrophe: This is not a good check for that, because I could apply the function in Reach and still send it all DLVars... like the result of a publish. So this should really be based on something else...

jeapostrophe: Easier test would be a `forall`, which returns a `DLVar`

chrisnevers: I don't understand using `forall` in this scenario?

jeapostrophe: `forall` returns a `DLVar`, although... I can't think of a way to actually use it usefully. Sorry

chrisnevers: I ended up using the mode to determine whether to use assume or assert. In `SLM_Module`, the only dynamic values that'll exist are arguments from the frontend to exported functions

hs/src/Reach/Eval/Core.hs
```diff
@@ -2412,11 +2412,11 @@ evalApplyVals rator randvs =
     SLV_Clo clo_at sc -> evalApplyClosureVals clo_at sc randvs
     SLV_CloTyped clo_at sc tf -> do
       at <- withAt id
-      let isDLVar = \case { SLV_DLVar _ -> True; _ -> False } . snd
-      let ct = if all isDLVar randvs then CT_Assume True else CT_Assert
+      m <- readSt st_mode
+      let ct = if m == SLM_Module then CT_Assume True else CT_Assert
```
jeapostrophe, line 2416: I feel like such a pain... but what about:

```
const f_impl = (x) => x + 1;
const f = is(f_impl, Refine(Fun([UInt], UInt), ([x] => x > 1), (_ _ => true)));
const x = f(0);
```

This should fail, because `0` is not `> 1` and this is at the module level, but I think your code will assume it rather than assert it

chrisnevers: oh duh, thanks for catching that. I should have checked both: `all isDLVar randvs && mode == SLM_Module`

jeapostrophe:
```
const y = forall(UInt);
const x = f(y);
````

This should also fail, because we can't prove that `y` is `> 1`

jeapostrophe: I think there needs to be "another way" to know that you are exporting. Like `SLM_Exporting` that gets switched to `SLM_Module` on the application or something like that (i.e. that only unravels once)

jeapostrophe: Another option would be for the exporting code to see through `evalApplyVals` and pass it an extra arg or something

chrisnevers: Okay everything works with using `SLM_Exporting` to define the exported functions and when applying a typed closure in `Exporting` mode, it switches to `SLM_Module` temporarily to statically check it's arguments. Pushing up..

jeapostrophe (approved):

PR #93, opened 2021-04-06
Trello 1159: RPC tutorial run instructions

mattaudesse:


jeapostrophe (changes requested):
Your additions are explaining the ins and outs, with useful examples. I believe it is trying to "explain THAT" the options work a certain way. The goal of the tutorial, though, is to "explain HOW" to make it run. The whole Reach documentation is inspired by a system that includes this section on tutorials: https://documentation.divio.com/tutorials/

Your additions never actually get to the "explain HOW" because it never says, "Type THIS into your computer and you will see it work." You can then have a link to the options page with a note that says something like, "We could have specified these arguments with a dictionary rather than in environment variables."

Next, the trace of messages is not useful to see. If it is still always prints these out, we should change it so that it doesn't or only does if `REACH_DEBUG` is enabled.

The goal of this addition is for people to see the exact same output as the bottom of https://docs.reach.sh/tut-7.html but replace `./reach run` with `./reach server; python the-python.py` or whatever the simple command line is.

I think you should ditch the scribble revisions and start fresh

PR #96, opened 2021-04-07
Have Frontend Provide Labels for JS Debugging

chrisnevers:
Have the frontend provide the labels to use for debugging instead of generating them from the Participant name. If a label is not provided it will default to the first 4 digits of the address

### DIFF
--- a/docs-src/ref-frontends-js.scrbl
+++ b/docs-src/ref-frontends-js.scrbl
@@ -95,7 +95,7 @@ These functions create and interact with @tech{account} representations.
 @(hrule)
 @(mint-define! '("getDefaultAccount"))
 @js{
- getDefaultAccount() => Promise<acc> }
+ getDefaultAccount(?label: string) => Promise<acc> }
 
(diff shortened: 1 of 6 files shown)


docs-src/ref-frontends-js.scrbl
```diff
@@ -105,38 +105,43 @@ When running in node.js while connected to one of reach's standard devnets,
 the default account will be connected to a faucet on the devnet.
 This promise will be rejected with an exception
 if no sensible default account can be accessed for the current context.
+A @tt{label} that distinguishes the account may optionally be provided to use in debug logs. If no label is provided, then the first 4 digits of the account address will be used.
```
jeapostrophe, line 108: four

jeapostrophe: See Strunk & White (spell out <100, numerals >100)

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #97, opened 2021-04-09
WIP: Exceptions

chrisnevers:
Add support for `try...catch/throw` statements


docs-src/ref-programs.scrbl
```diff
@@ -1194,6 +1194,20 @@ evaluates to @reachin{4}, but
 
 is erroneous, because the identifier @reachin{x} is not bound outside the @tech{block statement}.
 
+@subsubsection{Try...Catch statements}
```
jeapostrophe, line 1197: Let's stylize this a Try/Catch

jeapostrophe: Also, should include throw... so "Try/Catch & Throw Statements"?

docs-src/ref-programs.scrbl
```diff
@@ -1194,6 +1194,20 @@ evaluates to @reachin{4}, but
...
+
+@reach{
+  try {
+    throw 10;
+  } catch (v) {
+    transfer(v).to(A); }}
+
+A @deftech{try statement}, written @reachin{try BLOCK catch (VAR) BLOCK}, allows a block
+of code to execute with a specified handler should an exception be thrown. A @deftech{throw statement},
```
jeapostrophe, line 1206: Semantic newline

hs/src/Reach/AST/DL.hs
```diff
@@ -108,6 +110,8 @@ instance Pretty DLStmt where
       DLS_FluidRef _ dv fv ->
         pretty dv <+> "<-" <+> "fluid" <+> pretty fv
       DLS_MapReduce _ _mri ans x z b a f -> prettyReduce ans x z b a f
+      DLS_Throw _ dv -> "throw" <+> pretty dv
+      DLS_Try _ e hv hs -> "try" <+> render_dls e <+> "catch" <+> parens (pretty hv) <+> render_dls hs
```
jeapostrophe, line 114: I think you'll want to use `ns`

hs/src/Reach/AST/DL.hs
```diff
@@ -170,6 +178,8 @@ instance IsLocal DLStmt where
     DLS_FluidSet {} -> True
     DLS_FluidRef {} -> True
     DLS_MapReduce {} -> True
+    DLS_Throw {} -> True
+    DLS_Try {} -> True
```
jeapostrophe, line 182: It is local only if the body and handler are

hs/src/Reach/AST/DL.hs
```diff
@@ -151,6 +157,8 @@ instance IsPure DLStmt where
     DLS_FluidSet {} -> False
     DLS_FluidRef {} -> True
     DLS_MapReduce {} -> True
+    DLS_Throw {} -> False
+    DLS_Try {} -> True
```
jeapostrophe, line 161: Only if body and handler are

hs/src/Reach/Eval/Core.hs
```diff
@@ -3915,7 +3921,61 @@ evalStmt = \case
...
+    unless (e_exn_in_throw exn) $ expect_ Err_Throw_No_Catch
+    -- Set or ensure the type that this `try` block is expected to catch
+    case e_exn_ty exn of
+      Nothing -> liftIO $ modifyIORef exn_ref (\ ex -> ex { e_exn_ty = Just dv_ty })
+      Just ty
+        | ty /= dv_ty -> expect_ $ Err_Try_Type_Mismatch ty dv_ty
+        | otherwise   -> return ()
+    saveLift =<< withAt (`DLS_Throw` dv)
+    liftIO $ modifyIORef exn_ref (\ ex -> ex { e_exn_st = Just curSt })
+    st <- asks e_st
```
jeapostrophe, line 3939: This should be a mergeSt because you have to be compatible with all previous throws. Try an example where you have an if where one branch introduces a participant and the other doesn't but both throw

hs/src/Reach/Eval/Core.hs
```diff
@@ -3915,7 +3921,61 @@ evalStmt = \case
...
+      SLRes try_stmts try_st try_ret <- captureRes $ locTry $ evalStmt stmts
+      exn_env <- liftIO $ readIORef exn_env_ref
+      case exn_env of
+        -- Get the type of thrown/caught expression & the state during throw
+        ExnEnv { e_exn_ty = Just arg_ty, e_exn_st = Just thrown_st } -> do
+            -- Bind `catch` argument before evaluating handler
+            handler_arg <- ctxt_mkvar (DLVar at Nothing arg_ty)
+            handler_env <- evalDeclLHS True Public mempty (SLV_DLVar handler_arg) ce
+            sco <- asks e_sco
+            let sco_env' = handler_env `M.union` sco_cenv sco
```
jeapostrophe, line 3962: This should use `envMerge` which protects against double binding, but really the `mempty` on line 3960 should be `sco_cenv sco`

hs/src/Reach/Eval/Core.hs
```diff
@@ -3915,7 +3921,61 @@ evalStmt = \case
...
+            SLRes handler_stmts handler_st handler_ret <-
+              locSt thrown_st $
+                captureRes $
+                  locSco (sco { sco_cenv = sco_env' }) $
+                    evalStmt handler
+            -- Ensure the successful try block and catch block end with compatible states
+            setSt try_st
+            mergeSt handler_st
+            saveLift $ DLS_Try at try_stmts handler_arg handler_stmts
+            retSeqn handler_ret ks
```
jeapostrophe, line 3973: I believe you are losing the `try_ret`

hs/src/Reach/Linearize.hs
```diff
@@ -138,6 +143,16 @@ dk1 at_top ks s =
     DLS_FluidSet {} -> com
     DLS_FluidRef {} -> com
     DLS_MapReduce {} -> com
+    DLS_Throw at dv -> do
```
jeapostrophe, line 146: dv should be da

hs/test-examples/features/exception_return.rsh
```diff
@@ -0,0 +1,27 @@
...
+  }
+}
+
+export const main =
+  Reach.App(
+    {},
+    [Participant('A', {})],
+    (A) => {
+      assert(returnsFromCatch());
+      assert(returnsFromTry());
```
jeapostrophe, line 26: I think this only works because the try never fails. If you were to instead have written `try { if (x < x) { throw false; } else { return true } } catch (e) { return false }` where `x` was an interact variable, then I believe this assert would fail

jeapostrophe (changes requested):

chrisnevers:
As mentioned in slacked, I need to tag the `throw` with `local` depending on the mode still, but just pushing all the other stuff I have now

hs/src/Reach/Eval/Core.hs
```diff
@@ -3934,8 +3934,9 @@ evalStmt = \case
       Just ty
         | ty /= dv_ty -> expect_ $ Err_Try_Type_Mismatch ty dv_ty
         | otherwise   -> return ()
-    saveLift =<< withAt (`DLS_Throw` dv)
+    saveLift =<< withAt (\at -> DLS_Throw at dv $ not $ isConsensusStep $ st_mode curSt)
```
jeapostrophe, line 3937: I believe that we are interested in the state of the `try`, not the state of the `throw`, because the `try` is essentially going to turn into an `if`

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #100, opened 2021-04-12
Update tut.scrbl

hagenhaus:
Fixed typos.

### DIFF
--- a/docs-src/tut.scrbl
+++ b/docs-src/tut.scrbl
@@ -180,11 +180,11 @@ The first step is to change the Reach program to specify that Alice and Bob's fr
 
 @itemlist[
 
-@item{Lines 3 through 5 defines a @tech{participant interact interface} that will be shared between the two players.
+@item{Lines 3 through 5 define a @tech{participant interact interface} that will be shared between the two players.
 In this case, it provides two methods: @reachin{getHand}, which returns a number; and @reachin{seeOutcome}, which receives a number.}
 
 @item{Line 10 uses this interface for both participants.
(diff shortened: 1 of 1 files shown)


docs-src/tut.scrbl
```diff
@@ -1619,7 +1619,7 @@ We will use @link["https://reactjs.org/"]{React.js} for this tutorial, but the s
 @item{React programs are JavaScript programs that use a special library that allows you to mix HTML inside of the body of your JavaScript.}
 @item{React has a special compiler that combines a bundle of JavaScript programs, and all their dependencies, into one large file that can be deployed on a static Web server.
 This is called "packing".}
-@item{When you're developing and testing with React, you run a special development Web server that watches updates this packed file every time you modify your source files, so you don't have to constantly run the compiler.}
+@item{When you're developing and testing with React, you run a special development web server that watches and updates this packed file every time you modify a source file, so you don't have to constantly run the compiler.}
```
jeapostrophe, line 1622: Web is capitalized

jeapostrophe: If you wanted you could say HTTP server, but I don't like the techy sounding ness of that

jeapostrophe (changes requested):

docs-src/tut.scrbl
hagenhaus: [AP Style Book](https://www.apstylebook.com/) lowercased "internet" and "web" in 2016. I still see "Web" around here and there, but "web" is most common these days. I'll go with your preference. No worries.

jeapostrophe: Yea, I had this beaten in to me during my PhD in early 2000 and can't go back. I'll be writing `Web` and `Pluto` forever I think :)

PR #102, opened 2021-04-16
Create Interference Graphs For Each Variable Type

chrisnevers:
Update `CPProg` to maintain `Map`s that track: 
* `Type`   -> max color
* `DLVar` -> type specific color

### DIFF
--- a/hs/src/Reach/AST/PL.hs
+++ b/hs/src/Reach/AST/PL.hs
@@ -261,6 +261,18 @@ data ColorGraphs = ColorGraph {
   varGraph  :: M.Map DLVar Int }
   deriving (Eq)
 
+instance Pretty ColorGraphs where
+  pretty = \case
+    ColorGraph tg vg ->
+      "type graph:" <+> pretty tg <> hardline <>
+      "var graph:" <+> prettyVarGraph vg
+    where
+    prettyVarGraph g =
+      let rows = map (\ (v, i) ->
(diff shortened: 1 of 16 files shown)


hs/src/Reach/AST/PL.hs
```diff
@@ -256,14 +256,19 @@ instance Pretty a => Pretty (CHandlers_ a) where
...
     render_obj m
 
+data ColorGraphs = ColorGraph {
+  typeGraph :: M.Map DLType Int,
+  varGraph  :: M.Map DLVar Int }
+  deriving (Eq)
+
 data CPProg a
-  = CPProg SrcLoc (CHandlers_ a)
+  = CPProg SrcLoc ColorGraphs (CHandlers_ a)
```
jeapostrophe, line 265: I think the new analysis should be totally separate and just return this information

hs/src/Reach/Compiler.hs
```diff
@@ -71,9 +72,10 @@ compile copts = do
         showp "el" el
         pil <- epp el
         showp "pil" pil
-        let runConnector c = (,) (conName c) <$> conGen c woutnMay pil
+        cpil <- colorProgram pil
```
jeapostrophe, line 75: I think that we will optionally use this pass just in Sol

hs/src/Reach/AST/PL.hs
chrisnevers: Separated this out to just return the graphs from the function

PR #104, opened 2021-04-23
Allow Token Use in Remote Objects

chrisnevers:
## Additions
* Allow `remote.pay()` to accept an argument like `DLPayAmt` to specify the non-network tokens and amounts that Reach will `approve` the remote contract to withdraw. After the remote call completes, Reach will ensure the funds were transferred to the remote contract via `allowance`.
* Allow `remote.bill()` to accept an argument like `DLPayAmt` to specify which non-network tokens and amounts the Reach program will accept from the remote contract. The remote contract is responsible for transferring the funds to the Reach program.
* Allow `remote.withBill()` to accept a tuple of `Token` to specify the tokens (and ordering) in which we expect to receive some amount from the remote contract. The return result changed from a pair to a triple of shape 
    ```javascript
        [ netTokRecv, [ nonNetTokRecv+ ] | Null, callResult ]
    ```
    
Reach will ensure any tokens that are not specified in `withBill/bill` will have an unchanged balance after the remote call is completed.

## Need To Do
* Update docs
* Update tests


examples/remote/index.rsh
```diff
@@ -45,11 +49,45 @@ export const main = Reach.App(
     A.publish();
     const r0 = ctx.getX();
     const r1 = cty.getX.bill(amt / 2)();
-    const [ amtr, r2 ] = cty.getX.withBill()();
+    const [ amtr, _, r2 ] = cty.getX.withBill()();
```
jeapostrophe, line 52: I think it should only be a triple when a token is provided

hs/src/Reach/AST/DLBase.hs
```diff
@@ -398,7 +404,10 @@ instance Pretty DLExpr where
       DLE_MapSet _ mv kv nv ->
         pretty mv <> "[" <> pretty kv <> "]" <+> "=" <+> pretty nv
       DLE_MapDel _ mv i -> "delete" <+> pretty mv <> brackets (pretty i)
-      DLE_Remote _ _ av m amta as -> "remote(" <> pretty av <> ")." <> viaShow m <> ".pay" <> parens (pretty amta) <> parens (render_das as)
+      DLE_Remote _ _ av m amta as bill (DLWithBill _ nonNetTokRecv _) ->
```
jeapostrophe, line 407: I'm still reading, but I don't understand the `bill` field

jeapostrophe (changes requested):

hs/src/Reach/AST/DLBase.hs
chrisnevers: `bill` is the non-network amounts/tokens provided when the user calls `remote.bill( AMTS )`. Since the network amount is already handled, I just passed the non-network stuff along to solidity instead of passing a `DLPayAmt` object. 

`DLWithBill ` is used to represent the tokens we should/should not receive via `withBill`.

PR #106, opened 2021-04-26
NOT A REAL PR (yet): CFX ugly impl vs classy impl 

DanBurton:
**NOT A REAL PR (yet)**

About the new source files:

* CFX.ts is the "ugly" copy/pasted impl (deprecated)
* ETH_like.ts is the "classy" impl (proposed)
* cfxers.ts and cfxers_providers.ts are used for both impls
* classy_*.ts are used for the classy impl
* "classy" impl is currently "plugged in" to loader.ts
* haven't deleted any files that these things replace, yet

About the state of this implementation:

* both classy CFX and ETH impls can correctly run examples/overview
  - CFX (for now) still requires that you bring your own devnet & faucet
* some more impl for ETH still needs copying over to support everything ETH is supposed to support (at least: firstMsg deploy mode, browser support)
* some more impl for CFX still needs implementing in cfxers to support "real" CFX (at least: Provider on/off hooks for monitoring the passage of time with real delays on a "real" network)
* attempts to follow the philosophy of "avoid setBLAH, instead just create the thing with BLAH set as desired via constructor opts"
* attempts to follow the philosophy of "don't read env vars deep in the code, instead pass them in as config at the surface level"
* attempts to use `readonly` attributes as much as possible
* attempts to use `private` class methods and leading-underscore naming conventions for private methods, albeit not very consistently yet

### DIFF
--- a/js/stdlib/ts/loader.ts
+++ b/js/stdlib/ts/loader.ts
@@ -1,5 +1,6 @@
 import * as stdlib_ETH from './ETH';
 import * as stdlib_ALGO from './ALGO';
+import * as stdlib_CFX from './CFX';
 import {getConnectorMode, canonicalizeConnectorMode, getConnector} from './ConnectorMode';
 import {process, window} from './shim';
 
@@ -30,6 +31,7 @@ export async function loadStdlib(connectorModeOrEnv?: string | {[key: string]: s
   switch (connector) {
     case 'ETH': stdlib = stdlib_ETH; break;
     case 'ALGO': stdlib = stdlib_ALGO; break;
+    case 'CFX': stdlib = stdlib_CFX; break;
     default: throw Error(`impossible: unknown connector ${connector}`);
   }
   if (connectorModeOrEnv && typeof connectorModeOrEnv !== 'string') {
--- a/js/stdlib/ts/loader.ts
+++ b/js/stdlib/ts/loader.ts
@@ -1,4 +1,4 @@
-import * as stdlib_ETH from './ETH';
+// import * as stdlib_ETH from './ETH';
 import * as stdlib_ALGO from './ALGO';
 // import * as stdlib_CFX from './CFX';
 import * as stdlib_ETH_like from './ETH_like';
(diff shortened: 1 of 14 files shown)


js/stdlib/ts/loader.ts
```diff
@@ -8,10 +10,10 @@ export {getConnectorMode, getConnector};
 // XXX make an interface for Stdlib, return Promise<Stdlib>
 // The connectorMode arg is optional;
 // It will use REACH_CONNECTOR_MODE if 0 args.
-export async function loadStdlib(connectorModeOrEnv?: string | {[key: string]: string}): Promise<any> {
+export function loadStdlib(connectorModeOrEnv?: string | {[key: string]: string}): any {
```
jeapostrophe, line 13: You're removing `async` because it isn't actually used? I think it might be good to leave it in because in the future another network might need some sort of initialization? Is that wrong to think? Maybe because we can't ensure that if you /DON'T/ use the loader that it will be "inited" first?

jeapostrophe (commented):
Overall, I feel that this is a very "big" change and I feel like it would be easier to understand if it were more incremental, but probably the incremental version is how you got here. I feel like I want to DL and use `diff` to see the difference between the classy and non-classy versions, which Github doesn't provide

js/stdlib/ts/loader.ts
DanBurton: Removing `async` is something I want to do anyway, because it lets you do `const stdlib = loadStdlib()` at the top-level instead of having to be inside of an `async` function to call it. My opinion is that any sort of "async" initialization should be hidden from the user via using promises internal to the stdlib's implementaiton.

DanBurton: (Similar to how we made that change a while ago so that attach/deploy are no longer `async` functions, even though `deploy` does kick off an async deployment behind the scenes.)

DanBurton:
An idea for review:

Take a look at the `attach` method of the `EAcc` class, and notice how the values closed over get passed into the `ECtc` constructor, so that its methods have access to them.

Then go compare any given method of `ECtc` to the existing implementation of that method. Most (all?) of the method names are unchanged, except that some now have underscores in front. I wasn't entirely consistent on my styling. Sometimes I did `const {blah} = this` to unwrap the needed bindings, other times I switched to `this.blah`. (I'm paranoid about JavaScript methods on classes forgetting their `this`, so whenever `blah` is a function I tended to prefer `this.blah()`.)

js/stdlib/ts/loader.ts
jeapostrophe: Sounds great

PR #130, opened 2021-05-05
Improve `doParallelReduce` Compilation

chrisnevers:
Change `doParallelReduce` to inject the `continue` and it's corresponding assignments of the while loop inside of the `case` bodies of the generated `fork`


jeapostrophe:
Great stuff. I merged it myself because I was testing

PR #131, opened 2021-05-07
Compile All Exported `Reach.App`s by Default

chrisnevers:
I moved the `evalLib` code up so that we could discover the tops of a program, then process them sequentially. I re-use the same env that was used to parse the module to process each top.  And when processing a new top, I reset the `lifts` to those of the module, since it could have been altered when processing the previous top.


jeapostrophe (approved):

jeapostrophe:
Awesome

PR #132, opened 2021-05-07
Adding Default `Reach.App`

chrisnevers:
By default, if there are no exported `Reach.App`s, the compiler will add:

```javascript
export const default = Reach.App({}, [], () => {})
```

so the App can be verified and compiled for use as a library or `getExports`

### DIFF
--- a/hs/test-examples/features/lib_default_app.rsh
+++ b/hs/test-examples/features/lib_default_app.rsh
@@ -9,4 +9,3 @@ export const winner = (handA, handB) =>
 forall(UInt, handA =>
   forall(UInt, handB =>
     assert(isOutcome(winner(handA, handB)))));
-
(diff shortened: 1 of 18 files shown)


hs/test-examples/features/lib_default_app.rsh
```diff
@@ -0,0 +1,12 @@
+'reach 0.1';
+
+// Test assertions for libaries
+export const [ isOutcome, B_WINS, DRAW, A_WINS ] = makeEnum(3);
+
+export const winner = (handA, handB) =>
```
jeapostrophe, line 6: Can you go one step more and put this in `examples` and have a test case that actually uses these exports?

jeapostrophe (changes requested):
Documentation

jeapostrophe:
You are killing it

PR #136, opened 2021-05-13
Draft: Revamp reach app

chrisnevers:
## Revamp `Reach.App` to use a single thunk. 

Now, an application starts in the `SLM_AppInit` mode, where you may:
* Bind `AppArgs`, i.e. `Participant/ParticipantClass/View`, to identifiers.
* Call `setOptions` to change the app parameters such as `verifyOverflow` etc.
* Call `deploy()` to switch the mode to `SLM_Step`. 

### Changes to `Env`

Before, `compileDApp` could process all the `AppArgs` at once before the application starts, but now the state of the app can change during evaluation when new `AppArgs` are declared, so `e_ios`, `e_dlo`, `e_classes` are mutable. I created `e_pie` and `e_views`, which track the the participant interact environments and views, respectively. These values will be extracted after processing the `Reach.App` body and passed along to `DLProg`.

When processing multiple `Reach.App`s in one compilation, the driver will reset all the necessary `IORef`s from the last app in `Compiler::resetEnvRefs`

### Handling `AppArgs` Effects

Now, a `Reach.App` body will be processed with a clean state as opposed to knowing a bunch of information up front. All the logic that `compileDApp` used for validating `Views`, recording what `Participants` are classes etc. will occur while evaluating `AppArg` assignments in `SLM_AppInit`. This strategy allows for `AppArg`s to be declared in `SLM_Module` and have no effect on state until used within a `Reach.App`. 

An `SLV_AppArg SLA_Participant` will be unraveled to an `SLV_Participant` when assigned in `SLM_AppInit` and `SLV_AppArg SLA_View` will be converted to an `SLV_Object`, just like before.

### `setOptions`

A `Reach.App` will use the default app options from the start, until the user calls `setOptions` with their own object. All of these changes are pretty much moving around the existing code.

### Convert Deprecated `Reach.App`s

Ternary `Reach.App`s are converted to the new form, by generating:

```javascript
Reach.App(() => {
  setOptions(<user-options>);
  const <user-program-arg[0]> = <user-program-participant[0]>;
  ...;
  deploy();
  <user-program-body>
})
```

I updated `evalAppArg` used by the deprecated `Reach.App` into `verifyAppArg`, which will just check that everything is of the right type etc. This allows for better, more specific error messages than blindly converting the old form to the new form and getting poorer error messages later. 

I've made a deprecation warning for old forms, but do not have it enabled right now because of unit tests. I can add it in and accept the new unit test outputs, if desired.


jeapostrophe:
I've squashed this and am about to merge. Thanks!

PR #141, opened 2021-05-14
cfx-mv

DanBurton:
this is an incomplete PR that includes the "ugly" CFX impl, plus moving ETH -> ETH_like and ETH_compiled -> ETH_like_compiled.

This provides a useful "checkpoint" so that stuff changed from this point on doesn't have an insane diff.


DanBurton:
rebased and merged

PR #143, opened 2021-05-17
js: cfx beautiful impl

DanBurton:
* deleted ugly impl
* works w/ `simplified` branch of https://github.com/reach-sh/cfxers
* CFX now verifies some things about contract, but not ctc code, yet
* CFX T_Address backend representation is the "user-friendly" network-specific display, not the "0x" pubkey. We may be able to change this, but it would require changes to things like `cfxers.Wallet.sendTransaction` and possibly `T_Address.{,un}munge`. Unclear which is better but the 0x representation might be necessary at some point for full support of address-related Reach features. This has an impact on whether `digest(addr)` differs across networks for the same pubkey. Conflux people might argue that it should differ, because their whole thing with the new address format is to make it harder to use an addr on the "wrong" network by accident.
* I haven't tried more stuff yet but more stuff probably works on CFX.


jeapostrophe:
This looks good to me and I think that the address thing is fine.

PR #145, opened 2021-05-19
Refactor `instDefaultArgs` into `evalDeclLHS`

chrisnevers:
Add support for defaulting into `evalDeclLHS` and use `evalDeclLHS` for applying closure arguments


jeapostrophe (approved):
This is an awesome change. I love deleting code and getting more features

PR #146, opened 2021-05-20
Fixed ethers imports

burkongla:
I am not very familiar with your codebase, but had some issues with the npm package and saw that it is updated from here. I was able to fix it by manually changing imports in node_modules. I have attempted to fix those imports here. I also tried running the test, and as far as I could tell it went well.


esjay:
@jeapostrophe any sense about what is needed for this to PR to be approved? we came across this issue ourselves when attempting to use the reach lib with webpack/next.js.

DanBurton:
I'm taking a look at this and doing some testing today.

jeapostrophe:
We've done some more work on this. Something we've discovered is that Webpack 5 made a breaking change where it is no longer compatible with something that Webpack 4 allows and vice-versa. So basically, we're in a situation where Reach works with 4 and not 5. We're looking for a way to support both. As a short-term workaround, if it is possible for you to make progress with 4, I recommend doing that. We won't close this though, because we think it may be possible to work with both.

DanBurton:
Although I wasn't able to incorporate this particular PR into the changes I ultimately made to fix the issue, this effort was appreciated!

See also: https://github.com/reach-sh/reach-lang/pull/161

PR #148, opened 2021-05-20
Add Support for Lifting Interact Function Calls With `Null` Range

chrisnevers:
Add a form to convert the following

```javascript
Alice.interact.f(1, 2, 3);
```

into

```javascript
Alice.only(() => {
  interact.f(1, 2, 3);
});
```

This shorthand only works if the `interact` field is a function with a `null` range. In `evalAsEnv`, I grab the `Participant`'s interact object and wrap the values in `SLForm_liftInteract`, which will handle executing the function calls inside an `only`

### DIFF
--- a/docs-src/ref-programs-step.scrbl
+++ b/docs-src/ref-programs-step.scrbl
@@ -40,6 +40,14 @@ is a @tech{valid} program where @reachin{Alice}'s @tech{local state} includes th
 
 is an @tech{invalid} program, because @reachin{Bob} does not know @reachin{x}.
 
+The shorthand @reachin{PART.interact.METHOD(EXPR_0, ..., EXPR_n)} is available for calling an @reachin{interact} function
+from outside of an @reachin{only} block. Such functions must return @reachin{Null}; therefore, they are only useful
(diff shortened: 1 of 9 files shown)


docs-src/ref-programs.scrbl
```diff
@@ -666,7 +666,16 @@ An @deftech{interaction expression}, written @reachin{interact.METHOD(EXPR_0, ..
 
 An @tech{interaction expression} may also be written @reachin{interact.KEY}, where @reachin{KEY} is bound in the @tech{participant interact interface} to a non-function type.
 
-An @tech{interaction expression} may only occur in a @tech{local step}.
+An @tech{interaction expression} may only occur in a @tech{local step}—with the exception of the following shorthand.
```
jeapostrophe, line 669: I think you should remove this caveat and instead document the short-hand on `only`

jeapostrophe (changes requested):

docs-src/ref-programs.scrbl
chrisnevers: Moved these docs to the `only` section

PR #153, opened 2021-05-26
CFX working examples (wip)

DanBurton:


.circleci/rebuild.hs
```diff
@@ -58,6 +58,57 @@ skipAlgo =
...
+  , "chicken-fork"
+  , "chicken-parallel"
+  , "chicken-race"
+  , "race"
+  , "weird-swap"
+  , "workshop-hash-lock"
+  , "workshop-relay"
+  , "workshop-trust-fund"
+
+  -- Tx with same nonce already inserted.
```
jeapostrophe, line 79: My guess is that these errors are okay because it is just a sign that these participants are competing to be the next publisher

.circleci/rebuild.hs
```diff
@@ -58,6 +58,57 @@ skipAlgo =
...
+  , "raffle"
+  , "rent-seeking"
+  , "workshop-fomo"
+  , "workshop-fomo-generalized"
+
+  -- data: '"tx already exist"'
+  , "nft-auction"
+
+  -- Transaction ${txn} is discarded due to a too stale nonce
+  , "nft-dumb"
```
jeapostrophe, line 91: Ditto on this and the above

PR #157, opened 2021-05-28
Explain Unsupported/Bad ALGO at Runtime

chrisnevers:


hs/src/Reach/Connector/ALGO.hs
```diff
@@ -288,22 +289,22 @@ check_rekeyto = do
...
 bad :: LT.Text -> App ()
 bad lab = do
-  bad_
+  bad_ lab
   output $ comment_ $ "BAD " <> lab
 
 xxx :: LT.Text -> App ()
 xxx lab = do
-  let lab' = "XXX " <> lab
+  let lab' = "XXX This program uses " <> lab
```
jeapostrophe, line 305: We can remove the `XXX`

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -355,7 +356,7 @@ salloc fm = do
   Env {..} <- ask
   let eSP' = eSP - 1
   when (eSP' == eHP) $ do
-    bad "too much memory"
+    bad "Too much memory required"
```
jeapostrophe, line 359: Too many scratch slots

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -373,7 +374,8 @@ talloc tk = do
   liftIO $ modifyIORef eTxnsR $ (+) 1
   txni <- liftIO $ readIORef eTxnsR
   when (txni >= algoMaxTxGroupSize) $ do
-    bad "too many txns"
+    bad $ texty $
+      "Exceeded the maximum size of an atomic transfer group: " <> show algoMaxTxGroupSize
```
jeapostrophe, line 378: If we want to be helpful, we could say "This is caused by too many transfers in one atomic step."

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -550,7 +552,8 @@ check_concat_len :: Integer -> App ()
 check_concat_len totlen =
   case totlen <= 4096 of
     True -> nop
-    False -> bad $ "concat too big: 4096 < " <> texty totlen
+    False -> bad $ "Cannot `concat` " <>  texty totlen <>
```
jeapostrophe, line 555: This is caused by the Reach type being too big

jeapostrophe:
Push it

PR #158, opened 2021-05-30
Added new rsh file to popularity contest workshop

Bhaney44:
I simplified the code using standard JavaScript variables in a single function to allow new users to understand and execute the logic faster.

### DIFF
--- a/examples/popularity-contest/index.rsh
+++ b/examples/popularity-contest/index.rsh
@@ -1,72 +1,28 @@
+// reach
 'reach 0.1';
-'use strict';
 
-const [ _, ALICE_WINS, BOB_WINS, TIMEOUT ] = makeEnum(3);
+// vote function
+function contest (){
+    // three voter variables - user input boolean
+    // vote 1 for contestant 1 and 0 for contestant 0
+    var Voter0, Voter1, Voter2
+    Voter0 = input boolean
+    Voter1 = input boolean
+    Voter2 = input boolean
+    // two contestant variables - user input boolean
+    var contestant0, contestant1
+    contestant0 = 0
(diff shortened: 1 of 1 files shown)


jeapostrophe:
Thanks @Bhaney44 , this is an interesting idea. It would be cool to be able to have Reach do a lot more inference about what the structure of the DApp protocol is based on things like this. And it would be very powerful to be able to reason about mutation outside of `while` loops. However, right now we're quite far from this, so I think I'll have to close the pull request, since we want all of the code in the repository to be executable and passing tests.

I do want to push back a tiny bit on your description "standard JavaScript variables in a single function". Four comments:
- First, Reach uses JavaScript syntax, but it doesn't exactly have JavaScript semantics, because the DApp domain is so different. This may be a bit of a bad thing because there's not an exact match between those things, but I think it helps more than it hurts.
- Second, your program isn't in JavaScript syntax, `input boolean` isn't valid; you don't use `var` for the variables `total_votes` or `consensus`; and you use Python `if` notation.
- Third, what constitutes "standard" JavaScript has changed a lot over in its 25 year lifespan. I think that most people consider `var` to be deprecated these days. This article, from 6 years ago, is representative --- https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75
- Fourth, a Reach program *is* specified by a single function, plus some global variables. That's the function which is the argument to `Reach.App`.

Thanks for your ideas and I'm happy to chat in the Discord to get you working productively in Reach!

Bhaney44:
Thanks for this!

Bhaney44:
Thanks for this @jeapostrophe !

PR #162, opened 2021-06-02
Protect Against Solidity Keywords in Structs/Objects

chrisnevers:
* Check `Struct` names against a list of reserved words
* Have `Object`s preface field names with `_` . `munge/unmunge` adds/removes the `_` appropriately

### DIFF
--- a/hs/src/Reach/Connector/ETH_Solidity.hs
+++ b/hs/src/Reach/Connector/ETH_Solidity.hs
@@ -35,6 +35,7 @@ import System.Exit
 import System.FilePath
 import System.IO.Temp
 import System.Process
+import Data.Bifunctor (Bifunctor(first))
 
 --- Debugging tools
 
@@ -471,7 +472,7 @@ solLargeArg' dv la =
       where
         go i a = one (".elem" <> pretty i) <$> solArg a
     DLLA_Obj m ->
-      solLargeArg' dv $ DLLA_Struct $ M.toAscList m
+      solLargeArg' dv $ DLLA_Struct $ map (first ("_" <>)) $ M.toAscList m
     DLLA_Data _ vn vv -> do
       t <- solType $ largeArgTypeOf la
       vv' <- solArg vv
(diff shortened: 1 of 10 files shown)


hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -471,7 +472,7 @@ solLargeArg' dv la =
       where
         go i a = one (".elem" <> pretty i) <$> solArg a
     DLLA_Obj m ->
-      solLargeArg' dv $ DLLA_Struct $ M.toAscList m
+      solLargeArg' dv $ DLLA_Struct $ map (first ("_" <>)) $ M.toAscList m
```
jeapostrophe, line 475: Can you abstract these three uses of `("_" <>)` into a function in case we need to change it?

hs/src/Reach/Eval/Core.hs
```diff
@@ -2191,7 +2191,11 @@ evalPrim p sargs =
       kts <- mapM go as
       retV $ (lvl, SLV_Type $ ST_Struct kts)
       where
-        verifyStructId r s =
+        verifyStructId r s = do
+          at <- withAt id
+          connectors <- readDlo dlo_connectors
+          when ("ETH" `elem` connectors && s `elem` map show solReservedNames) $
```
jeapostrophe, line 2197: Make a note to extend the `Connector` type to communicate stuff like this

jeapostrophe (changes requested):

hs/src/Reach/Eval/Types.hs
```diff
@@ -101,3 +101,58 @@ data TransferType
 
 instance Show TransferType where
   show k = drop 3 $ conNameOf k
+
+data SolReservedNames
```
DanBurton, line 105: Nit: `data SolReservedName` since `SOL_address :: SolReservedName` makes more sense than `SOL_address :: SolReservedNames`

hs/test-examples/nl-eval-errors/Err_Sol_Reserved.txt
```diff
@@ -0,0 +1 @@
+reachc: error: ./Err_Sol_Reserved.rsh:5:24:application: `super` is not a valid field name because it is reserved in Solidity.
```
DanBurton, line 1: Does the `_` prefix treatment not handle this use case? Do we intend for it to handle it in the future?

DanBurton (approved):

hs/test-examples/nl-eval-errors/Err_Sol_Reserved.txt
chrisnevers: It's not handled because the `Struct`s have to obey an interface with remote contracts

PR #163, opened 2021-06-02
More Array Standard Library Functions

chrisnevers:
Add `find`, `withIndex`, `mapWithIndex`, and `reduceWithIndex`


docs-src/ref-programs-compute.scrbl
```diff
@@ -914,6 +934,28 @@ the value is not present in the array, @reachin{None} is returned.
...
+@subsubsection{@tt{Array.find} && @tt{.find}}
+
+@(mint-define! '("find"))
+@reach{
+  Array.find(arr, ty, f)
+  arr.find(ty, f) }
+
+@index{Array.find} @reachin{Array.find(arr, ty, f)} returns the first element of type, @reachin{ty}, in the array, @reachin{arr},
+that satisfies the predicate @reachin{f}. The return value is of type @reachin{Maybe(ty)}. If no value in the
+array satisfies the predicate, @reachin{None} is returned.
```
jeapostrophe, line 946: It is unfortunate that we need to pass the type, given that it is stored in the array. I think these means there should be something like `arr.elemType` to read off the type as an `SLType`

docs-src/ref-programs-compute.scrbl
```diff
@@ -892,6 +902,16 @@ This may be abbreviated as @reachin{arr.reduce(z, f)}.
...
 
+@subsubsection{@tt{Array.reduceWithIndex} && @tt{.reduceWithIndex}}
+
+@(mint-define! '("reduceWithIndex"))
+@reach{
+  Array.reduceWithIndex(arr, z, f)
+  arr.reduceWithIndex(z, f) }
+
+@index{Array.reduceWithIndex} @reachin{Array.reduceWithIndex(arr, z, f)} is the same as @reachin{Array.reduce}, except it
+provides @reachin{f} with an additional argument, which is the index of the current element in @reachin{arr}.
```
jeapostrophe, line 913: This, and `mapWithIndex` are not exactly alike, because I believe that those functions take N arguments and do implicit zipping. I think these functions should mention that they only work on one argument and if you want the indices, then use `withIndex` first on one of the arrays

jeapostrophe (changes requested):

jeapostrophe (approved):
Great!

PR #170, opened 2021-06-04
#1271: Dockerized `reach` script in Haskell

mattaudesse:
WIP


hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+import Options.Applicative
+import Options.Applicative.Help.Pretty
+import System.Posix.IO (stdError)
+
+import qualified Data.Text.Lazy    as T
+import qualified Data.Text.Lazy.IO as T
+
+default (T.Text)
+
+-- TODO usage help currently prints `reach-cli` instead of `reach`, which may
```
jeapostrophe, line 18: It should be correct from the start

hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+-- be confusing to end-users who shouldn't have to care that their CLI session
+-- is being forwarded to this binary... Consider renaming
+
+-- TODO
+reachVersion :: T.Text
+reachVersion = "0.1"
```
jeapostrophe, line 24: This should come from `Reach.Version`

hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+  | RpcRun
+  | RpcServer
+  | RpcServerDown
+  | Run
+  | Scaffold
+  | Unscaffold
+  | Update
+  | Upgrade
+  | Version
+  | Whoami
```
jeapostrophe, line 59: What's the deal with this?

hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+   <$$> text "If:"
+   <$$> text " * MODULE is a directory then `cd $MODULE && rm -f \"build/index.$IDENT.mjs\";"
+   <$$> text " * MODULE is <something-else> then `rm -f \"build/$MODULE.$IDENT.mjs\""
+  opts = Clean
+    <$> strArgument (metavar "MODULE" <> value "index" <> showDefault)
+    <*> strArgument (metavar "IDENT"  <> value "main"  <> showDefault)
+
+
+-- | > rm -f "build/$m.$i.mjs
+clean :: FilePath -> T.Text -> Script ()
```
jeapostrophe, line 80: Can there not be 5x triplicate duplication of information?

`cmdClean`, `clean`, `Clean`, `cmds`, and `main` all have the same information

hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+--------------------------------------------------------------------------------
+cmdHashes :: Mod CommandFields Cli
+cmdHashes = command "hashes" $ info (pure Hashes) desc where
+  desc = progDesc "Display git hashes used to build each Docker image"
+
+
+hashes :: Script ()
+hashes = flip mapM_ reachImages $ \i -> do
+  let t = "reachsh/" <> i <> ":" <> reachVersion
+      s = cmd "docker" "run" "--entrypoint" "/bin/sh" t "-c" (quote "echo $REACH_GIT_HASH")
```
jeapostrophe, line 239: Please follow our style with one binding per `let` unless you want mutually recursive bindings

hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+
+  -- Hidden
+  NumericVersion -> sh numericVersion
+  ReactDown      -> sh reactDown
+  RpcServerDown  -> sh rpcServerDown
+  Unscaffold     -> sh unscaffold
+  Whoami         -> sh whoami
+
+ where
+  sh f = T.putStrLn . script $ do
```
jeapostrophe, line 350: I think the commands should be in `IO` and return scripts

hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+rpcServerDown :: Script ()
+rpcServerDown = undefined
+
+
+--------------------------------------------------------------------------------
+cmdUnscaffold :: Mod CommandFields Cli
+cmdUnscaffold = command "unscaffold" $ info (pure Unscaffold) fullDesc
+
+
+unscaffold :: Script ()
```
jeapostrophe, line 285: I think you should use a type to encapsulate what a command is expected to be

hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+
+
+run' :: Script ()
+run' = undefined
+
+
+--------------------------------------------------------------------------------
+cmdDown :: Mod CommandFields Cli
+cmdDown = command "down" $ info (pure Down) desc where
+  desc = progDesc "Halt any dockerized devnets for this app"
```
jeapostrophe, line 123: You should be taking the opportunity to improve things. This should be `Dockerized`

hs/app/reach-cli/Main.hs
```diff
@@ -0,0 +1,352 @@
...
+  opts = Clean
+    <$> strArgument (metavar "MODULE" <> value "index" <> showDefault)
+    <*> strArgument (metavar "IDENT"  <> value "main"  <> showDefault)
+
+
+-- | > rm -f "build/$m.$i.mjs
+clean :: FilePath -> T.Text -> Script ()
+clean m i = do
+  let f m' = cmd "rm" "-f" $ "build/" <> T.pack m' <> "." <> i <> ".mjs"
+  if m == "index"
```
jeapostrophe, line 83: We don't use `if` unless it fits and looks good on one line, we use `case`

jeapostrophe (changes requested):

hs/app/reach-cli/Main.hs
mattaudesse: I was actually already planning to do something about all the repetition but I usually focus on consolidation once more of the body has taken shape.

How do you feel about this way instead?

hs/app/reach-cli/Main.hs
mattaudesse: I've created `type Subcommand = ParserInfo (Script ())` in response to the other thread but I don't think that's really what you're asking for here...

Can you foresee a case where we'd need to pass something other than `()` up to the `optparse` interpreter?

jeapostrophe: How does the `()` go up to the interpreter? I thought the `()` is the return from the shell script

hs/app/reach-cli/Main.hs
jeapostrophe: Yea, I like it better. Although I think I'd probably put the name with the code

hs/app/reach-cli/Main.hs
mattaudesse: I basically just mean this line: `join . fmap sh $ customExecParser (prefs showHelpOnError) cmds`.

The scripts don't really return anything meaningful back to the Haskell code.

hs/app/reach/Main.hs
```diff
@@ -0,0 +1,351 @@
...
+  f = undefined
+
+
+--------------------------------------------------------------------------------
+reachVersionShort :: TS.Text -- TODO
+reachVersionShort = TS.pack V.compatibleVersionStr
+
+
+initRsh :: TS.Text -> TL.Text
+initRsh v = TL.fromStrict [N.text|
```
jeapostrophe, line 110: I think these things should be external files that get embedded and/or are in the environment of the Docker image

hs/app/reach/Main.hs
```diff
@@ -0,0 +1,351 @@
...
+    const Bob   = Participant('Bob', {});
+    deploy();
+    // write your program here
+
+  });
+|]
+
+
+initMjs :: TS.Text -> TL.Text
+initMjs app = TL.fromStrict [N.text|
```
jeapostrophe, line 124: This too

hs/app/reach/Main.hs
```diff
@@ -37,11 +37,36 @@ reachImages =
   ]
 
 
+--------------------------------------------------------------------------------
+printfl :: PrintfArg r => String -> r -> TL.Text
+printfl s r = TL.pack $ printf s r
+
+
+fmtInitRsh :: String
+fmtInitRsh = $(mkEmbed "init/index.rsh")
```
jeapostrophe, line 46: If it is not toooo distracting, maybe try to make it so that we support different templates as we talked about?

hs/src/Reach/EmbeddedFiles.hs
```diff
@@ -14,3 +15,8 @@ stdlib_sol = $(makeRelativeToProject "./sol/stdlib.sol" >>= embedFile)
 
 stdlib_rsh :: ByteString
 stdlib_rsh = $(makeRelativeToProject "./rsh/stdlib.rsh" >>= embedFile)
+
+-- Exported as a shorthand convenience for app/reach/Main.hs to work around
+-- GHC's "stage restriction" rules
```
jeapostrophe, line 20: What's this? And would the other things benefit from the below?

reach2
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+case "$1" in
+  "bootstrap") echo "TODO" ;; # Document this in `optparse` as well
```
jeapostrophe, line 4: I'm reading your mind and thinking that you shouldn't have this... the normal path should just check to see if the Docker image is around and download it if it isn't around

reach2
```diff
@@ -0,0 +1,13 @@
...
+
+  *)
+    s=$(docker run -it --rm reachsh/reach-cli:latest "$@")
```
jeapostrophe, line 7: I was thinking
```
if docker run .... > /tmp/reach.out ; then
 /bin/sh /tmp/reach.out
else
 cat /tmp/reach.out
fi
```

I am anxious about searching in the output, the echo command-line getting too long, and the echo command doing interpolation

mattaudesse: I agree with you about not liking this way but the problem is finding some reliable means of distinguishing `/bin/sh`-able script (I'm thinking of regular usage but also shell completions if we get that far) and `optparse` output.

If I were to pass `reach2 --help` to your code above it'd fail and spit a bunch of lines like:
```bash
/bin/sh: n1: some text: not found
/bin/sh: n2: something else: not found
...
```

If I instead ran `reach2 this-subcommand-does-not-exist` then it would fall through to `cat` and just print the usage info correctly.

Experimenting with several variations yesterday led me to think we're not going to get around doing something like`RUNME`, but maybe there's a better way.

Perhaps we could branch on a specific exit code instead of inspecting the output? I considered doing that but wasn't thrilled with that solution either.

Also, I think interpolation is an issue whether we stuff the output in a file or not, no?

jeapostrophe: Ok, I think that "branch on a specific exit code" is the way to go. `RUNME` replaced with `exit 42` at the end, then inspect `$?`

jeapostrophe: And piping to a file and then executing it is guaranteed not to have the wrapper shell involved

mattaudesse:
```bash
#!/bin/sh

IMG='reachsh/reach-cli:latest'

run () {
  docker run -t --rm "$IMG" "$@"
}

if ! (which make docker docker-compose >/dev/null 2>&1); then
  echo "Reach relies on an installation of make, docker, and docker-compose"
  exit 1
fi

if [ "$(docker image ls -q $IMG)" = '' ]; then
  if ! (docker pull "$IMG" && run update | sed 's/\r//' | sh); then exit 1; fi
fi

run "$@" >/tmp/reach.out
case "$?" in
  42) sed 's/\r//' /tmp/reach.out | sh ;;
   *) cat /tmp/reach.out ;;
esac
```

How about that?

reach2
```diff
@@ -1,13 +1,19 @@
...
-    fi
-    ;;
+IMG='reachsh/reach-cli:latest'
+
+if ! (which make docker docker-compose >/dev/null 2>&1) ; then
+  echo "Reach relies on an installation of make, docker, and docker-compose"
+  exit 1
+fi
+
+if [ "$(docker image ls -q $IMG)" = '' ] ; then
```
jeapostrophe: I bet shellcheck will want an `x` in there :)

reach2
```diff
@@ -1,13 +1,19 @@
...
+  docker pull "$IMG"
+  docker run -t --rm "$IMG" update | sed 's/\r//' | sh
+fi
+
+docker run -t --rm "$IMG" "$@" >/tmp/reach.out
+case "$?" in
+  42) sed 's/\r//' /tmp/reach.out | sh ;;
```
jeapostrophe, line 20: Why are there `\r`s?

reach2
```diff
@@ -2,17 +2,20 @@
...
+if ! (which make docker docker-compose >/dev/null 2>&1); then
   echo "Reach relies on an installation of make, docker, and docker-compose"
   exit 1
 fi
 
-if [ "$(docker image ls -q $IMG)" = '' ] ; then
-  docker pull "$IMG"
-  docker run -t --rm "$IMG" update | sed 's/\r//' | sh
+if [ "$(docker image ls -q $IMG)" = '' ]; then
+  if ! (docker pull "$IMG" && run update | sed 's/\r//' | sh); then exit 1; fi
```
jeapostrophe, line 15: You get the exit for free if you do `/bin/sh -e`

jeapostrophe: Oh, but I guess we can't because of the exit 42

mattaudesse: Yeah, for that and a few other reasons I don't think we can use `set -e` or `sh -e` in the wrapper script.

(The interpreted script coming from our Haskell code does `set -e` though.)

reach2
mattaudesse: I've been running `shellcheck` all along and it has no objection to this line. Would you rather there be an `x` anyway?

reach2
mattaudesse: This is the Docker output quirk that I alluded to before but I don't have a good answer for why they do that... I just know everything blows up if I don't strip them out before interpreting. My guess without digging too deep is it's something to do with cross-platform concerns like capturing containers' output on Windows.

Running outside of Docker doesn't produce the extraneous `\r`s.

reach2
jeapostrophe: No, I'm good

reach2
jeapostrophe: What if the script creates a file and then we `docker cp` it out of the container? Maybe this is unnecessary, but I'm scared of arbitrarily changing the output... what if there's `\r` that's supposed to be there?

mattaudesse: Yeah, perhaps that's better?

I'm actually worried about arbitrary edits too, but every time I consider these edge cases I invariably come back to shell scripts being messy and heavily flawed... I don't know if it's even possible to construct this architecture such that it won't be vulnerable to subtle problems roughly similar to what you're describing (e.g. my hesitation about exit codes, or interpolation being somewhat unavoidable).

I also have a small aversion to leaving (and possibly overwriting) files on the users' filesystem, so I lean towards keeping things in-memory if I can help it.

How would you feel about the wrapper script choosing a "vacant" spot under `/tmp` (rather than overwriting) and cleaning up after itself?

jeapostrophe: I think it is safe to use `/tmp/reach.$$.out`

hs/app/reach/Main.hs
```diff
@@ -49,56 +61,30 @@ reachImages =
...
-
-
-cat' :: CmdParams p => p
-cat' = cmd "cat"
-
+reachVersion :: App
+reachVersion =
+  let v = T.pack V.versionStr
+  in write [N.text|
+    if [ "x$$REACH_VERSION" = "x" ]; then
```
jeapostrophe, line 68: `$$`?

jeapostrophe: Ah, this is Haskell templates. That's scary

reach2
```diff
@@ -15,8 +15,9 @@ if [ "$(docker image ls -q $IMG)" = '' ]; then
   if ! (docker pull "$IMG" && run update | sed 's/\r//' | sh); then exit 1; fi
 fi
 
-run "$@" >/tmp/reach.out
+run "$@" >/tmp/reach.$$.out
 case "$?" in
-  42) sh -ac "$(sed 's/\r//' /tmp/reach.out)" "$0" ;;
-   *) cat /tmp/reach.out ;;
+  # TODO use less fragile workaround for `\r` characters?
```
jeapostrophe, line 20: Mount and have it write. Put that in a comment if you can't do it fast

hs/app/reach/Main.hs
mattaudesse: Yeah... It's pretty gnarly doing it this way.

My goal is to gradually push as much out of the quasi-quotes as possible and keep only those snippets small enough that template indirection would be overkill.

hs/app/reach/embed/sh/_common/ensureConnectorMode.sh
```diff
@@ -0,0 +1,67 @@
+# Makes sure to set REACH_CONNECTOR_MODE to one of:
```
jeapostrophe, line 1: Something I want to do is simplify and normalize all these names so that we can more easily programmatically generate this page. I want `$CONN(-(live|devnet|browser))?` to be the only options with `devnet` being the default and `browser` only being allowed when directly using the stdlib

hs/app/reach/embed/sh/subcommand/scaffold/docker-compose.yml
```diff
@@ -0,0 +1,86 @@
+version: '3.4'
+x-app-base: &app-base
+  image: ${IMAGE_TAG}
+services:
+  devnet-cfx:
```
jeapostrophe, line 5: I imagine that each connector will have a single `docker-compose.yml` excerpt that we maintain in one place and plop in when needed. Meaning that this file would be generated by splicing all of those together

Makefile
```diff
@@ -54,9 +54,18 @@ docs:
 
 .PHONY: sh-lint
 sh-lint:
-	@(set -e; for f in $$(find . -not \( -path '*/node_modules/*' -prune \) -name '*.sh') ./reach; do \
-		echo linting $$f; \
-		shellcheck --external-sources --source-path SCRIPTDIR $$f ; \
+	@(set -e; for f in \
+	    $$(find . \
+	      -not \( -path '*/node_modules/*' -prune \) \
+	      -not \( -path './hs/app/reach/embed/*' -prune \) \
```
jeapostrophe, line 60: I think another option would be to change your use of templates so that the Haskell templates don't use `$` but instead use a character like `@` or `%` and then shellcheck will be able to work with them.

jeapostrophe (changes requested):

hs/app/reach/Main.hs
```diff
@@ -65,15 +65,161 @@ writeFrom :: FilePath -> App
...
+            - ethereum-devnet
+          environment:
+            - REACH_DEBUG
+            - REACH_CONNECTOR_MODE=ETH-test-dockerized-geth
+            - ETH_NODE_URI=http://ethereum-devnet:8545
+      |]
+
+
+--------------------------------------------------------------------------------
+data ConnectorMode
```
jeapostrophe, line 130: I think this type should be `(Connector, Mode)` where `Connector` is `ALGO | CFX | ETH` and `Mode` is `Devnet|Live|Browser`

hs/app/reach/Main.hs
```diff
@@ -65,15 +65,161 @@ writeFrom :: FilePath -> App
...
+  | LiveAlgo
+  | LiveCfx
+  | LiveEth
+  | BrowserAlgo
+  | BrowserCfx
+  | BrowserEth
+
+
+instance Show ConnectorMode where
+  show = \case
```
jeapostrophe, line 143: And this should be mechanical

hs/app/reach/Main.hs
```diff
@@ -65,15 +65,161 @@ writeFrom :: FilePath -> App
...
+    DevnetAlgo -> "ALGO-test-dockerized-algod"
```
jeapostrophe, line 144: And these should change to the new naming scheme

hs/app/reach/Main.hs
```diff
@@ -65,15 +65,161 @@ writeFrom :: FilePath -> App
...
+    DevnetCfx -> "CFX-devnet"
+    DevnetEth -> "ETH-test-dockerized-geth"
```
jeapostrophe, line 146: Ditto

hs/app/reach/Main.hs
```diff
@@ -65,15 +65,161 @@ writeFrom :: FilePath -> App
...
+    BrowserAlgo -> "ALGO-browser"
+    BrowserCfx -> "CFX-browser"
+    BrowserEth -> "ETH-browser"
+
+
+connectorModeNonBrowser :: IO ConnectorMode
+connectorModeNonBrowser = lookupEnv "REACH_CONNECTOR_MODE" >>= maybe (pure DevnetEth) (\case
+  "ALGO" -> pure DevnetAlgo
+  "ALGO-test" -> pure DevnetAlgo
+  "ALGO-test-dockerized" -> pure DevnetAlgo
```
jeapostrophe, line 159: Ditto

hs/app/reach/Main.hs
```diff
@@ -65,15 +65,161 @@ writeFrom :: FilePath -> App
...
+    LiveCfx -> "CFX-live"
+    LiveEth -> "ETH-live"
+    BrowserAlgo -> "ALGO-browser"
+    BrowserCfx -> "CFX-browser"
+    BrowserEth -> "ETH-browser"
+
+
+connectorModeNonBrowser :: IO ConnectorMode
+connectorModeNonBrowser = lookupEnv "REACH_CONNECTOR_MODE" >>= maybe (pure DevnetEth) (\case
+  "ALGO" -> pure DevnetAlgo
```
jeapostrophe, line 157: This `case` can also split based on - and then look at each separately

hs/app/reach/Main.hs
```diff
@@ -275,39 +393,30 @@ clean = command "clean" . info f $ fullDesc <> desc <> fdoc where
...
---  * --install-pkgs
---  * --dir-dot-reach
---  * [SOURCE]
---  * [EXPORTS...]
 compile :: Subcommand
 compile = command "compile" $ info f d where
   d = progDesc "Compile an app"
-  f = pure . script $ do
+
+  f = go
```
jeapostrophe, line 400: It seems painful to have to maintain this copy of the args

hs/app/reach/embed/sh/_common/_docker-compose/service-devnet-algo-postgres-db.yml
```diff
@@ -0,0 +1,6 @@
+algorand-postgres-db:
```
jeapostrophe, line 1: I am nervous about this being a separate file rather than just having one list of containers for algo and another list of containers for eth/cfx

jeapostrophe (changes requested):

hs/app/reach/Main.hs
mattaudesse: I used `data ConnectorMode = ConnectorMode Connector Mode` because there's already a `Show` instance for tuples that would overlap here.

.circleci/example.sh
```diff
@@ -14,12 +14,20 @@ case "${CONN}" in
   ETH) TIMEOUT=$((5 * 60)) ;;
 esac
 
+D="devnet-$(echo "$CONN" | tr '[:upper:]' '[:lower:]')"
+printf 'Bringing up %s...' "$D"
```
jeapostrophe, line 18: just say Bringing up devnet... no need to be fancy

.circleci/example.sh
```diff
@@ -14,12 +14,20 @@ case "${CONN}" in
...
 ../reach devnet >/dev/null 2>&1 &
+while true; do
+  if [ "$(docker ps -qf "label=sh.reach.devnet-for=$D" | wc -l)" -gt 0 ]; then break; fi
```
jeapostrophe, line 21: You should just make it so the label is the actual connector mode we use

jeapostrophe (changes requested):

.circleci/example.sh
```diff
@@ -52,16 +52,29 @@ go() {
...
 
-while true; do
-  wait -n || true
-  KIDS=$((KIDS - 1))
-  if [ $KIDS -eq 0 ]; then break; fi
-done
+EXS="$(find . -maxdepth 1 -type d | sed 'sX./XX' | sort | tail -n +2 | awk "NR % ${SIZE} == ${RANK}")"
+
+case "${CONN}" in
+  # XXX Conflux' transparent nonce selection currently isn't robust enough to
```
jeapostrophe, line 59: This is too slow though... you should instead make it so they are forked but have their own devnets

.circleci/example.sh
```diff
@@ -52,16 +52,29 @@ go() {
...
+  # http://developer.confluxnetwork.org/conflux-doc/docs/send_transaction#send-transaction-by-javascript-program
+  CFX) for WHICH in $EXS; do go "${WHICH}" 'Running'; done ;;
+
+  *)
+    for WHICH in $EXS; do
+      KIDS=$((KIDS + 1))
+      go "${WHICH}" 'Forking' &
+    done
+
+    while true; do
```
jeapostrophe, line 71: Why not just `wait`?

mattaudesse: Discussed via Slack

PR #176, opened 2021-06-09
WIP: Revise How SMT Errors Are Reported

chrisnevers:
https://trello.com/c/cOOlddhv/1290-smt-revise-how-errors-are-reported-again

### DIFF
--- a/hs/src/Reach/AST/DLBase.hs
+++ b/hs/src/Reach/AST/DLBase.hs
@@ -17,9 +17,7 @@ import Control.Monad.Identity
 import Control.Monad.Reader
 import Data.Functor ((<&>))
 
-data PrettySubstEnv = PrettySubstEnv {
-  pse_model_vals :: M.Map DLVar Doc,
-  pse_inline :: M.Map DLVar Doc }
+type PrettySubstEnv = M.Map DLVar Doc
 
 type PrettySubstApp = ReaderT PrettySubstEnv Identity
 
@@ -240,11 +238,11 @@ data DLArg
   deriving (Eq, Ord, Generic, Show)
 
 instance Pretty DLArg where
-  pretty = runIdentity . flip runReaderT (PrettySubstEnv mempty mempty) . prettySubst
+  pretty = runIdentity . flip runReaderT mempty . prettySubst
 
 instance PrettySubst DLArg where
   prettySubst a = do
-    env <- asks pse_inline
+    env <- ask
     case a of
       DLA_Var v -> return $ fromMaybe (viaShow v) (M.lookup v env)
       DLA_Constant c -> return $ pretty c
@@ -302,7 +300,7 @@ instance (PrettySubst a, PrettySubst b) => PrettySubst (a, b) where
     return $ parens $ hsep $ punctuate comma [x', y']
 
 instance Pretty DLLargeArg where
-  pretty = runIdentity . flip runReaderT (PrettySubstEnv mempty mempty) . prettySubst
+  pretty = runIdentity . flip runReaderT mempty . prettySubst
 
 instance PrettySubst DLLargeArg where
   prettySubst = \case
@@ -453,7 +451,7 @@ instance PrettySubst a => PrettySubst (Maybe a) where
     Nothing -> return "Nothing"
 
 instance Pretty DLExpr where
-  pretty = runIdentity . flip runReaderT (PrettySubstEnv mempty mempty) . prettySubst
+  pretty = runIdentity . flip runReaderT mempty . prettySubst
 
 instance PrettySubst DLExpr where
   prettySubst = \case
@@ -745,7 +743,7 @@ data DLPayAmt = DLPayAmt
   deriving (Eq, Generic, Ord)
 
 instance Pretty DLPayAmt where
-  pretty = runIdentity . flip runReaderT (PrettySubstEnv mempty mempty) . prettySubst
+  pretty = runIdentity . flip runReaderT mempty . prettySubst
 
 instance PrettySubst DLPayAmt where
   prettySubst (DLPayAmt {..}) = do
--- a/hs/src/Reach/AST/DLBase.hs
+++ b/hs/src/Reach/AST/DLBase.hs
@@ -17,7 +17,9 @@ import Control.Monad.Identity
 import Control.Monad.Reader
 import Data.Functor ((<&>))
 
-type PrettySubstEnv = M.Map DLVar Doc
+data PrettySubstEnv = PrettySubstEnv {
+  pse_model_vals :: M.Map DLVar Doc,
+  pse_inline :: M.Map DLVar Doc }
 
 type PrettySubstApp = ReaderT PrettySubstEnv Identity
 
@@ -238,12 +240,13 @@ data DLArg
   deriving (Eq, Ord, Generic, Show)
 
 instance Pretty DLArg where
-  pretty = runIdentity . flip runReaderT mempty . prettySubst
+  pretty = runIdentity . flip runReaderT (PrettySubstEnv mempty mempty) . prettySubst
 
 instance PrettySubst DLArg where
   prettySubst a = do
+    env <- asks pse_inline
     case a of
-      DLA_Var v -> return $ viaShow v
+      DLA_Var v -> return $ maybe (viaShow v) parens (M.lookup v env)
       DLA_Constant c -> return $ pretty c
       DLA_Literal c -> return $ pretty c
       DLA_Interact who m t ->
@@ -299,7 +302,7 @@ instance (PrettySubst a, PrettySubst b) => PrettySubst (a, b) where
     return $ parens $ hsep $ punctuate comma [x', y']
 
 instance Pretty DLLargeArg where
-  pretty = runIdentity . flip runReaderT mempty . prettySubst
+  pretty = runIdentity . flip runReaderT (PrettySubstEnv mempty mempty) . prettySubst
 
 instance PrettySubst DLLargeArg where
   prettySubst = \case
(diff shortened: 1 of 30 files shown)


hs/src/Reach/AST/DLBase.hs
```diff
@@ -13,6 +13,14 @@ import Reach.Counter
 import Reach.Pretty
 import Reach.Texty
 import Reach.Util
+import Control.Monad.Identity
+import Control.Monad.Reader
+import Data.Functor ((<&>))
+
+type PrettyT = ReaderT (M.Map DLVar Doc) Identity
```
jeapostrophe, line 20: This would be named `PrettyApp` if we're following our naming convention, although probably `PrettySubstApp` is more accurate. We'd also give the `M.Map` a new name `Env`

hs/src/Reach/AST/DLBase.hs
```diff
@@ -228,12 +236,16 @@ data DLArg
   deriving (Eq, Ord, Generic, Show)
 
 instance Pretty DLArg where
-  pretty = \case
-    DLA_Var v -> pretty v
-    DLA_Constant c -> pretty c
-    DLA_Literal c -> pretty c
-    DLA_Interact who m t ->
-      "interact(" <> render_sp who <> ")." <> viaShow m <> parens (pretty t)
+  pretty = runIdentity . flip runReaderT mempty . prettySubst
```
jeapostrophe, line 239: This could be an instance `PrettySubst a => Pretty a`

hs/src/Reach/AST/DLBase.hs
```diff
@@ -228,12 +236,16 @@ data DLArg
...
+
+instance PrettySubst DLArg where
+  prettySubst a = do
```
jeapostrophe, line 242: Use `\case` and repeat yourself with `return`, I think

hs/src/Reach/AST/DLBase.hs
```diff
@@ -260,13 +272,42 @@ data DLLargeArg
...
-    DLLA_Array t as -> "array" <> parens (pretty t <> comma <+> pretty (DLLA_Tuple as))
-    DLLA_Tuple as -> brackets $ render_das as
-    DLLA_Obj env -> render_obj env
-    DLLA_Data _ vn vv -> "<" <> pretty vn <> " " <> pretty vv <> ">"
-    DLLA_Struct kvs -> "struct" <> brackets (render_das kvs)
+  pretty = runIdentity . flip runReaderT mempty . prettySubst
+
+instance PrettySubst DLLargeArg where
+  prettySubst = \case
+    DLLA_Array t as -> return $ "array" <> parens (pretty t <> comma <+> pretty (DLLA_Tuple as))
```
jeapostrophe, line 304: These `as` may contain variables

hs/src/Reach/Verify/SMT.hs
```diff
@@ -413,268 +365,112 @@ seVars se =
...
+    T_Address -> do
+      case v of
+        Atom i -> return $ SMV_Address $ B.pack i
+        _ -> impossible "parseVal: Address"
+    T_Bytes _ -> do
+      case v of
+        Atom i -> return $ SMV_Bytes $ B.pack i
+        _ -> impossible "parseVal: Bytes"
+    T_Array (T_Tuple [T_Token, T_UInt]) _ ->
+      -- XXX This is a Map
```
jeapostrophe, line 435: Arrays are also arrays when the key is a number

hs/src/Reach/AST/DLBase.hs
```diff
@@ -301,7 +303,9 @@ instance Pretty DLLargeArg where
 
 instance PrettySubst DLLargeArg where
   prettySubst = \case
-    DLLA_Array t as -> return $ "array" <> parens (pretty t <> comma <+> pretty (DLLA_Tuple as))
+    DLLA_Array t as -> do
+      t' <- prettySubst (DLLA_Tuple as)
+      return $ "array" <> parens (pretty t <> comma <+> t')
     DLLA_Tuple as -> render_dasM as <&> brackets
     DLLA_Obj env -> render_objM env
     DLLA_Data _ vn vv -> return $ "<" <> pretty vn <> " " <> pretty vv <> ">"
```
jeapostrophe, line 311: The `vv` could have variables

hs/src/Reach/Verify/SMTAst.hs
```diff
@@ -170,26 +175,43 @@ instance Pretty SMTTrace where
...
+    inlinable' <- mapM (prettySubst . toWitness) inlinable
+    lets' <- mapM prettySubst others
     return $
+      "  // Violation Witness" <> hardline <> hardline <>
+      concatWith (surround hardline) inlinable' <> hardline <> hardline <>
+      "  // Theorem Formalization" <> hardline <> hardline <>
       concatWith (surround hardline) lets' <> hardline <>
       "  " <> pretty tk <> parens (pretty dv) <> ";" <> hardline
+    where
+      -- xxx nail down how to get witness variables
```
jeapostrophe, line 188: Why can't the annotation on `SMTLet` be trusted? It may be useful for the PR to have some updates to a few choice test outputs so I can follow what the current thing prints out with

chrisnevers:
@jeapostrophe regarding maps and using `List` over `Set`: 

This change is due to the fact that `add_counts` requires traversing a statement list in the correct order to determine var usage. Before, the `SMTLet` Set could be converted to a `List` fine because it would use the ordering of `DLVar` counters, which happened to coincide with the order of evaluation for `DLStmt`s. 

However, Maps will use counters produced by `smt_alloc`, which differ from the other `DLVar`s, and they will be out of order if not stored in a list. So, I'm storing them explicitly according to evaluation order. This will allow for `add_counts` to properly determine var usage.

# 

Regarding vars assigned twice, it happens with `var`s in `while` loops. The var will get assigned before and at the end of the loop. So, to avoid this I should make `newCtxtScope` create a new copy of the current `ctxt_smt_trace` env?

jeapostrophe:
> regarding maps and using `List` over `Set`:

I am totally down with using a `List`. I was confused and thought you meant that in the future you would use a Haskell `Map`, not that this would be useful for Reach's `Map`.

> Regarding vars assigned twice, it happens with `var`s in `while` loops. The var will get assigned before and at the end of the loop. So, to avoid this I should make `newCtxtScope` create a new copy of the current `ctxt_smt_trace` env?

I don't understand why that happens. If you have:

```
var x = 2;
invariant(x > 1);
while (x < 10) {
 A.publish();
 x = x + 1;
 continue;
}
```

Then we have this in the theorem prover on the `continue`:

```
x = ?
assume (x > 1)
x' = x + 1
assert (x' > 1)
```

where `x'` is a new variable that comes out of `freshen`. The variables are never actually assigned twice. Am I wrong?

chrisnevers:
@jeapostrophe I think the Map reductions aren't very descriptive because anything in a while loop is in a temporary scope.. I'm having difficulties figuring out if there's anything to make it better?

hs/t/n/map-dset.txt
```diff
@@ -7,14 +7,27 @@ Verification failed:
...
+  //      from: ./map-dset.rsh:16:51:application
+  const b/36 = "Bob".interact.get();
+  //    ^ could = 0
+  //      from: ./map-dset.rsh:21:51:application
+
+
+  // Theorem Formalization
+
+  const map0_0/57 = new Map();
+
```
jeapostrophe, line 23: Where do these extra lines come from?

hs/t/n/map-dset.txt
```diff
@@ -7,14 +7,27 @@ Verification failed:
...
+  const map0_1/70 = set(map0_0/57["Address!val!0"], Just a/27);
```
jeapostrophe, line 24: I'm surprised that `map0_0` didn't get substituted.

hs/t/n/map-dset.txt
```diff
@@ -7,14 +7,27 @@ Verification failed:
...
+
+  const map0_2/76 = set(map0_1/70["Address!val!0"], Just b/36);
```
jeapostrophe, line 26: I would format this like in PL: `map0_1["Address!val!0" <- Just b]`

jeapostrophe: Also, I think that you should consider formatting `"Address!val!0"` as something like `<abstract address 0>`

jeapostrophe:
> @jeapostrophe I think the Map reductions aren't very descriptive because anything in a while loop is in a temporary scope.. I'm having difficulties figuring out if there's anything to make it better?

I don't understand which example you're referring or what is lacking about it

hs/t/n/map-dset.txt
chrisnevers: For maps and null values, I don't currently show what the values would be. The extra line is a poorly placed `hardline` in my output that I'll fix. I think I'll change this to show `would be null` and `would be <map>` for such values

hs/t/n/map-dset.txt
chrisnevers: It didn't get subbed because `canDupe SMTMapNew` is false. I figured the only `SynthExpr` that could be dupe'd is `SMTMapRef`. Should new maps be able to be duplicated?

jeapostrophe: Well, we're not going to be executing these, they are just informative. So I agree that if we're talking about a real program we should not turn `x = new(); y = x; z = x;` into `y = new(); z = new();` because we would lose sharing. In this case though, we are really saying `x = something-that-came-from-new; y = x; z = x;` and in that case, it is more like the pointer that comes from `new` than the call itself. Same with set.

jeapostrophe (approved):
This looks good to go. Double check that any docs that excerpt errors are still using the correct lines and then merge

PR #178, opened 2021-06-09
Revise docs

temptemp3:


jeapostrophe:
The original text is correct. Also, we tend not to use `@emph`, because we always use `@deftech` and `@tech` instead.

jeapostrophe:
Thank you for helping to improve things :)

temptemp3:
No problem. I guess I will just have to get used to interword anchor tags. I am still getting used to all the tags. The usage I found for @emph was exactly like another page in the docs. I am thinking it may be nice to have some sort of glossary for things showing up in @deftech and @tech; whereas we may not consider doing the same for @emph because it only has to do with presentation. Anyways. thanks for the tip!

PR #185, opened 2021-06-14
js/CFX: try to un-parallelize sends from same sender

DanBurton:


js/stdlib/ts/cfxers.ts
```diff
@@ -282,3 +274,93 @@ export class Wallet {
     throw Error(`Account 'from mnemonic' not supported on Conflux, please use secret key`);
   }
 }
+
+// XXX This is nutty
```
jeapostrophe, line 278: I think I need some understanding dropped on me. You're trying to "un-parallelize sends from same sender"... how are there /any/ parallel sends? Every send has an `await` with it... so they are serialized?

js/stdlib/ts/cfxers.ts
```diff
@@ -282,3 +274,93 @@ export class Wallet {
...
+// Remember the last epoch that a given sender has sent
+// and don't try to send again until it is later than that epoch.
+// Note: requires addrs to be canonicalized first.
+const lastEpochSent: Record<string, number> = {};
+const epochWaitLock: Record<string, boolean> = {};
+
+// XXX implement a queue, maybe?
```
jeapostrophe, line 285: I think we can and should be more willing to require other npm packages. There must be a good thread-safe JS queue

js/stdlib/ts/cfxers.ts
DanBurton: I noticed this specifically as an issue with accessing the faucet. Something like this:

```
  18   │   const accPlayer_arr = await Promise.all(
  19   │     Array.from({ length: numOfPlayers }, () =>
  20   │       stdlib.newTestAccount(startingBalance)));
```

can cause the faucet to try to send funds to a bunch of people in parallel, and CFX will throw errors about reusing the same nonce.

DanBurton: that's from examples/raffle

jeapostrophe: Ah; that makes sense. I think that you might consider a tighter focused attack on only the faucet rather than trying to do this more general thing

DanBurton:
Rewrote and merged in separate PR. See: https://github.com/reach-sh/reach-lang/pull/188

PR #189, opened 2021-06-17
examples/tut-8: update to include cfx

DanBurton:


jeapostrophe:
This seems fine to me

PR #194, opened 2021-06-24
Sort ETH logs by blockNumber then logIndex

chrisnevers:


js/stdlib/ts/ETH_like.ts
```diff
@@ -614,14 +614,16 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account>
...
-      // XXX Are we sure that this returns the logs in order? Algo didn't maybe
-      // sort to be safe?
-      return await (await getProvider()).getLogs({
+      const logs = await (await getProvider()).getLogs({
         fromBlock,
         toBlock,
         address: ethersC.address,
         topics: [ethersC.interface.getEventTopic(ok_evt)],
       });
+      return logs.sort((x: Log, y: Log) =>
```
jeapostrophe, line 623: You should look at the code that calls this function, but I'm almost positive that we only care about the first one, so this should be more like a findMin

jeapostrophe (changes requested):

js/stdlib/ts/ETH_like.ts
jeapostrophe: Meaning, change it to return just the one?

PR #195, opened 2021-06-24
Prevent Multiple Stdlibs From Being Loaded By Default

chrisnevers:
Add `unsafeAllowMultipleStdlibs: () -> null` to allow users to explicitly allow multiple stdlib usage

### DIFF
--- a/docs-src/ref-frontends-js.scrbl
+++ b/docs-src/ref-frontends-js.scrbl
@@ -115,6 +115,19 @@ In such scenarios, we recommend that you call this function like so:
 You may instead pass in the string @litchar{'ETH'} or the string @litchar{'ALGO'}
 to select the desired stdlib directly.
 
+By default, this method allows a user to load a single standard library.
+That is, this method may not be called multiple times, with the same or varying @tech{connectors}.
+To bypass this restriction, use @jsin{unsafeAllowMultipleStdlibs}.
+
+@(hrule)
+@(mint-define! '("unsafeAllowMultipleStdlibs"))
+@js{
+  unsafeAllowMultipleStdlibs() => null
+}
+
+@index{unsafeAllowMultipleStdlibs} Calling this function will lift the restriction that
+@jsin{loadStdlib} imposes on loading multiple standard libraries.
+
 @section[#:tag "ref-frontends-js-acc"]{Accounts}
 
 These functions create and interact with @tech{account} representations.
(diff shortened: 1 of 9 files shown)


docs-src/ref-frontends-js.scrbl
```diff
@@ -115,6 +115,19 @@ In such scenarios, we recommend that you call this function like so:
 You may instead pass in the string @litchar{'ETH'} or the string @litchar{'ALGO'}
 to select the desired stdlib directly.
 
+By default, this method allows a user to load a single standard library.
+That is, this method may not be called multiple times, with the same or varying @tech{connectors}.
```
jeapostrophe, line 119: Can you make it work on the same connector a second time?

examples/maybe-send/index.mjs
```diff
@@ -62,6 +60,7 @@ const demo = async (x) => {
 };
 
 (async () => {
-  await demo(1);
-  await demo(null);
+  const stdlib = await stdlib_loader.loadStdlib();
```
jeapostrophe, line 63: That would make it so this file doesn't need to change

examples/variable-transfers/index.mjs
```diff
@@ -41,7 +40,8 @@ export const run = async (n) => {
 };
 
 (async () => {
-  await run(0);
-  await run(1);
-  await run(2);
+  const stdlib = await loadStdlib();
```
jeapostrophe, line 43: Ditto

examples/workshop-trust-fund/index.mjs
```diff
@@ -57,7 +56,8 @@ const runDemo = async (delayReceiver, delayFunder) => {
 };
 
 (async () => {
-  await runDemo(false, false);
-  await runDemo(true, false);
-  await runDemo(true, true);
+  const stdlib = await stdlib_loader.loadStdlib();
```
jeapostrophe, line 59: Ditto

js/stdlib/ts/registry.ts
```diff
@@ -0,0 +1,15 @@
+import { Connector } from "./ConnectorMode";
+
+let currentConnector : null | true | Connector = null;
+
+export const unsafeAllowMultipleStdlibs = () => {
+  currentConnector = true;
+};
+
+export const doStdlibLoad = (connector: Connector) => {
+  if (currentConnector == null || currentConnector == true) {
```
jeapostrophe, line 10: `currentConnector === connector`

jeapostrophe (changes requested):

docs-src/ref-frontends-js.scrbl
chrisnevers: okay, sorry I misinterpreted the ticket

jeapostrophe:
It's good to merge

PR #200, opened 2021-06-27
Minting

jeapostrophe:
- no Algorand


hs/Makefile
```diff
@@ -10,11 +10,30 @@ install-mo:
 mo:
 	@which mo 1>/dev/null || (echo 'FAILURE: Please call `make install-mo` first' && exit 1)
 
-package.yaml: package.mo.yaml ../VERSION mo
+package.yaml: package.mo.yaml ../VERSION
```
DanBurton, line 13: Why the change on this line?

jeapostrophe: Because `mo` is phony, it thinks that this always needs to be rebuilt

PR #203, opened 2021-06-30
WIP: Error Codes & Pretty Error Messages

chrisnevers:


docs-src/lib.rkt
```diff
@@ -387,3 +387,6 @@ You should start off by initializing your Reach program:
 
 (define (note-ctransfer)
   @margin-note{If you're unsure of what kind of @tech{consensus transfer} to use, you may want to read the @seclink["guide-ctransfers"]{explanation of the differences} in the Guide.})
+
+(define (error-version x y)
+  @margin-note{This error will not happen before version @tt{@x} or after version @tt{@y}})
```
jeapostrophe, line 392:
```
(define (error-version #:from [from #f] #:to [to #f])
 (cond
  [(and from to)
    ... what you have ...]
  [from
   ... skip after ...]
  [to
   ... skip before ...])
```

docs-src/ref-error-codes.scrbl
```diff
@@ -0,0 +1,334 @@
...
+
+@;{
+  What is a "good" error description?
+  * Intuitive explanation
+  * Example errorneous program
+  * How to fix to fix that program
+  * Any other advice
+}
+
+@section{RE0001}
```
jeapostrophe, line 17: You should make `@error` or something like `(define (error x) @section[#:tag x]{@|x|})`

examples/pkg/index-master.txt
```diff
@@ -1 +1,10 @@
-reachc: error: ./index-master.rsh:5:27:import from: Cannot install packages; did you mean to run with `--install-pkgs`?
+reachc: error: RI0003
```
jeapostrophe, line 1: I think you should work on this format more. I like the rust style:

```
error[E0001]: unreachable pattern
  --> src/main.rs:10:9
   |
10 |         Known => 1,
   |         ^^^^^ this is an unreachable pattern
   |
```

The URLs need an `https` for link discoverers to turn on.

hs/src/Reach/AST/Base.hs
```diff
@@ -60,6 +66,9 @@ data ImpossibleError
   = Err_Impossible String
```
jeapostrophe, line 66: This PR essentially tells us that `Err_Impossible` should not exist and instead `DLE_Impossible` should take a different constructor that is more specific so we can write real explanations.

hs/src/Reach/Connector.hs
```diff
@@ -38,6 +39,9 @@ data ConnectorError
   = Err_IntLiteralRange Integer Integer Integer
   deriving (Eq, Generic, ErrorMessageForJson, ErrorSuggestions)
 
+instance HasErrorCode ConnectorError where
+  errCode e = "RC" <> leftPad 4 '0' (show $ gconIndex e)
```
jeapostrophe, line 43: You should make a helper in `Reach.Util` for this pattern AND you need to make a comment on the constructors that the indexes are now part of the external interface and cannot be changed. That's dangerous though. There *is* going to be a day when we put them in the wrong order, delete one, or something like that. I think it might be better to have them written out by hand. `HasErrorCode` will provide a default implementation of `errCode` that looks at the `errPrefix` and the `errIndex` and we manually type out a bunch of numbers to get `errIndex`

docs-src/ref-error-codes.scrbl
```diff
@@ -0,0 +1,334 @@
...
+For example, the code below erroneously uses a @jsin{for} loop, which is not
+supported in Reach:
+
+@reach{
+  for (let i = 0; i < arr.length; i++) {
+    // ...
+  }
+}
+
+To fix this code, either use a @reachin{while} loop or a combination of
```
jeapostrophe, line 58: Use active speech, "This code should be written..." or "You can fix this by..." etc

docs-src/ref-error-codes.scrbl
```diff
@@ -0,0 +1,334 @@
...
+      curriedAdd1: (x) => { x_ = x; },
+      curriedAdd2: (y) => { return x_ + y; }
+    })
+  ]);
+}
+
+@section{RE0013}
+
+This error indicates that you provided an unrecognized option to @reachin{setOptions}.
+There is most likely a typo in your code.
```
jeapostrophe, line 287: Link to options?

docs-src/ref-error-codes.scrbl
```diff
@@ -0,0 +1,334 @@
...
+
+@section{RE0014}
+
+This error indicates that you did not provide an acceptable value for a specific option
+in @reachin{setOptions}. Please review the eligible values listed in the documentation
```
jeapostrophe, line 292: Semantic newline

PR #207, opened 2021-07-03
wip cfx testnet

DanBurton:
~~Doesn't work yet, due to an issue with mixing up epochs and blocks.~~

Picked this branch back up; it's soon approaching mergeability...


jeapostrophe (commented):
I don't understand what the "idea" behind the query lower bound functions is

DanBurton:
The idea is: when the stdlib performs queries for event logs, you can set the lower bound for when it should start querying.

PR #211, opened 2021-07-09
Add `define` component to `parallelReduce`

chrisnevers:
Add `.define(() => BLOCK)` to `parallelReduce` where `BLOCK` translates directly to the `BLOCK` of a `while` loop.


jeapostrophe:
Merge it!

PR #215, opened 2021-07-12
js: expose more types on interfaces

DanBurton:


js/stdlib/ts/shared_backend.ts
```diff
@@ -76,7 +76,7 @@ export function Array_set <T>(arr: Array<T>, idx: number, elem: T): Array<T> {
   arrp[idx] = elem;
   return arrp;
 }
-export const mapRef = (m: any, f: any): any => {
+export const mapRef = <A>(m: {[key: string]: A}, f: string): ['Some', A]|['None', null] => {
```
jeapostrophe, line 79: I think you should make a `MaybeRep` type

jeapostrophe (changes requested):

js/stdlib/ts/interfaces.ts
```diff
@@ -1,44 +1,46 @@
...
-  mapRef: any,
-  simMapRef: any,
-  simMapSet: any,
-  simMapDupe: any,
-  bytesConcat: any,
+  checkedBigNumberify: (at: string, max: BigNumber, n: any) => BigNumber
+  protect: (t: any, v: unknown, ai?: string) => unknown
+  Array_zip: <A, B>(a1: A[], a2: B[]) => [A, B][]
+  mapRef: <A>(m: {[key: string]: A}, f: string) => MaybeRep<A>
+  simMapRef: (sim_r: unknown, mapi: number, f: string) => MaybeRep<unknown>
```
jeapostrophe, line 40: Is this `unknown` because it is too hard to use `ISimRes` here?

js/stdlib/ts/interfaces.ts
```diff
@@ -64,64 +66,79 @@ export interface Stdlib_Backend<Ty> extends Stdlib_Backend_Base<Ty> {
...
-  isHex: any
-  bigNumberify: any
-  stringToHex: any
-  addressEq: any
-  digest: any
+  isHex: (x: unknown) => boolean
+  bigNumberify: (n: num|string) => BigNumber
+  stringToHex: (s: string) => string
+  addressEq: (addr1: string, addr2: string) => boolean
+  digest: (t: any, a: unknown) => string
```
jeapostrophe, line 105: `t` is a list of Types... is not included because it's hard?

jeapostrophe (approved):

js/stdlib/ts/interfaces.ts
DanBurton: `t` is a single type (which can be an array type). I haven't tried very hard to do type params yet, but yeah it needs to be ETH ty or ALGO ty depending.

js/stdlib/ts/interfaces.ts
DanBurton: fsvo "too hard" yes

PR #218, opened 2021-07-15
js: delete algo custom types

DanBurton:


jeapostrophe:
Wow! It just works with the real types?

jeapostrophe:
If so, merge it!

PR #221, opened 2021-07-16
Uniswap

chrisnevers:
* Add Uniswap example
* Add `Token.destroyed()`
* Use `auflia` logic for SMT instead of `qflia`, is there any reason not to use this logic? Verification seems to take forever to with `qflia`

Currently, takes 57 seconds to verify, any tips for quicker verification?


hs/src/Reach/Eval/Core.hs
```diff
@@ -1413,6 +1415,7 @@ evalForm f args = do
       case mmode of
         Just FM_Case -> do
           a <- withAt srcloc2annot
+          -- XXX Default to 0 for all non-network tokens if `paySpec` specified
```
jeapostrophe, line 1418: That makes sense... any reason not to do it right now?

hs/src/Reach/Eval/Core.hs
```diff
@@ -2095,6 +2098,14 @@ evalPrim p sargs =
           lookupBalanceFV FV_supply $ Just da
         _ -> illegal_args
       doFluidRef s
+    SLPrim_Token_destroyed -> do
+      s <- case args of
+        [v] -> do
```
jeapostrophe, line 2103: You should use `one_arg`

docs-src/ref-programs-consensus.scrbl
```diff
@@ -330,6 +331,11 @@ This must be called before the application exits.
 
 @(hrule)
 
+@reachin{Token.destroyed(tok)}, or @reachin{tok.destroyed()}, where @reachin{tok} is a @reachin{Token} value, returns whether @reachin{destroy}
```
jeapostrophe, line 334: This needs to go in the changelog

jeapostrophe (changes requested):

jeapostrophe:
The ALGO failure is a real mistake in the frontend. You probably need to do `acc.tokenAccept(token)`

examples/uniswap/index.rsh
```diff
@@ -0,0 +1,289 @@
...
+          // Amount of each token in reserve to return to Provider
+          const amtOuts = balances.map(bal => liquidity * bal / poolStart);
+
+          // Payout provider
+          const currentProvider = this;
+          transfer(amtOuts[0], tokA).to(currentProvider);
+          transfer(amtOuts[1], tokB).to(currentProvider);
+
+          // Burn the liquidity tokens
+          pool.burn(liquidity);
```
jeapostrophe, line 174: This is forever in Reach, so probably just subtract from poolMinted

jeapostrophe (approved):

PR #222, opened 2021-07-20
Small Reach.Texty refactor

tippenein:
I was testing the build tools setup for the haskell code and saw some comments about removing these from the module. The synopsis is
- remove `group`, 
- remove the int arg from `nest` (default 2)


jeapostrophe:
Thanks for this. Looking at it, I think that I like `hardline` more than having the consumers actually look at the constructor. Also, we don't use merge commits, but instead do rebasing so we can have as linear a git history as possible

jeapostrophe:
This is great. Thank you :)

PR #230, opened 2021-07-27
Cache Event Queries In Runtime

chrisnevers:
## Benchmark (ALGO)

`cache-events` program has participant catch up on 1200 publications

|With Cache| Time (minutes) |
|------------|------|
|`false`|`5:13.71`|
|`true`|`3:49.86`|

I can adjust the test program to less publications to decrease test time?


js/stdlib/ts/ALGO.ts
```diff
@@ -589,20 +589,83 @@ type QueryResult =
...
-  const txn = ptxns.reduce((accum: any, x: any) =>
-    (x['confirmed-round'] < accum['confirmed-round']) ? x : accum, ptxns[0]);
-  return { succ: true, txn };
-};
+  async query(dhead: string, ApplicationID: number, roundInfo: RoundInfo, pred: ((x:any) => boolean)): Promise<QueryResult> {
+    const { minRound, maxRound, specRound } = roundInfo;
+    debug(dhead, `EventCache.query`, ApplicationID, minRound, specRound, maxRound, this.currentRound);
+
+    // Clear cache of stale transactions
+    const filterRound = minRound || specRound || 0;
```
jeapostrophe, line 621: I am very suspicious of specRound... I'll have to see where that gets used

js/stdlib/ts/ALGO.ts
```diff
@@ -589,20 +589,83 @@ type QueryResult =
...
+    this.cache = this.cache.filter(x => x['confirmed-round'] > filterRound);
+
+    // Check to see if the transaction we want is in cache
+    const initPtxns = this.cache.filter(pred);
```
jeapostrophe, line 625: This is not quite right either, because imagine that we are looking for a call to m3 and m3 is called in a loop. The actual trace could be `m2 -> m3` (because `m2` is a timeout that goes back to the top of the while and enables another `m3`. In that case, the query we REALLY want is "Is m3 the earliest transaction?" Now, the only way this could happen is if the first `m3` (that we're looking for) is only valid in `[100, 200]` but it fails so `m2` happens at `201`, then the next `m3` happens at `202`. Your code is ignoring the maxRound specified, so presumably the real query is `m3 \in [100, 200]`, but you are doing `m3 \in [100, inf)` so you find the `202` one. You need to first filter the query up to the new minimum (not the specified), then compute the part of the txns that are valid for THIS query (the spec and max) and then see if the predicate applies

js/stdlib/ts/ALGO.ts
```diff
@@ -589,20 +589,83 @@ type QueryResult =
...
+    if (initPtxns.length != 0) {
+      debug(`Found transaction in Event Cache`);
+      const txn = chooseMinRoundTxn(initPtxns)
+      this.currentRound = txn['confirmed-round'];
+      return { succ: true, txn };
+    }
+
+    debug(`Transaction not in Event Cache. Querying network...`);
```
jeapostrophe, line 633: This is not right either, because it's not that we ask the network when the query isn't answered, it is if the range of the query is not included in the cache. For example, if the cache holds `[100, 200]`, but we are looking for something in the range `[150, 1000]`, then we need to got to the network to get more txns. Now, we could first check `[150, 200]` in the cache, then if that fails, go to the network for `[200, 1000]`

js/stdlib/ts/ALGO.ts
```diff
@@ -1794,13 +1845,17 @@ async function waitCtorTxn(shad: string, ApplicationID: number): Promise<void> {
     debug(shad, 'waitCtorTxn waiting (ms)', waitMs);
     await Timeout.set(waitMs);
     debug(shad, 'waitCtorTxn trying attempt #', tries, 'of', maxTries);
-    icr = await queryCtorTxn(`${shad} deploy`, ApplicationID);
+    icr = await queryCtorTxn(`${shad} deploy`, ApplicationID, new EventCache());
```
jeapostrophe, line 1848: This function (`waitForCtorTxn`) could take a cache to share over time

js/stdlib/ts/ETH_like.ts
```diff
@@ -266,6 +266,84 @@ const stepTime = async (): Promise<TransactionReceipt> => {
   return await transfer(faucet, acc, parseCurrency(0));
 };
 
+// ****************************************************************************
+// Event Cache
```
jeapostrophe, line 270: I'm not going to read this closely. I assume it has the same logic as the ALGO above, so please revise and/or confirm it doesn't then I'll make more comments.

jeapostrophe (changes requested):

js/stdlib/ts/ALGO.ts
chrisnevers: Fixed this to work as you described

js/stdlib/ts/ALGO.ts
```diff
@@ -593,17 +593,21 @@ type QueryResult =
 // Event Cache
 // ****************************************************************************
 
-export const chooseMinRoundTxn = (ptxns: any[]) =>
+const chooseMinRoundTxn = (ptxns: any[]) =>
   ptxns.reduce((accum: any, x: any) =>
       (x['confirmed-round'] < accum['confirmed-round']) ? x : accum, ptxns[0])
 
+const chooseMaxRoundTxn = (ptxns: any[]) =>
```
jeapostrophe, line 600: These two functions suggest adding `argmax` to our JS utilities --- https://docs.racket-lang.org/reference/pairs.html#%28def._%28%28lib._racket%2Flist..rkt%29._argmax%29%29

js/stdlib/ts/ALGO.ts
```diff
@@ -641,21 +651,26 @@ export class EventCache {
...
-    }
-    if (maxRound) {
-      query = query.maxRound(maxRound);
+      // If cache has: [100, 200]
+      // & querying  : [150, 1000]
+      // We already searched cache for [150, 200] so query network for [201, 1000]
+      query = query.minRound(Math.max(this.currentRound, minRound) + 1);
     }
     if (specRound) {
       query = query.round(specRound);
```
jeapostrophe, line 660: If you include this, then when you verify the contract, you won't get all of the events that will be needed next, so you'll have to two indexer trips. The `specRound` should just be the lower bound. In other words...

js/stdlib/ts/ALGO.ts
```diff
@@ -641,21 +651,26 @@ export class EventCache {
...
 
     if (minRound) {
-      query = query.minRound(minRound + 1);
-    }
-    if (maxRound) {
-      query = query.maxRound(maxRound);
+      // If cache has: [100, 200]
+      // & querying  : [150, 1000]
+      // We already searched cache for [150, 200] so query network for [201, 1000]
+      query = query.minRound(Math.max(this.currentRound, minRound) + 1);
```
jeapostrophe, line 657: Based on the next comment, I think that this could should not use `minRound`, but should use `filterRound`

jeapostrophe: You really want `Math.max(this.currentRound + 1, minRound)` because the user really wants the lower bound to actually be `minRound` (i.e. `filterRound`)

js/stdlib/ts/ETH_like.ts
```diff
@@ -276,24 +276,39 @@ const getMinBlock = (logs: any[]) =>
       ? (x.logIndex < acc.logIndex ? x : acc)
       : (x.blockNumber.toString() < acc.blockNumber.toString() ? x : acc), logs[0]);
 
+const getMaxBlock = (logs: any[]) =>
+  logs.reduce((acc: Log, x: Log) =>
```
jeapostrophe, line 280: Another `argmax`

js/stdlib/ts/ETH_like.ts
```diff
@@ -276,24 +276,39 @@ const getMinBlock = (logs: any[]) =>
...
+  cache: any[] = [];
+
+  currentBlock = 0;
 
   constructor() {
     this.cache = [];
   }
 
-  async query_(fromBlock: number, toBlock: number, topic: string, getLogs: () => Promise<any[]>) {
+  async query_(fromBlock: number, toBlock: number, topic: string, getLogs: (currentBlock: number) => Promise<any[]>) {
```
jeapostrophe, line 296: I continue to not read this closely :)

jeapostrophe (changes requested):

js/stdlib/ts/ETH_like.ts
chrisnevers: I don't think these ETH functions translate well to `argmax`, I translated the ALGO ones to use `argmax`

jeapostrophe: Sounds good

js/stdlib/ts/ETH_like.ts
jeapostrophe: Should I read the ETH code closely now? Or are you still working on it?

chrisnevers: yeah, I think it's good now

js/stdlib/ts/ETH_like.ts
```diff
@@ -266,6 +266,103 @@ const stepTime = async (): Promise<TransactionReceipt> => {
...
+    const initLogs = this.cache.filter(filterFn);
+    if(initLogs.length > 0) {
+      debug(`Found transaction in Event Cache`);
+      return getMinBlock(initLogs);
+    }
+
+    debug(`Transaction not in Event Cache. Querying network...`);
+
+    // If no results, then contact network
+    this.cache = await getLogs(this.currentBlock + 1);
```
jeapostrophe, line 320: There's an opportunity to go from `fromBlock`... it probably won't matter given that we expect there to not be any between them

js/stdlib/ts/ETH_like.ts
```diff
@@ -266,6 +266,103 @@ const stepTime = async (): Promise<TransactionReceipt> => {
...
+
+    this.currentBlock =
+      (this.cache.length == 0)
+        ? toBlock
```
jeapostrophe, line 324: I wonder if `getLogs` is hiding anything from us, such as the actual time of the query? At the very least, you could get the current time BEFORE the `getLogs` call and use that as the "current time" analog from ALGO.

Also, it is possible that we queried for `[200, 1000]` but the real network time is `500`, so you're setting it to `1000`, when you should be setting it to `500`.

jeapostrophe (changes requested):

PR #231, opened 2021-07-30
Adding Community page with initial list of community links.

hagenhaus:


docs-src/community.scrbl
```diff
@@ -0,0 +1,35 @@
+#lang scribble/manual
+@(require "lib.rkt")
+@(mint-scope "Community")
+
+@title[#:version reach-vers #:tag "community"]{Community}
+
+Below are links to Reach-related instructional content from the Reach community. To submit a link for consideration, contact mhagen@"@"reach.sh.
```
jeapostrophe, line 7: We use semantic newlines (i.e. add a newline after `.;,` and only then)

jeapostrophe: Please change to active voice "If you would like to..." and don't have your email, instead say to submit a Github issue and we'll add it.

docs-src/community.scrbl
```diff
@@ -0,0 +1,35 @@
...
+@item{
+  @link["https://devtimeblog.medium.com/connecting-funding-an-algorand-test-wallet-from-a-vanillajs-parcel-reach-application-54f9fc996d53"]{Connecting/Funding an Algorand (test) Wallet from a VanillaJS + Parcel + Reach Application} by Krysztof on June 28, 2021.
+}
+
+@item{
+  @link["https://devtimeblog.medium.com/connecting-funding-an-algorand-test-wallet-from-a-reach-react-application-e6affaadcf65"]{Connecting/Funding an Algorand (test) Wallet from a Reach + React Application} by Krysztof on May 29, 2021.
+}
+
+@item{
+  @link["https://devtimeblog.medium.com/the-reach-algorand-bounty-hack-c46e8b823003"]{The Reach Algorand Bounty Hack} by Krysztof on May 25, 2021.
```
jeapostrophe, line 32: Please reformat so the dates are in a sane format (2021/05/25) and at the front.

jeapostrophe (changes requested):

docs-src/community.scrbl
```diff
@@ -0,0 +1,35 @@
+#lang scribble/manual
+@(require "lib.rkt")
+@(mint-scope "Community")
```
jeapostrophe, line 3: This doesn't mean what you think it does, so you should delete it

docs-src/community.scrbl
```diff
@@ -0,0 +1,35 @@
...
+
+@title[#:version reach-vers #:tag "community"]{Community}
```
jeapostrophe, line 5: You need to link this page from the main one with `@include-section`

jeapostrophe (changes requested):

jeapostrophe:
If you're abandoning these, we should close them

PR #234, opened 2021-08-02
Docs community

hagenhaus:
Jay, made changes as requested per my understanding.


docs-src/community.scrbl
```diff
@@ -0,0 +1,34 @@
+#lang scribble/manual
+@(require "lib.rkt")
+
+@title[#:version reach-vers #:tag "community"]{Community}
+
+Below are links to Reach-related instructional content from the Reach community. If you would like to submit a link for consideration, please create a @link["https://github.com/reach-sh/reach-lang/issues"]{Github} issue.
```
jeapostrophe, line 6: You should add a semantic newline here

jeapostrophe (changes requested):

PR #236, opened 2021-08-03
add initial structure of the simulator

chikeabuah:


hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+import qualified Data.Map.Strict as M
+import Reach.AST.Base
+import Reach.AST.DLBase
+import Reach.AST.LL
+
+-- NOTE: let several base types be Integer/String as a placeholder for now
+
+type ConsensusEnv = M.Map DLVar DLVal
+
+type Balances = [Double]
```
jeapostrophe, line 16: They won't be doubles... they will be naturals.

jeapostrophe: Why a list? Each account would have just one, unless you're thinking of tokens. In which case, I'd say that the ledger would be `M.Map Account (M.Map Token Balance)`

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+type Frontends = [ Frontend ]
+
+type NewPartActions = M.Map SLPart [ DLTail ]
+
+type State = (Ledger, ConsensusEnv, Frontends, NewPartActions, Action)
+
+-- possible backend actions
+data Action
+  = ConsAction LLConsensus
+  | StepAction LLStep
```
jeapostrophe, line 31: I think that the action would be something like `compute` and it means "do one step of computation", however, maybe you could make it so that all such boring "compute" actions are ignored and not presented to the user... we only present to them the options that they can control? It might be tedious to press "next" 17 times.

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+
+data Contract = Contract
+  { contr_contract_info :: Integer
+  , address :: String
+  , contr_token :: String
```
jeapostrophe, line 36: What's the idea behind these? I don't think a contract will be explicitly represented, because the entire DApp is just one contract. In other words, the ConsensusEnv and the code pointer ARE the contract

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+data Account = Account
+  { nw_acc :: Integer
+  , backend :: LLProg
+  , contract :: Contract
+  , acc_contract_info :: Integer
+  , acc_token :: String
```
jeapostrophe, line 44: Same question. My assumption is that an address is an opaque thing... probably just its address, which would be abstracted to a nat.

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+data DLVal
+  = V_Null
+  | V_Bool Bool
+  | V_UInt Integer
+  | V_Bytes Integer
+  | V_Digest String
```
jeapostrophe, line 52: The digest is abstract and wouldn't be a concrete thing. Maybe you'd just expose what it was a digest of (i.e. a DLVal)

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+  , contract :: Contract
+  , acc_contract_info :: Integer
+  , acc_token :: String
+  }
+
+data DLVal
+  = V_Null
+  | V_Bool Bool
+  | V_UInt Integer
+  | V_Bytes Integer
```
jeapostrophe, line 51: Bytes are strings

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+  | V_Digest String
+  | V_Address String
```
jeapostrophe, line 53: These would be the ledger entry

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+  | V_Token String
```
jeapostrophe, line 54: What's this supposed to be?

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+  | V_Array [Integer]
```
jeapostrophe, line 55: Arrays can contain any DLVals

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+  | V_Object (M.Map SLVar DLVal)
+  | V_Data (M.Map SLVar DLVal)
+  | V_Struct [(SLVar, DLVal)]
+  deriving (Eq, Ord)
+
+type App = ReaderT Env IO
+
+data Env = Env
+  { e_id :: Integer
+  , e_state :: State
```
jeapostrophe, line 66: The environment would be the mapping of ids to states, not the "current" id/state

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+-- Identify program states
+type Id = Int
+
+-- All of the states visited
+type Session = M.Map Id State
```
jeapostrophe, line 73: Basically this is the Env, although mutable

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,94 @@
...
+
+type Ledger = M.Map Account (M.Map Token Balance)
+
+type Frontend = ( Account, SLPart, M.Map DLVar DLVal, [ DLTail ] )
+
+type Frontends = [ Frontend ]
+
+type NewPartActions = M.Map SLPart [ DLTail ]
+
+type State = (Ledger, ConsensusEnv, Frontends, NewPartActions, DAppCode)
```
jeapostrophe, line 26: `Ledger` is really `ConsensusNetworkState`

hs/src/Reach/Debugger.hs
```diff
@@ -0,0 +1,94 @@
...
+import Reach.AST.DLBase
+import Reach.AST.LL
+
+type ConsensusEnv = M.Map DLVar DLVal
+
+type Balance = Integer
+
+type Token = String
+
+type Ledger = M.Map Account (M.Map Token Balance)
```
jeapostrophe, line 18: The ConsensusNetworkState also contains the next account and the next token

hs/src/Reach/Debugger.hs
```diff
@@ -5,15 +5,16 @@ module Reach.Debugger where
...
 import Reach.AST.DLBase
 import Reach.AST.LL
 
 type ConsensusEnv = M.Map DLVar DLVal
 
-type Balance = Integer
+type Balance = Natural
 
-type Token = Integer
+type Token = Natural
```
jeapostrophe, line 17: I know I said `Nat`, but that's just because I'm an idiot. `Integer` is fine... we don't need to add more types. Sorry

hs/src/Reach/Debugger.hs
```diff
@@ -81,15 +80,19 @@ type Session = M.Map Id State
...
+type Gas = Integer
 
 -- determine if program is already a value
 get_prog_val :: LLProg -> Maybe DLVal
 get_prog_val = undefined
 
 interp :: Gas -> LLProg -> App (Maybe DLVal)
 interp 0 prog = return $ get_prog_val prog
-interp _n _prog = undefined
+interp _n prog = case prog of
```
jeapostrophe, line 91: I'd do this case in the binding position, that's our Reach style

hs/src/Reach/Debugger.hs
```diff
@@ -81,15 +80,19 @@ type Session = M.Map Id State
...
+  (LLProg _at _llo _ps _dli _dex _dvs st) -> case st of
```
jeapostrophe, line 92: And this case as `\case`

hs/src/Reach/Debugger.hs
```diff
@@ -81,15 +80,19 @@ type Session = M.Map Id State
...
 
 -- interpreter
-type Gas = Natural
+type Gas = Integer
 
 -- determine if program is already a value
 get_prog_val :: LLProg -> Maybe DLVal
 get_prog_val = undefined
 
 interp :: Gas -> LLProg -> App (Maybe DLVal)
```
jeapostrophe, line 89: FWIW, this signature isn't going to work, because `LLProg`s don't return to values, they would reduce to more `LLProg`s, as you're going to need an environment to find the values. My taste re "gas" is to have a single step function and then a wrapper that calls many times, like how Redex works

hs/src/Reach/Debugger.hs
```diff
@@ -361,27 +361,60 @@ instance Interp DLBlock where
...
+          let (switch_binding, cons) = (M.!) switch_cases k
+          case switch_binding of
+            Nothing -> interp cons
+            Just ident -> add_to_store ident v $ interp cons
+    LLC_FromConsensus _at1 _at2 step -> interp step
+    LLC_While at asn _inv cond body k -> do
+      case asn of
+        DLAssignment asn' -> do
+          mapM (\(k,v)-> interp $ DL_Set at k v) $ M.toList asn'
+          while cond body
```
jeapostrophe, line 388: I am skeptical this will work well, because of my comment about FromConsensus

hs/src/Reach/Debugger.hs
```diff
@@ -361,27 +361,60 @@ instance Interp DLBlock where
...
+        _ -> impossible "consensus interpreter"
+    LLC_Switch _at var switch_cases -> do
+      ev <- interp $ DLA_Var var
+      case ev of
+        V_Data k v -> do
+          let (switch_binding, cons) = (M.!) switch_cases k
+          case switch_binding of
+            Nothing -> interp cons
+            Just ident -> add_to_store ident v $ interp cons
+    LLC_FromConsensus _at1 _at2 step -> interp step
```
jeapostrophe, line 383: You should actually block at this point and return to the context (or maybe you want to think about that blocking happening inside of the `step`)

hs/src/Reach/Debugger.hs
jeapostrophe: I assumed you'd want to make `interp s@(while c b ; k) = interp (if c { b; s } else { k })`

hs/src/Reach/Debugger.hs
```diff
@@ -361,27 +361,60 @@ instance Interp DLBlock where
...
+    -- QUESTION
+    LLC_ViewIs _at _part _var _export cons -> interp cons
+
+instance Interp LLStep where
+  interp = \case
+    LLS_Com stmt step -> do
+      _ <- interp stmt
+      interp step
+    LLS_Stop _at -> return V_Null
+    LLS_ToConsensus _at _tc_send _tc_recv _tc_mtime -> undefined
```
jeapostrophe, line 404: This is going to cause a block while you wait for the user to choose what happens next (i.e. who wins this race)

hs/src/Reach/Debugger.hs
```diff
@@ -361,27 +361,60 @@ instance Interp DLBlock where
...
+          mapM (\(k,v)-> interp $ DL_Set at k v) $ M.toList asn'
+          while cond body
+          interp k
+    LLC_Continue at asn -> do
+      case asn of
+        DLAssignment asn' -> do
+          mapM (\(k,v)-> interp $ DL_Set at k v) $ M.toList asn'
+          return V_Null
+    -- QUESTION
+    LLC_ViewIs _at _part _var _export cons -> interp cons
```
jeapostrophe, line 396: Part of the debug UI should be the values of the views, and this would record those. I think V1 doesn't need this feature

hs/src/Reach/Debugger.hs
```diff
@@ -1,4 +1,5 @@
 {-# OPTIONS_GHC -Wno-missing-export-lists #-}
+{-# OPTIONS_GHC -fno-warn-incomplete-patterns #-}
```
jeapostrophe, line 2: Heresy!

hs/src/Reach/Debugger.hs
```diff
@@ -57,11 +59,12 @@ type Frontend = M.Map String ([DLVal] -> DLVal)
 --   | BTAction -- print the backtrace
 --   | ShowAction String -- print the variable described by this string
 
--- data Action
---   = TieBreakAction
---   | NewAccAction
---   | NewPartAction
---   | ImitateFrontendAction
+data Action
+  = TieBreakAction
```
jeapostrophe, line 63: Our style is names like `${PREFIX}_Blah` rather than `Blah${POSTFIX}`

hs/src/Reach/Debugger.hs
```diff
@@ -57,11 +59,12 @@ type Frontend = M.Map String ([DLVal] -> DLVal)
...
+  | NewAccAction
+  | NewPartAction
+  | ImitateFrontendAction
+  | InteractAction [DLVal]
```
jeapostrophe, line 67: You'd need to know the function too and the expected return type

hs/src/Reach/Debugger.hs
```diff
@@ -105,315 +110,365 @@ type Id = Integer
...
+      ev2 <- forceSucc <$> interp dlarg2
       case (ev1,ev2) of
-        (V_Array arr, V_UInt n) -> return $ arr !! (fromIntegral n)
+        (V_Array arr, V_UInt n) -> return $ Succ $ arr !! (fromIntegral n)
         _ -> impossible "expression interpreter"
     DLE_ArraySet _at dlarg1 dlarg2 dlarg3 -> do
-      ev1 <- interp dlarg1
-      ev2 <- interp dlarg2
-      ev3 <- interp dlarg3
+      ev1 <- forceSucc <$> interp dlarg1
```
jeapostrophe, line 244: Needing `forceSucc` is a sign that you haven't got the types of the continuation transformer cooperating enough. The whole idea of the monad is that it will hide the fact that capture and suspension happens beneath you. If you need to force it, then that means you're implementing things yourself

jeapostrophe: (The same is true with needing `Succ` itself.) The `Result` would be something that is part of the state of a thread, not part of the interpreter

hs/src/Reach/Debugger.hs
chikeabuah: @jeapostrophe without this i basically get a billion useless warnings from the compiler: 

```
Patterns not matched:
            (ADD,
             ((V_UInt (integer-gmp-1.0.3.0:GHC.Integer.Type.S# _)):
              (V_UInt (integer-gmp-1.0.3.0:GHC.Integer.Type.S# _)):_:_))
            (ADD,
             ((V_UInt (integer-gmp-1.0.3.0:GHC.Integer.Type.S# _)):
              (V_UInt (integer-gmp-1.0.3.0:GHC.Integer.Type.Jp# _)):_:_))
            (ADD,
             ((V_UInt (integer-gmp-1.0.3.0:GHC.Integer.Type.S# _)):
              (V_UInt (integer-gmp-1.0.3.0:GHC.Integer.Type.Jn# _)):_:_))
            (ADD,
             ((V_UInt (integer-gmp-1.0.3.0:GHC.Integer.Type.Jp# _)):
              (V_UInt (integer-gmp-1.0.3.0:GHC.Integer.Type.S# _)):_:_))
            ...
```

jeapostrophe: Sounds like you need `_ -> impossible "bad!"` in that function

hs/src/Reach/Debugger.hs
chikeabuah: @jeapostrophe oh i thought that was the design of the interpreter based on your code example: 

```

Jay
–
Oct 25th
View in channel
Jay 11:20 AM
interp :: a -> App Result

data Result
  = Succ DLVal
  | Block Signal (DLVal -> App Result)

``` 

so you're saying that the interpreter shouldn't return a `Result`?

jeapostrophe: There's two options for the interpreter:
- "Hide blocking" --- Like by using a monad, so the interpreter will just return `DLVal` but the blocking will be hidden behind the scenes
- "Expose blocking" --- In this case, the `Result` type is useful because it allows us to do it by hand

Regardless of what we do for `interp`, there's another issue of how to keep track of what's going on in a single thread and it needs something like `Result` that says if it is done or if it is blocked. That is needed whether it is hidden or exposed in interp

hs/src/Reach/Debugger.hs
```diff
@@ -59,11 +59,11 @@ type Frontend = M.Map String ([DLVal] -> DLVal)
...
-  = ActionTieBreak
-  | ActionNewAcc
-  | ActionNewPart
-  | ActionImitateFrontend
-  | ActionInteract [DLVal]
+  = Action_TieBreak
+  | Action_NewAcc
+  | Action_NewPart
+  | Action_ImitateFrontend
+  | Action_Interact [DLVal]
```
jeapostrophe, line 66: Sorry for being so sucky. We also use really short prefixes, like `A`. See https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/AST/DLBase.hs for what I mean

hs/src/Reach/Compiler.hs
```diff
@@ -22,6 +22,7 @@ import Reach.Eval
 import Reach.Linearize
 import Reach.Optimize
 import Reach.Parser (gatherDeps_top)
+import Reach.SimServer
```
jeapostrophe, line 25: I think you should make a `Simulator` directory with `Server` and `Eval` (or `Core` or something)

hs/src/Reach/SimClient.hs
```diff
@@ -0,0 +1,37 @@
+{-# OPTIONS_GHC -Wno-missing-export-lists #-}
```
jeapostrophe, line 1: You should write this program in JS

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+type PartCont = Env -> DLVal -> PartState
+-- type Part = Expr
+
+data PartState
+  = PS_Done Env DLVal
+  | PS_Suspend Action Env PartCont
+
+-- Manual Monad
+newtype App a =
+  App (Env -> (Env -> a -> PartState) -> PartState)
```
jeapostrophe, line 33: It's not really an "Env" anymore, it is more of a "Store" or "State"

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+  e <- globalGet
+  let ledger = e_ledger e
+  let token_id = nw_next_token ledger
+  let new_nw_ledger = M.insert 0 (M.singleton token_id supply) (nw_ledger ledger)
+  let new_ledger = ledger { nw_ledger = new_nw_ledger, nw_next_token = token_id + 1 }
+  globalSet $ e {e_ledger = new_ledger}
+  f
+
+data Action
+  = Action_TieBreak
```
jeapostrophe, line 89: `A_TieBreak` to obey our naming scheme

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+  | V_Object (M.Map SLVar DLVal)
+  | V_Data SLVar DLVal
+  | V_Struct [(SLVar, DLVal)]
+  deriving (Eq, Ord, Show, Generic)
+
+instance ToJSON DLVal
+instance FromJSON DLVal
+
+addToStore :: DLVar -> DLVal -> App DLVal -> App DLVal
+addToStore x v app = do
```
jeapostrophe, line 122: There's no reason to have `app` here

jeapostrophe: Because this is just an effect

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+updateLedger :: Account -> Token -> (Integer -> Integer) -> App DLVal -> App DLVal
+updateLedger acc tok f app = do
+  e <- globalGet
+  let ledger = e_ledger e
+  let map_ledger = nw_ledger ledger
+  let prev_amt = flip (M.!) tok $ (M.!) map_ledger acc
+  let new_amt = f prev_amt
+  let new_nw_ledger = M.insert acc (M.insert tok new_amt ((M.!) map_ledger acc)) map_ledger
+  globalSet $ e {e_ledger = ledger { nw_ledger = new_nw_ledger }}
+  app
```
jeapostrophe, line 151: Same with all these

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+        (V_Array arr, V_UInt n) -> return $ arr !! (fromIntegral n)
+        _ -> impossible "expression interpreter"
+    DLE_ArraySet _at dlarg1 dlarg2 dlarg3 -> do
+      ev1 <- interp dlarg1
+      ev2 <- interp dlarg2
+      ev3 <- interp dlarg3
+      case (ev1,ev2) of
+        (V_Array arr, V_UInt n) -> do
+          let n' = fromIntegral n
+          return $ V_Array $ take n' arr ++ [ev3] ++ drop (n' + 1) arr
```
jeapostrophe, line 236: I think we've written this like five times in the code base. You should find those cases and make a little helper function that they and you can all call

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+        _ -> impossible "expression interpreter"
+    DLE_ArrayConcat _at dlarg1 dlarg2 -> do
+      ev1 <- interp dlarg1
+      ev2 <- interp dlarg2
+      case (ev1,ev2) of
+        (V_Array arr1, V_Array arr2) -> return $ V_Array $ arr1 ++ arr2
```
jeapostrophe, line 242: FWIW, I try to always use `<>` in case we need to some day replace our lists with something like sequences

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+        V_Tuple arr -> return $ arr !! (fromIntegral n)
+        _ -> impossible "expression interpreter"
+    DLE_ObjectRef _at dlarg str -> do
+      ev <- interp dlarg
+      case ev of
+        V_Object obj -> return $ obj M.! str
+        _ -> impossible "expression interpreter"
+    DLE_Interact _at _slcxtframes _slpart _str _dltype dlargs -> do
+      args <- mapM interp dlargs
+      suspend $ PS_Suspend (Action_Interact args)
```
jeapostrophe, line 263: You need to know the function name and the type. It might be helpful to show `at` and the frames too for a beautiful debugging experience

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+          case maybe_dlarg of
+            Nothing -> updateLedger acc 0 (+n) $ return V_Null
+            Just tok -> do
+              ev <- interp tok
+              case ev of
+                V_UInt tok' -> updateLedger acc tok' (+n) $ return V_Null
+                _ -> impossible "unexpected error"
+        _ -> impossible "expression interpreter"
+    DLE_TokenInit _at _dlarg -> return V_Null
+    DLE_CheckPay _at _slcxtframes dlarg _maybe_dlarg -> interp dlarg
```
jeapostrophe, line 287: This is no-op

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+    DLE_Wait _at dltimearg -> case dltimearg of
+      Left dlarg -> do
+        ev <- interp dlarg
+        case ev of
+          V_UInt n -> incrNWtime n $ return V_Null
```
jeapostrophe, line 292: This should actually be a blocking moment, because things might happen while someone or the contract is waiting for a certain time. Also `n` is not an AMOUNT to wait it is a TIME TO WAIT UNTIL

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+          let new_nw_ledger = M.insert 0 (M.delete tok ((M.!) map_ledger 0)) map_ledger
+          globalSet $ e {e_ledger = ledger { nw_ledger = new_nw_ledger }}
+          return V_Null
+        _ -> impossible "expression interpreter"
+    -- TODO: new stuff
+    DLE_TimeOrder _at _assoc_maybe_arg_vars -> undefined
+    DLE_GetContract _at -> undefined
+    DLE_GetAddress _at -> undefined
+    DLE_EmitLog at _str dlvar -> interp $ DL_Var at dlvar
+    DLE_setApiDetails _at _slpart _dltpes _maybe_str -> undefined
```
jeapostrophe, line 354: This will be a no-op

hs/src/Reach/Simulator.hs
```diff
@@ -0,0 +1,477 @@
...
+
+instance Interp DLStmt where
+  interp = \case
+    DL_Nop _at -> return V_Null
+    DL_Let _at let_var expr -> case let_var of
+      DLV_Eff -> do
+        _ <- interp expr
```
jeapostrophe, line 361: I like `void $` for this pattern, even though it is more letters, I find it clearer

chikeabuah: i feel like most Haskell people would disagree :)

chikeabuah: based on the GHC warning and a quick reddit search

hs/src/Reach/SimClient.hs
chikeabuah: where in the project  filespace does it go? @jeapostrophe

hs/src/Reach/Simulator/Core.hs
```diff
@@ -50,17 +61,28 @@ instance Monad App where
         let App con = conF vVal
          in con gvVal kAns))
 
-suspend :: (Env -> PartCont -> PartState) -> App DLVal
+unsuspend :: PartState -> State -> DLVal -> PartState
+unsuspend ps st v =
+ case ps of
+   PS_Done _ _ -> do
+     ps
```
jeapostrophe, line 68: I think this should be an error

hs/src/Reach/Simulator/Core.hs
```diff
@@ -276,25 +300,33 @@ instance Interp DLExpr where
         -- TODO: "Remove funds from contract..." which contract?
         (V_UInt n, V_Address acc) -> do
           case maybe_dlarg of
-            Nothing -> updateLedger acc 0 (+n) $ return V_Null
+            Nothing -> do
+              updateLedger acc 0 (+n)
```
jeapostrophe, line 304: It is weird that `updateLedger` is taking a `Num -> Num` function, because it needs to know the amount itself so it can remove from one and give to the other

hs/src/Reach/Simulator/Core.hs
```diff
@@ -326,14 +358,16 @@ instance Interp DLExpr where
       supply <- interp (dtn_supply dltokennew)
       case supply of
         V_UInt supply' -> do
-          ledgerNewToken supply' $ return V_Null
+          ledgerNewToken supply'
```
jeapostrophe, line 361: You should include the rest of the new information for the display

hs/src/Reach/Simulator/Server.hs
```diff
@@ -0,0 +1,154 @@
...
+initProgSim ll = do
+  let initSt = C.initState
+  ps <- return $ C.initApp ll initSt
+  processNewState ps
+
+processNewState :: C.PartState -> WebM C.PartState
+processNewState ps = do
+  sid <- gets e_nsid
+  _ <- case ps of
+    C.PS_Done _ _ -> do
```
jeapostrophe, line 73: There could be a final effect, so you'll want to save the state

hs/src/Reach/Simulator/Server.hs
```diff
@@ -0,0 +1,154 @@
...
+      _ <- return $ putStrLn "EVAL DONE"
+      registerAction sid C.A_None
+    C.PS_Suspend a _ _ -> registerAction sid a
```
jeapostrophe, line 76: There may be a modified state as well

hs/src/Reach/Simulator/Server.hs
```diff
@@ -0,0 +1,154 @@
...
+
+unblockProg :: StateId -> ActionId -> C.DLVal -> WebM ()
+unblockProg sid _aid v = do
+  stks <- gets e_states_ks
+  case M.lookup sid stks of
+    Nothing -> do
+      _ <- return $ putStrLn "previous state not found"
+      return ()
+    Just (C.PS_Suspend _a cst k) -> do
+      let ps = k cst v
```
jeapostrophe, line 107: The state may have changed since `cst` was saved, because another action might have been chosen and its effects would be lost if you used `cst`

hs/src/Reach/Simulator/Core.hs
chikeabuah: it gets called once for each of those

hs/src/Reach/Simulator/Core.hs
```diff
@@ -96,20 +96,28 @@ data Ledger = Ledger
...
   deriving (Eq)
 
+nwToken :: Token
+nwToken = -1
+
+simContract :: Account
+simContract = -1
+
+simContractAmt :: Balance
+simContractAmt = 4444
```
jeapostrophe, line 106: The contract will start with nothing, FWIW. And it doesn't exist until after the first message, so it might not be the "lowest" one. Because of remote objects, you're going to have a way of creating new contracts out there in the world too

hs/src/Reach/Simulator/Core.hs
```diff
@@ -301,12 +312,14 @@ instance Interp DLExpr where
         (V_UInt n, V_Address acc) -> do
           case maybe_dlarg of
             Nothing -> do
-              updateLedger acc 0 (+n)
+              updateLedger simContract nwToken (subtract n)
+              updateLedger acc nwToken (+n)
```
jeapostrophe, line 316: I think it is less error-prone to guarantee the ledger is always balanced by having a `transfer` function

hs/src/Reach/Simulator/Server.hs
chikeabuah: @jeapostrophe yes, we save that later on

hs/src/Reach/Simulator/Core.hs
```diff
@@ -12,13 +12,33 @@ import Data.Bits
...
+-- initLocalState = LocalState
+--   { l_acct = -1
+--   , l_who = Nothing
+--   }
+
+type PartId = Int
+type Locals = M.Map PartId LocalState
+
+-- TODO: state here and in the server should
+-- probably all be IORefs
```
jeapostrophe, line 32: I don't think so because then you'll have to find and duplicate them all. It's a trade-off... do you had complexity trying to duplicate, or do you add complexity functionally updating them? My gut says the first is riskier

hs/src/Reach/Simulator/Core.hs
```diff
@@ -12,13 +12,33 @@ import Data.Bits
 import Data.Aeson (FromJSON, ToJSON)
 import GHC.Generics
 
--- state
+data LocalState = LocalState
+  { l_acct :: Account
+  , l_who :: Maybe SLPart
+  -- TODO QUESTION: partstates for each participant?
```
jeapostrophe, line 18: Don't know what this means, if the question is for me

hs/src/Reach/Simulator/Core.hs
```diff
@@ -28,6 +48,8 @@ initState = State
   , e_linstate = M.empty
   , e_nwtime = 0
   , e_nwsecs = 0
+  , e_locals = M.empty
```
jeapostrophe, line 51: I find that using `mempty` is really convenient because if we change the representation it will just update without having to modify the code

hs/src/Reach/Simulator/Core.hs
```diff
@@ -521,9 +558,21 @@ instance Interp LLStep where
...
+      case v of
+        V_UInt n -> do
+          g <- globalGet
+          let locals = e_locals g
+          let lclst = l_who (locals M.! (fromIntegral n))
+          case lclst of
+            Nothing ->  impossible "expected participant id, received consensus id"
+            Just part -> case tc_send M.! part of
+              DLSend {..} -> do
+                ds_msg' <- mapM interp ds_msg
```
jeapostrophe, line 572: The way you have this won't work in the long-run.

First, simple thing, L572 is interpreting the messages with the consensus store, but they will only be bound in the participant store.

Second, L562 is assuming that anyone can be chosen, but actually there might be a situation where Alice and Bob are "racing", but actually Bob hasn't started the program at all, so you're going to have to track in some way who is actually available.

Third, L567 means that the identifier is the participant id, not the the index into the list in L562. You probably know this, but I would be confused and forget.

Fourth, you're going to need to have some sort of GlobalState counter of which message this is and save it for later in L572. Because, you're going to have to go back to e.g. Alice and run the same `dr_k` in her context, but then maybe 10 years later, when someone runs another instance of Bob, he'll need to see that same publish and run the code too. There's got to be some sort of global log of every call to the contract.

chikeabuah: "Second, L562 is assuming that anyone can be chosen, but actually there might be a situation where Alice and Bob are "racing", but actually Bob hasn't started the program at all, so you're going to have to track in some way who is actually available."

hs/src/Reach/Simulator/Core.hs
```diff
@@ -241,9 +273,13 @@ interpPrim = \case
...
   (SELF_ADDRESS _slpart _bool _int, _) -> do
     g <- globalGet
-    let partid = e_partid g
-    let locals = e_locals g
-    return $ V_Address $ l_acct $ locals M.! partid
+    let actorid = e_curr_actorid g
+    case actorid == (fromIntegral consensusId) of
+      False -> do
+        let locals = e_locals g
+        return $ V_Address $ l_acct $ (fromMaybe (impossible "SELF_ADDRESS") (M.lookup actorid locals))
```
jeapostrophe, line 280: I think part of the local state should just be the answer to this question.

hs/src/Reach/Simulator/Core.hs
```diff
@@ -331,7 +372,7 @@ instance Interp DLExpr where
     DLE_ObjectRef _at dlarg str -> do
       ev <- interp dlarg
       case ev of
-        V_Object obj -> return $ obj M.! str
+        V_Object obj -> return $ (fromMaybe (impossible "DLE_ObjectRef") (M.lookup str obj))
```
jeapostrophe, line 375: This seems like a pattern begging for a helper function

chikeabuah: yes

examples/simulator/client.mjs
```diff
@@ -16,17 +16,13 @@ let getStatus = () => {
...
-//
-// let respondWithVal = (s,a,v) => {
-//   fetch(`${hostname}/states/${s}/actions/${a}/?data=${v}`)
-//     .then(response => response.json())
-//     .then(data => console.log(data));
-// }
+async function getStateActions(s) {
+  const r = await fetch(`${hostname}/states/${s}/actions`)
+    .then(response => response.json())
+    .then(data => data);
```
jeapostrophe, line 22: You should use the full `await` style here and in this code generally IMHO

examples/simulator/client.mjs
```diff
@@ -58,6 +54,13 @@ let init = () => {
     });
 }
 
+let respondWithVal = (s,a,v,t) => {
```
jeapostrophe, line 57: Yo should use `const` rather than `let`, because I don't think you are actually mutating

hs/src/Reach/Simulator/Core.hs
```diff
@@ -31,34 +29,48 @@ data LocalState = LocalState
...
-  , e_curr_actorid :: ActorId
-  , e_nactid :: ActorId
+  , e_nactorid :: ActorId
   , e_naccid :: AccountId
   , e_nmsgid :: PoolId
+  , e_partacts :: M.Map Participant ActorId
   }
 
+data Local = Local
+  { e_locals :: Locals
```
jeapostrophe, line 47: Quibble: I don't understand why this isn't `LocalInfo`. In other words, I think that `Locals` is something that only lives in `Server.hs` not `Core.hs`

chikeabuah: yeah

chikeabuah: actually i'd rather not do this right now because: the Locals in the Server.hs could become out of sync with the LocalInfo in Core.hs if i do

examples/simulator-1/client.mjs
```diff
@@ -0,0 +1,104 @@
...
+    .then(response => response.json())
+    .then(data => data);
+  console.log(r)
+  return r
+}
+
+const sleep = (delay) => new Promise((resolve) => setTimeout(resolve, delay))
+
+const main = async () => {
+  await sleep(10000)
```
jeapostrophe, line 80: Why sleeping?

examples/simulator-1/Makefile
```diff
@@ -0,0 +1,14 @@
+REACH = ../../reach
+NODE = node
+
+.PHONY: build
+build:
+	$(REACH) compile --sim &
```
jeapostrophe, line 6: I think is really part of the `run`

examples/simulator-1/Makefile
```diff
@@ -0,0 +1,14 @@
...
+
+.PHONY: run
+run:
+	$(NODE) client.mjs
```
jeapostrophe, line 10: Do you know if the CI built this for you?

examples/simulator-1/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM reachsh/runner:0.1
```
jeapostrophe, line 1: This should be `latest` so it uses the one that was built by the CI

examples/simulator-1/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+ENV NODE_VERSION=16.13.0
```
jeapostrophe, line 3: This should use the one in `/DEPS` by inclusion, but actually `runner` already has `node`, so you should be able to use it

examples/simulator-1/client.mjs
```diff
@@ -1,28 +1,31 @@
...
-const hostname = "http://localhost:3000"
+const hostname = "http://localhost"
+const port = 3001
+const address = `${hostname}:${port}`
 
 const getStates = async () => {
-  const r = fetch(`${hostname}/states`)
+  const r = fetch(`${address}/states`)
     .then(response => response.json())
     .then(data => data);
```
jeapostrophe, line 14: I continue to think all these `.then`s should be replace with `await` style. This function, for instance, is wrong because you don't `await` the `fetch`

chikeabuah: oh sorry

jeapostrophe (approved):

chikeabuah:
gottem :)

examples/simulator-1/client.mjs
hamirmahal: What do the seven `.then(data => data);` in this file do?

Is it safe to remove them?

jeapostrophe:
They don't do anything and should definitely be removed :)

PR #237, opened 2021-08-03
Core return optimization

tippenein:
- [x] shallow comparison for `SLVal`s
- [x] proof via simple reach program


hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
   | SLV_Deprecated Deprecation SLVal
   deriving (Eq, Generic)
 
+-- | Equivalence operation on flattened or simplified structures.
+class Equiv a where
+  equiv :: a -> a -> Bool
```
tippenein, line 165: there's probably no reason to typeclass this atm but I figured semantically there is some use for a shallow equivalence class? Otherwise this would be fine as a simple function

docs-src/ref-programs-step.scrbl
```diff
@@ -247,7 +247,7 @@ fork()
 }
 
 where:
-@reachin{TOKENS_EXPR} is an expression that evalues to a tuple of @reachin{Token}s.
+@reachin{TOKENS_EXPR} is an expression that evaluates to a tuple of @reachin{Token}s.
```
tippenein, line 250: noticed this while reading through the docs

jeapostrophe: Best to submit these separately for easy merging :) or just push 'em when you have the power

hs/src/Reach/AST/SL.hs
jeapostrophe: Yea, I think it should be a class

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+
+instance Equiv SLVal where
+  equiv (SLV_Null _ s1) (SLV_Null _ s2) = s1 == s2
```
jeapostrophe, line 168: Naw, we'd ignore the strings. It's basically provenance information

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+  equiv (SLV_Bool _ b1) (SLV_Bool _ b2) = b1 == b2
+  equiv (SLV_Int _ i1) (SLV_Int _ i2) = i1 == i2
+  equiv (SLV_Bytes _ v1) (SLV_Bytes _ v2) = v1 == v2
+  equiv (SLV_Array _ t xs) (SLV_Array _ t2 ys) = t == t2 && xs == ys
```
jeapostrophe, line 172: You shouldn't need to compare the types

jeapostrophe: But you'd equiv the values

jeapostrophe: That is, you'd implement equiv for foldables/etc

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+  equiv (SLV_Tuple _ v1) (SLV_Tuple _ v2) = v1 == v2
+  equiv (SLV_Struct _ xs) (SLV_Struct _ ys) = xs == ys
+  equiv (SLV_DLC d1) (SLV_DLC d2) = d1 == d2
+  equiv (SLV_DLVar d1) (SLV_DLVar d2) = d1 == d2
```
jeapostrophe, line 176: They may have different source location information, so you should implement equiv

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+class Equiv a where
+  equiv :: a -> a -> Bool
+
+instance Equiv SLVal where
+  equiv (SLV_Null _ s1) (SLV_Null _ s2) = s1 == s2
+  equiv (SLV_Bool _ b1) (SLV_Bool _ b2) = b1 == b2
+  equiv (SLV_Int _ i1) (SLV_Int _ i2) = i1 == i2
+  equiv (SLV_Bytes _ v1) (SLV_Bytes _ v2) = v1 == v2
+  equiv (SLV_Array _ t xs) (SLV_Array _ t2 ys) = t == t2 && xs == ys
+  equiv (SLV_Tuple _ v1) (SLV_Tuple _ v2) = v1 == v2
```
jeapostrophe, line 173: This should use equiv, because a tuple contains slvals

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+  equiv (SLV_Struct _ xs) (SLV_Struct _ ys) = xs == ys
```
jeapostrophe, line 174: Ditto

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+  equiv (SLV_Kwd k1) (SLV_Kwd k2)  = k1 == k2
+  equiv (SLV_Deprecated d v) (SLV_Deprecated d2 v2) = d == d2 && v == v2
+  equiv SLV_Anybody SLV_Anybody = True
+
+  -- un-inspected structures
+
+  equiv (SLV_Participant _ _ _ _) _ = False
+  equiv (SLV_RaceParticipant _ _) _ = False
+  equiv (SLV_Map _) _ = False
+  equiv (SLV_Data _ _ _ _) _ = False
```
jeapostrophe, line 188: We do want to look into these

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+  equiv (SLV_Form _) _ = False
+  equiv (SLV_Type _) _ = False
+  equiv (SLV_Object _ _ _) _ = False
```
jeapostrophe, line 191: And these

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+  equiv SLV_Anybody SLV_Anybody = True
+
+  -- un-inspected structures
+
+  equiv (SLV_Participant _ _ _ _) _ = False
+  equiv (SLV_RaceParticipant _ _) _ = False
+  equiv (SLV_Map _) _ = False
+  equiv (SLV_Data _ _ _ _) _ = False
+  equiv (SLV_Form _) _ = False
+  equiv (SLV_Type _) _ = False
```
jeapostrophe, line 190: And these

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+  equiv (SLV_Deprecated d v) (SLV_Deprecated d2 v2) = d == d2 && v == v2
+  equiv SLV_Anybody SLV_Anybody = True
+
+  -- un-inspected structures
+
+  equiv (SLV_Participant _ _ _ _) _ = False
+  equiv (SLV_RaceParticipant _ _) _ = False
+  equiv (SLV_Map _) _ = False
+  equiv (SLV_Data _ _ _ _) _ = False
+  equiv (SLV_Form _) _ = False
```
jeapostrophe, line 189: These are like prim and can be inspected

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,38 @@ data SLVal
...
+  equiv (SLV_Prim p1) (SLV_Prim p2) = p1 == p2
+  equiv (SLV_Kwd k1) (SLV_Kwd k2)  = k1 == k2
+  equiv (SLV_Deprecated d v) (SLV_Deprecated d2 v2) = d == d2 && v == v2
+  equiv SLV_Anybody SLV_Anybody = True
+
+  -- un-inspected structures
+
+  equiv (SLV_Participant _ _ _ _) _ = False
+  equiv (SLV_RaceParticipant _ _) _ = False
+  equiv (SLV_Map _) _ = False
```
jeapostrophe, line 187: Everything is worth looking at EXCEPT closures

hs/src/Reach/Eval/Core.hs
```diff
@@ -3095,21 +3095,21 @@ evalApplyClosureVals clo_at (SLClo mname formals (JSBlock body_a body _) SLCloEn
...
-                    --- we're relying on the invariant that there
-                    --- was only one Return... this should be
-                    --- true, but if something changes in the
-                    --- future, this might be a place that an
-                    --- error could be introduced.
-                    body_lifts'
-                _ ->
-                  return $ DLS_Prompt body_at (Left ret) body_lifts
-        saveLifts lifts'
+  let no_prompt :: (SecurityLevel, SLVal) -> ReaderT Env IO SLAppRes
```
jeapostrophe, line 3098: Use the `App` abbreviation

hs/src/Reach/Eval/Core.hs
```diff
@@ -3119,11 +3119,9 @@ evalApplyClosureVals clo_at (SLClo mname formals (JSBlock body_a body _) SLCloEn
...
-      -- Note: This test might be too expensive, so try it
-      -- chicken-fork and many other fork-ing programs would be improved by
-      -- this when all pays are zero, but it doesn't work because ints have a
-      -- srcloc
-      let all_same = False && all (== xv) mvs
+      -- We check equivalence here with Equiv typeclass. it's shallow checking but performant
+      let all_same = all (equiv xv) mvs
+      -- let all_same = False && all (equiv xv) mvs
       case all_same of
         True -> no_prompt $ (lvl, xv)
```
jeapostrophe, line 3126: I realized recently that it is actually dangerous to remove the prompt in this situation (the case in `no_prompt`) because the final return might be at the end where there was an early return. The key point then is to (a) return the `xv` and (b) include the prompt in the dynamic code

jeapostrophe (changes requested):

tippenein:
There were 2 SLForm constructors I wasn't sure were possible to equiv, mainly because they're largely `JSExpression`:

- `SLForm_parallel_reduce_partial`
- `SLForm_fork_partial`

docs-src/ref-programs-step.scrbl
```diff
@@ -247,7 +247,7 @@ fork()
 }
```
jeapostrophe: The revert is going to be a pain. I would do an interactive rebase and delete the commit

hs/src/Reach/AST/SL.hs
```diff
@@ -1,4 +1,5 @@
 {-# OPTIONS_GHC -Wno-missing-export-lists #-}
+{-# Language DefaultSignatures #-}
```
jeapostrophe, line 2: What does this do? Should we add it to the standard list?

hs/src/Reach/AST/SL.hs
```diff
@@ -161,15 +162,74 @@ data SLVal
   deriving (Eq, Generic)
 
 -- | Equivalence operation on flattened or simplified structures.
-class Equiv a where
+class (Eq a) => Equiv a where
   equiv :: a -> a -> Bool
+  default equiv :: (a -> a -> Bool)
+  equiv = (==)
```
jeapostrophe, line 168: Won't this make it so that default instances use `==` internally rather than `equiv`?

hs/src/Reach/AST/SL.hs
```diff
@@ -161,15 +162,74 @@ data SLVal
...
+
+instance (Equiv a) => Equiv (Maybe a)
```
jeapostrophe, line 170: Meaning that this says "If a supports Equiv, then Maybe a supports Equiv, by using == on the a instance (ignoring the equiv definition)?"

hs/src/Reach/AST/SL.hs
```diff
@@ -161,15 +162,74 @@ data SLVal
...
+instance Equiv SLTypeFun
+instance Equiv SLSSVal
+instance Equiv DLMVar
+instance (Equiv a) => Equiv [a]
+instance (Equiv b, Equiv a) => Equiv (M.Map a b) where
+  equiv = (==)
+instance (Equiv a, Equiv b) => Equiv (a, b)
+
+instance Equiv SLForm where
+  equiv SLForm_App SLForm_App = True
```
jeapostrophe, line 186: We use `\case` rather than repeating the name of the function

hs/src/Reach/AST/SL.hs
```diff
@@ -161,15 +162,74 @@ data SLVal
...
+  equiv SLForm_each SLForm_each = True
+  equiv (SLForm_EachAns parts _ _ _) (SLForm_EachAns parts2 _ _ _) = equiv parts parts2
+  equiv (SLForm_Part_Only p _) (SLForm_Part_Only p2 _) = equiv p p2
+  equiv (SLForm_liftInteract p _ _) (SLForm_liftInteract p2 _ _) = equiv p p2
+  equiv (SLForm_Part_ToConsensus {slptc_whos = who, slptc_mv = mv}) (SLForm_Part_ToConsensus {slptc_whos = who2, slptc_mv = mv2}) =
+    equiv who who2 && equiv mv mv2
+  equiv SLForm_unknowable SLForm_unknowable = True
+  equiv SLForm_fork SLForm_fork = True
+  -- XXX Seems impossible to do this one?
```
jeapostrophe, line 195: Yea, I agree

hs/src/Reach/AST/SL.hs
```diff
@@ -161,15 +162,74 @@ data SLVal
...
+  equiv SLForm_fork_partial{} _ = False
+  equiv SLForm_parallel_reduce SLForm_parallel_reduce = True
+  -- XXX Seems impossible to do this one?
```
jeapostrophe, line 198: Ditto

hs/src/Reach/AST/SL.hs
```diff
@@ -161,15 +162,74 @@ data SLVal
...
+  equiv SLForm_parallel_reduce_partial{} _ = False
+  equiv SLForm_wait SLForm_wait = True
+  equiv _ _ = False
+
+instance Equiv SLType where
+  equiv ST_Null ST_Null = True
```
jeapostrophe, line 204: `\case`

jeapostrophe (changes requested):

hs/src/Reach/AST/SL.hs
tippenein: Yes. This is probably dangerous. It allowed me to write 
```
instance Equiv Int
```

instead of 

``` 
instance Equiv Int where
  equiv = (==)
```

for a bunch of primitive stuff

hs/src/Reach/AST/SL.hs
tippenein: it allows line 167 but will probably delete this anyway based on the comment below this

hs/src/Reach/Eval/Core.hs
tippenein: I opened an issue for this one since I'm not quite sure how to accomplish that: https://trello.com/c/fLRQSHyq/1471-fix-noprompt-safety-in-core

jeapostrophe: Make a function called something like `no_prompt'` that does the second half of `no_prompt`. Have `no_prompt` call it and have this `all_same` case call `no_promp'`. That way we will only ever remove the prompt if there was a single return

hs/src/Reach/AST/SL.hs
```diff
@@ -164,94 +163,120 @@ data SLVal
...
-  equiv _ _ = False
+  equiv a b = case (a,b) of
+    ((SLV_Null _ _), (SLV_Null _ _)) -> True
+    ((SLV_Bool _ b1), (SLV_Bool _ b2)) -> b1 == b2
+    ((SLV_Int _ i1), (SLV_Int _ i2)) -> i1 == i2
+    ((SLV_Bytes _ v1), (SLV_Bytes _ v2)) -> v1 == v2
+    -- Array types can be ignored
+    ((SLV_Array _ _ xs), (SLV_Array _ _ ys)) -> all (\(x,y) -> equiv x y) $ zip xs ys
+    ((SLV_Tuple _ v1), (SLV_Tuple _ v2)) -> v1 == v2
+    ((SLV_Struct _ xs), (SLV_Struct _ ys)) -> xs == ys
```
jeapostrophe, line 262: These need equiv

hs/src/Reach/AST/SL.hs
```diff
@@ -164,94 +163,120 @@ data SLVal
...
-  -- Closures are the only uninspected structures
-  equiv (SLV_Clo _ _ _) _ = False
-  equiv _ _ = False
+  equiv a b = case (a,b) of
+    ((SLV_Null _ _), (SLV_Null _ _)) -> True
+    ((SLV_Bool _ b1), (SLV_Bool _ b2)) -> b1 == b2
+    ((SLV_Int _ i1), (SLV_Int _ i2)) -> i1 == i2
+    ((SLV_Bytes _ v1), (SLV_Bytes _ v2)) -> v1 == v2
+    -- Array types can be ignored
+    ((SLV_Array _ _ xs), (SLV_Array _ _ ys)) -> all (\(x,y) -> equiv x y) $ zip xs ys
```
jeapostrophe, line 260: This should just leverage the list `equiv` instance

hs/src/Reach/AST/SL.hs
```diff
@@ -164,94 +163,120 @@ data SLVal
...
+    ((SLV_DLVar d1), (SLV_DLVar d2)) -> equiv d1 d2
+    ((SLV_Connector t1), (SLV_Connector t2)) -> t1 == t2
+    ((SLV_Prim p1), (SLV_Prim p2)) -> p1 == p2
+    ((SLV_Kwd k1), (SLV_Kwd k2))  -> k1 == k2
+    ((SLV_Deprecated d v), (SLV_Deprecated d2 v2)) -> d == d2 && v == v2
+    (SLV_Anybody, SLV_Anybody) -> True
+    ((SLV_Participant _ s sl dl), (SLV_Participant _ s2 sl2 dl2)) -> equiv s s2 && equiv sl sl2 && equiv dl dl2
+    ((SLV_RaceParticipant _ slSet1), (SLV_RaceParticipant _ slSet2)) -> equiv slSet1 slSet2
+    ((SLV_Map v1), (SLV_Map v2)) -> equiv v1 v2
+    ((SLV_Data _ _ _ _), _) -> False
```
jeapostrophe, line 273: You should compare data instances

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
```diff
@@ -3095,6 +3095,7 @@ evalApplyClosureVals clo_at (SLClo mname formals (JSBlock body_a body _) SLCloEn
         locAt body_at $
           locSco clo_sco' $
             evalStmt body
+  let no_prompt' = SLAppRes clo_sco''
```
jeapostrophe, line 3098: This should be something like:

```
let saveAndRet lifts (lvl, v) = do
  saveLifts lifts
  return $ SLAppRes clo_sco'' (lvl, v)
let promptAndStatic = saveAndRet $ DLS_Prompt body_at (Left ret) body_lifts
let maybeNoPromptStatic =
  case body_lifts of
    .... -> saveAndRet body_lifts'
    _ -> promptAndStatic
...
case all_same of
 True -> promptAndStatic
 False ->
  ...
```

hs/src/Reach/AST/SL.hs
tippenein: `SLV_Data SrcLoc (M.Map SLVar DLType) SLVar SLVal`
Do we need to compare anything other than the `SLVal` here?

jeapostrophe: You should compare the last three elements. They are: the variants, the variant, and then the value. For example, we don't want `Left 1` to match `Left 2`, so we compare the last. We don't want `Left 1` to match `Right 1`, so we compare the second to last. And, we don't want `X.Left 1` to match `Y.Left 1` where `X = Left Int | Right Int` and `Y = Left Int | Right Bool`, because they aren't the same kind of `Left`

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,127 @@ data SLVal
...
+    ((ST_Struct s), (ST_Struct s2)) -> equiv s s2
+    ((ST_Fun f), (ST_Fun f2)) -> equiv f f2
+    ((ST_UDFun f), (ST_UDFun f2)) -> equiv f f2
+    ((ST_Type s1), (ST_Type s2)) -> equiv s1 s2
+    ((ST_Refine _sLType v _m), (ST_Refine _sLType2 v2 _m2)) -> equiv v v2
+    _ -> False
+
+instance Equiv DLVar where
+  equiv (DLVar _ sl _dl i) (DLVar _ sl2 _dl2 i2) =
+    equiv (fmap snd sl) (fmap snd sl2) && i == i2
```
jeapostrophe, line 251: We actually only care about the numbers. Also, I like to have everything have a suffix if anything does, so I'd do `i1` and `i2` or maybe `x` and `y`

hs/src/Reach/AST/SL.hs
```diff
@@ -182,12 +182,46 @@ instance Equiv Char where
...
+
+instance Equiv FluidVar where
+  equiv = (==)
+
+instance Equiv DLArg where
+  equiv a b = case (a,b) of
+    (DLA_Var v1, DLA_Var v2) -> equiv v1 v2
+    (DLA_Constant c1, DLA_Constant c2) -> equiv c1 c2
+    (DLA_Literal lit1, DLA_Literal lit2) -> equiv lit1 lit2
+    (DLA_Interact part1 s1 t1, DLA_Interact part2 s2 t2) -> equiv part1 part2 && equiv s1 s2 && equiv t1 t2
```
jeapostrophe, line 214: You can skip comparing `t1` and `t2` because they're invariant for the same `part/s`

hs/src/Reach/AST/SL.hs
```diff
@@ -274,14 +314,14 @@ instance Equiv SLVal where
...
     ((SLV_DLC d1), (SLV_DLC d2)) -> equiv d1 d2
     ((SLV_DLVar d1), (SLV_DLVar d2)) -> equiv d1 d2
-    ((SLV_Connector t1), (SLV_Connector t2)) -> t1 == t2
-    ((SLV_Prim p1), (SLV_Prim p2)) -> p1 == p2
-    ((SLV_Kwd k1), (SLV_Kwd k2))  -> k1 == k2
-    ((SLV_Deprecated d v), (SLV_Deprecated d2 v2)) -> d == d2 && v == v2
+    ((SLV_Connector t1), (SLV_Connector t2)) -> equiv t1 t2
+    ((SLV_Prim p1), (SLV_Prim p2)) -> equiv p1 p2
+    ((SLV_Kwd k1), (SLV_Kwd k2))  -> equiv k1 k2
+    ((SLV_Deprecated d1 v1), (SLV_Deprecated d2 v2)) -> equiv d1 d2 && equiv v1 v2
```
jeapostrophe, line 324: I know you worked hard for this... but you can ignore the deprecation reason

hs/src/Reach/AST/SL.hs
```diff
@@ -607,6 +650,81 @@ data SLPrimitive
   | SLPrim_Token_destroyed
   deriving (Eq, Generic)
 
+instance Equiv SLPrimitive where
+  equiv a b = case (a,b) of
```
jeapostrophe, line 654: This case sucks so bad... I don't want to maintain it... I want to ignore it all except for the primop case. I know you worked hard but sometimes when you see the result you learn that it was a bad idea to send the soldiers into that valley

tippenein: lol it's brutal yea. I think it's not so bad to require adding a new case to the Equiv instance for SLPrimitive if we ever need it. Currently it's defined 
```
instance Equiv SLPrimitive where
  equiv a b = case (a,b) of
    (SLPrim_op p1, SLPrim_op p2) -> equiv p1 p2
    _ -> False
```
which seems less bad. I'll make a comment in the code to explain what's going on here

jeapostrophe: Maybe there's a special thing that Generics could do: Could we do something like "If a and b are the same constructor and they have no parameters, then compare them with `==`?"

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,188 @@ data SLVal
   | SLV_Deprecated Deprecation SLVal
   deriving (Eq, Generic)
 
+class NewerVar a where
+  containsVarNewerThan :: Int -> a -> Bool
+
+instance NewerVar SLVal where
+  containsVarNewerThan ret (SLV_DLVar (DLVar _ _ _ i)) = ret < i
+  containsVarNewerThan _ _ = False
```
jeapostrophe, line 168: Sadly, this needs to recur inside of all the various SLVals looking for embedded DLVars. For example, if the `SLVal` were a tuple and the second element were a DLVar. So... you need to go searching through 'em all and implement many instances of `NewerVar`.... probably for all the same structures as `Equiv`

jeapostrophe: Any time when `Equiv` would give up, this should say that there IS a variable and thus it fails to unify... i.e. it is a conservative check

hs/src/Reach/Eval/Core.hs
```diff
@@ -3096,42 +3096,48 @@ evalApplyClosureVals clo_at (SLClo mname formals (JSBlock body_a body _) SLCloEn
...
-      -- Note: This test might be too expensive, so try it
-      -- chicken-fork and many other fork-ing programs would be improved by
-      -- this when all pays are zero, but it doesn't work because ints have a
-      -- srcloc
-      let all_same = False && all (== xv) mvs
-      case all_same of
-        True -> no_prompt mt $ (lvl, xv)
+      -- We check equivalence here with Equiv typeclass. it's shallow checking but performant
+      let all_same = all (equiv xv) mvs
+      case all_same && not (containsVarNewerThan ret xv) of
```
jeapostrophe, line 3138: I assume the thing on the right is cheaper than the thing on the left and so we should give ghc a hand making it fast

hs/src/Reach/AST/SL.hs
tippenein: Definitely. We can do a lot in Core with Generics. Once I get this passing I'll try that out. btw I think it would also work with constructors with 1 value, because it could assume `(Something v1, Something v2) -> equiv v1 v2` if we restrict the fields to having `Equiv` instances. In fact, we could probably define something smarter for all cases by just defining how to access the "equivable" fields with a lens or simple accessor.
```
class Equiv a where
  accessors :: [(sum -> f)] -> [(f1,f2)]
  equiv :: sum -> sum -> Bool
  default equiv a b = all (equiv . _) . accessors a <> accessors b
  
instance Equiv Something where
  accessors = [\(Var _ _ v) -> v]
```

jeapostrophe: Let's not get too excited ;)

tippenein: lol for sure. I'd actually be more interested in combining the `NewerVar` and `Equiv` classes into one class, but I'm not really sure if that makes sense semantically. Either way, it's probably fine for now

hs/src/Reach/AST/SL.hs
tippenein: I added the constructors which had nested SLVals. It wasn't too many of them, so maybe I'm misunderstanding the tree we should be inspecting

jeapostrophe (commented):
Yea, it looks pretty good. I think my taste would be to have a list instance rather than duplicating `all` though

jeapostrophe (commented):
I believe we're just waiting for an example that confirms it really happens?

examples/return-opt/index.rsh
```diff
@@ -0,0 +1,28 @@
+'reach 0.1';
```
jeapostrophe, line 1: When you run `make hs-test`, this will error because there's no `index.txt` in the repo

hs/src/Reach/AST/SL.hs
```diff
@@ -160,6 +160,201 @@ data SLVal
   | SLV_Deprecated Deprecation SLVal
   deriving (Eq, Generic)
 
+class NewerVar a where
+  containsVarNewerThan :: Int -> a -> Bool
+
+instance (NewerVar a) => NewerVar [a] where
+  containsVarNewerThan ret = all (containsVarNewerThan ret)
+
+instance (NewerVar b) => NewerVar (a,b) where
```
jeapostrophe, line 169: This feels dangerous... because there are some structures in Reach that have real information on both sides, I think that the client (`Struct`?) should use `map snd`

jeapostrophe (changes requested):

hs/t/n/mint-fail2.txt
```diff
@@ -9,17 +9,17 @@ Verification failed:
 
   // Violation Witness
 
-  const UInt.max = 30880;
+  const UInt.max = 35424;
```
jeapostrophe, line 12: I bet these numbers are different because you have a different Z3 than the docker image/CI

examples/return-opt/index.txt
```diff
@@ -0,0 +1,7 @@
+Verifying knowledge assertions
+Verifying for generic connector
+  Verifying when ALL participants are honest
+  Verifying when NO participants are honest
```
jeapostrophe, line 4: You also need an `index.mjs` for the CI :(

hs/t/n/mint-fail2.txt
tippenein: yea, was just looking at this. I'll rollback the change and ignore

jeapostrophe (approved):
Ready to merge pending CI

jeapostrophe:
I'm about to push a merged version of this

PR #247, opened 2021-08-11
Fix `match` Case Check

### LINKED ISSUE #246: RE0073 on match statement.
The following experimental program will throw 

 ```
reachc: error[RE0073]: Switch contains duplicate case, deposit at ./index.rsh:74:60:case; first defined at ./index.rsh:74:60:case
```

when deciding for a variant **or** default, in a publish expression of a Parallel Reduce loop.

It is a bug or something I missed?
Version is latest reach at this time of writing.
Thank you.

```

'reach 0.1';
'use strict';

//
// Command is a variant-type (Data) containing the
// available commands to execute
//
const Command = Data({
  deposit: Tuple(UInt, Token)
})

//
// Interaction interfaces
//
const ICommon = {

  // Get command from front-end
  getCommand: Fun([], Command)
}

const IUser = {
  ...ICommon
}

const AppParams = Object({
  randDCTokenAddress: Token
})

const IGovernor = {
  ...ICommon,
  getParams: Fun([], AppParams)
}


// 
// Evaluates publish-expression and returns proper object
// if command matches.
//
const EvalCommand = (what) => {

}

export const main = Reach.App(
  {
    deployMode: 'constructor',
    verifyArithmetic: true  // This should be disabled in production
  },
  [
    ParticipantClass('User', IUser),
    Participant('Governor', IGovernor)
  ], (User, Governor) => {

    // Require governor to set some parameters
    //
    Governor.only(() => {
      const randDCToken = declassify(interact.getParams()).randDCTokenAddress;
    });
    Governor.publish(randDCToken)

    const [running] =
      parallelReduce([true])
        .invariant(balance() == 0)
        .while(running)
        .case(
          User,
          (() => ({
            when: (declassify(interact.getCommand())).match({
              deposit: () => { true },
              default: (x) => { false }
            }),
            msg: cmd
          })
          ),
          ((msg) => ([msg[0], msg[1]]),  // Token and amount.
            ((msg) => {
              transfer(1, randDCToken).to(User);
              commit();
            })))
        .timeout(false)
  }
);
```

chrisnevers:
Fixes #246 

* When checking the `match` cases, order the `case` names according to `SrcLoc` as opposed to ascending order. This way `default` behaves as expected.
* Give better error messages when a `case` occurs after `default`.

### DIFF
--- a/docs-src/ref-error-codes.scrbl
+++ b/docs-src/ref-error-codes.scrbl
@@ -2043,6 +2043,29 @@ It should be corrected by moving the tail of the @reachin{if} into the @reachin{
     };
 }
 
+@error{RE0118}
+
+This error indicates that a switch case is unreachable. This issue will occur when a @reachin{case} is listed
+after @reachin{default}.
+
+For example, the code below erroneously puts a @reachin{case} after @reachin{default}:
+
+@reach{
+  Maybe(UInt).Some(5).match({
+    default: (() => 0),
+    Some: ((i) => i),
+  });
+}
+
+This error can be corrected by either removing the @reachin{Some} case or placing it before the @reachin{default} case:
+
+@reach{
+  Maybe(UInt).Some(5).match({
+    Some: ((i) => i),
+    default: (() => 0),
+  });
+}
+
 @error{REP0000}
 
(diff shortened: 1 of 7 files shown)


jeapostrophe (approved):
Much better than my plan :)

PR #248, opened 2021-08-12
continue consensus mode

tippenein:
> Eval continue RHS before checking mode to allow more programs, such as a function call on the RHS, which does communication and comes back to the consensus state

### DIFF
--- a/docs-src/ref-programs-consensus.scrbl
+++ b/docs-src/ref-programs-consensus.scrbl
@@ -115,7 +115,7 @@ A @tech{continue statement} may be written without the preceding identifier upda
  [] = [];
  continue; }
 
-A @tech{continue statement} must be @tech{dominated} by a @tech{consensus transfer}, which means that the body of a @tech{while statement} must always @reachin{commit();} before calling @reachin{continue;}.
+A @tech{continue statement} must be @tech{dominated} by a @tech{consensus transfer}, which means that the body of a @tech{while statement} must always @reachin{commit();} before calling @reachin{continue;}. A continue statement may occur in a step, provided the @reachin{RHS} modifies the state into a consensus step.
 This restriction may be lifted in future versions of Reach, which will perform termination checking.
 
 @subsection{@tt{parallelReduce}}
@@ -438,4 +438,3 @@ set, @reachin{s}, or @reachin{s.remove(ADDRESS)} to remove the @reachin{ADDRESS}
 Such modifications may only occur in a @tech{consensus step}.
 
 @reachin{s.member(ADDRESS)} will return a @reachin{Bool} representing whether the address is in the set.
-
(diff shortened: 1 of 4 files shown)


hs/t/y/continue-mode.rsh
```diff
@@ -0,0 +1,24 @@
...
+  const Alice = Participant('Alice', {
+    choice: UInt,
+  });
+  deploy();
+  f(Alice);
+  var keepGoing = true;
+  invariant(keepGoing);
+  while (keepGoing) {
+    commit();
+    f(Alice)
```
jeapostrophe, line 20: We're looking for `keepGoing = f();` where you are in step before `f()` and consensus after `f()` (i.e. inside it is a publish). You don't need to pass Alice... just define the function in this scope right here

hs/src/Reach/Eval/Core.hs
```diff
@@ -4676,15 +4676,15 @@ evalStmt = \case
...
-        ensure_mode SLM_ConsensusStep lab
         let var_at = srcloc_jsa lab var_a
         rhs_sv <- locAtf var_at $ evalExpr rhs
+        ensure_mode SLM_ConsensusStep lab
         sco <- e_sco <$> ask
         locAtf var_at $ do
           whilem <-
             case sco_while_vars sco of
               Nothing -> locAtf (srcloc_jsa lab cont_a) $ expect_ $ Err_Eval_ContinueNotInWhile
-              Just x -> return $ x
```
jeapostrophe, line 4687: You'll notice I put a lot of these `$` in on returns because sometimes I need to change them to a longer expression and then I need it

jeapostrophe (changes requested):

jeapostrophe (changes requested):
This looks right to me, but I haven't seen the CI results and you need to add the output file. I would also simplify it by removing the `choice` and extra newlines and stuff so it is as small as possible.

jeapostrophe:
I can't tell if you consider this done...?

tippenein:
Yeah this is done now

jeapostrophe:
Okay, in that case, you need to update the docs to explain the new rule. This is a test :)

docs-src/ref-programs-consensus.scrbl
```diff
@@ -115,8 +115,7 @@ A @tech{continue statement} may be written without the preceding identifier upda
  [] = [];
  continue; }
 
-A @tech{continue statement} must be @tech{dominated} by a @tech{consensus transfer}, which means that the body of a @tech{while statement} must always @reachin{commit();} before calling @reachin{continue;}.
-This restriction may be lifted in future versions of Reach, which will perform termination checking.
+A @tech{continue statement} must perform a @tech{consensus transfer} before calling @reachin{continue;}.
```
jeapostrophe, line 118: The original sentences are good.

The thing that your commit does is it makes so that a continue statement, despite being listed in the consensus step section, can actually occur in a step, provided the RHS modifies the state into a consensus step.

jeapostrophe (changes requested):

jeapostrophe:
Merged

PR #249, opened 2021-08-12
small compile refactors

tippenein:
I was running through the compile step to understand it better. These were some simplifications I made. Totally fine with taking or leaving these. Thought I'd open it if nothing else for a convo on style

### DIFF
--- a/hs/src/Reach/Compiler.hs
+++ b/hs/src/Reach/Compiler.hs
@@ -43,41 +43,31 @@ all_connectors =
       ]
 
 compile :: CompilerOpts -> IO ()
-compile copts = do
-  let outn = output copts
-  let outnMay outn_ = case intermediateFiles copts of
-        True -> Just outn_
-        False -> Nothing
+compile CompilerOpts {output, dirDotReach, intermediateFiles, installPkgs, source, tops} = do
+  let outnMay = flip doIf intermediateFiles
   let interOut outn_ = case outnMay outn_ of
         Just f -> LTIO.writeFile . f
         Nothing -> \_ _ -> return ()
-  dirDotReach' <- makeAbsolute $ dirDotReach copts
-  let doPkgs = installPkgs copts
-  djp <- gatherDeps_top (source copts) doPkgs dirDotReach'
+  dirDotReach' <- makeAbsolute dirDotReach
+
+  djp <- gatherDeps_top source installPkgs dirDotReach'
   -- interOut outn "bundle.js" $ render $ pretty djp
-  unless doPkgs $ do
+  unless installPkgs $ do
     (avail, compileDApp) <- evalBundle all_connectors djp
-    let chosen = fromMaybe avail $ tops copts
-    forM_ (S.toAscList chosen) $ \which -> do
-      let addWhich = ((T.pack which <> ".") <>)
-      let woutn = outn . addWhich
+    let chosen = S.toAscList $ fromMaybe avail tops
+    forM_ chosen $ \which -> void $ do
+      let woutn = output . ((T.pack which <> ".") <>)
       let woutnMay = outnMay woutn
-      let winterOut = interOut woutn
-      let showp :: (forall a. Pretty a => T.Text -> a -> IO ())
-          showp l = winterOut l . render . pretty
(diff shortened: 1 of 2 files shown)


hs/src/Reach/Linearize.hs
```diff
@@ -513,7 +514,7 @@ defluid (DKProg at (DLOpts {..}) sps dli dex dvs k) = do
     return $ LLProg at opts' sps dli dex' dvs k'
 
 -- Stich it all together
-linearize :: (forall a. Pretty a => String -> a -> IO ()) -> DLProg -> IO LLProg
+linearize :: (forall a. Pretty a => T.Text -> a -> IO ()) -> DLProg -> IO LLProg
```
tippenein, line 517: this can just use OverloadedStrings and remove `showp'` from `Compile.hs`

hs/src/Reach/Compiler.hs
```diff
@@ -43,41 +43,31 @@ all_connectors =
       ]
 
 compile :: CompilerOpts -> IO ()
-compile copts = do
-  let outn = output copts
-  let outnMay outn_ = case intermediateFiles copts of
-        True -> Just outn_
-        False -> Nothing
+compile CompilerOpts {output, dirDotReach, intermediateFiles, installPkgs, source, tops} = do
```
tippenein, line 46: seemed easier to trace where these variables were coming from if unpacked here

jeapostrophe: I'd rather change it to `(CompilerOpts {..})` if you're going to do that. (Note: I always parenthesize a record splat like this because I think it is confusing that one thing has a space in it

hs/src/Reach/Compiler.hs
```diff
@@ -43,41 +43,31 @@ all_connectors =
...
-      let winterOut = interOut woutn
-      let showp :: (forall a. Pretty a => T.Text -> a -> IO ())
-          showp l = winterOut l . render . pretty
-      let showp' :: (forall a. Pretty a => String -> a -> IO ())
-          showp' = showp . T.pack
+      let showp :: Pretty a => T.Text -> a -> IO ()
+          showp l = interOut woutn l . render . pretty
       dl <- compileDApp which
-      let DLProg _ (DLOpts {..}) _ _ _ _ _ = dl
+      let DLProg _ (DLOpts {dlo_verifyPerConnector, dlo_connectors}) _ _ _ _ _ = dl
```
jeapostrophe, line 64: Gross, no way!

hs/src/Reach/Compiler.hs
```diff
@@ -88,4 +78,8 @@ compile copts = do
       let runConnector c = (,) (conName c) <$> conGen c woutnMay pl
       crs <- HM.fromList <$> mapM runConnector connectors
       backend_js woutn crs pl
-      return ()
```
jeapostrophe, line 91: I need this because sometimes I comment out the js generation and then the function is an error because it ends in a bind

hs/src/Reach/Compiler.hs
```diff
@@ -43,41 +43,31 @@ all_connectors =
...
+compile CompilerOpts {output, dirDotReach, intermediateFiles, installPkgs, source, tops} = do
+  let outnMay = flip doIf intermediateFiles
   let interOut outn_ = case outnMay outn_ of
         Just f -> LTIO.writeFile . f
         Nothing -> \_ _ -> return ()
-  dirDotReach' <- makeAbsolute $ dirDotReach copts
-  let doPkgs = installPkgs copts
-  djp <- gatherDeps_top (source copts) doPkgs dirDotReach'
+  dirDotReach' <- makeAbsolute dirDotReach
+
```
jeapostrophe, line 52: No blank lines in do blocks (you did this in your other commit too and I neglected to complain)

jeapostrophe (changes requested):

hs/src/Reach/Compiler.hs
tippenein: Yeah, I'm way more into `{..}` then manually applying the field accessors :+1: especially since the data structure is defined within this file anyway

jeapostrophe (approved):

PR #252, opened 2021-08-18
Update Solc version and install on Ubuntu

stoneric:
Updated the README to include instructions for installing solc 0.8.7 and PPA install vs. snap

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Read the [documentation](https://docs.reach.sh).
 If you want to work on the Reach compiler, you'll need:
 - stack v2.7.1
 - z3 v4.8.10
-- solidity v0.8.5
+- solidity v0.8.7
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ Read the [documentation](https://docs.reach.sh).
 If you want to work on the Reach compiler, you'll need:
 - stack v2.7.1
 - z3 v4.8.10
-- solidity v0.8.5
+- solidity v0.8.7
```
jeapostrophe, line 15: We should upgrade, but we actually don't require this one. We really do use v0.8.5 right now: https://github.com/reach-sh/reach-lang/blob/master/hs/Dockerfile#L13

If you were to use a different one, it is possible to get bad results.

I'll flag this for updating.

stoneric: ok ill change the pr, the snap install is the issue

PR #255, opened 2021-08-25
ganache exmpl pls review

brohamgoham:
with out the options it is only 3 lines to use ganache, the import, ethers.Web3Provider , and then setting provider for the reach program.


examples/ganache/index.mjs
```diff
@@ -0,0 +1,41 @@
...
+    port: 7545,
+    ws: true,
+    vmErrorsonRPCResponse: true,
+    debug: true,
+    allowUnlimitedContractSize: true
+  };
+  //pass ganacheProvider into setProvider to run a reach program in a ganache env
+  const ganacheProvider = new ethers.providers.Web3Provider(ganache.provider(options));
+  await stdlib.setProvider(ganacheProvider);
+  const accAlice = await stdlib.connectAccount(ganacheProvider.getSigner(999));
```
jeapostrophe, line 24: Can you verify that `newTestAccount` works?

examples/ganache/index.mjs
```diff
@@ -0,0 +1,41 @@
...
+    gasPrice: '21000',
+    gasLimit: '5000000',
+    hardfork: "petersburg",
+    default_balance_ether: 7777777,
+    port: 7545,
+    ws: true,
+    vmErrorsonRPCResponse: true,
+    debug: true,
+    allowUnlimitedContractSize: true
+  };
```
jeapostrophe, line 20: Can you make these options as minimal as possible and as recommended as possible? For example, I think we should suggest only having one account, because I *think* the faucet will be the first one. I also think we don't want people to have to know about the hardfork options. Whatever we put in the example will be copied over and over.

examples/ganache/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@reach-sh/ganache",
+  "type": "module",
+  "dependencies": {
+    "ethers": "^5.0.19",
+    "ganache-core": "2.13.2"
```
jeapostrophe, line 6: I think, but could be wrong, that only this file is needed and we can remove the Makefile, Dockerfile, and docker-compose and they will be picked up by `reach run`

jeapostrophe (changes requested):
Can you add a section to the guide about using Ganache that links to this example?

examples/ganache/index.mjs
brohamgoham: it would throw an error "Having trouble with newTestAccount 0x2930.." when i try to use it as a signer, connectAccount was my work around that error

examples/ganache/index.mjs
brohamgoham: ok. i only used high number of  accounts just to make sure alice connected to the right acct #999 when i printed logs and Alice was connected to ganacheAccount #999

jeapostrophe (commented):
This is great! Thanks! Would you like to try writing a guide section about "Using Ganache with Reach" that shows this example and talks about using ETH live? If you don't want to write that, Hagen or I can, but I want to give you the opportunity if you're interested

brohamgoham:
Id like to do it I’ll get to writing a guide today

PR #269, opened 2021-09-09
CFX fails & timeouts

brohamgoham:
fixing some non deterministic failures, i believe  there is just one issue causing these timeouts
cache-events no longer times out on CFX but does on ETH

### DIFF
--- a/examples/raffle/index.rsh
+++ b/examples/raffle/index.rsh
@@ -95,7 +95,7 @@ export const main =
             require(randomMatches(player, ticket));
             ticketsM[player] = howManyReturned;
             delete randomsM[player];
-            return [ (hwinner + (ticket % howMany)) % howMany,
+            return [ hwinner +  howMany,
                      howManyReturned + 1 ];
           })
         )
@@ -122,8 +122,9 @@ export const main =
         const sponsort = declassify(_sponsort); });
       Sponsor.publish(sponsort);
       require(sponsortc == digest(sponsort));
-
-      const winningNo = (hwinner + (sponsort % howManyReturned)) % howManyReturned;
+      // CFX does like doing division with UInts );
+      // shared_impl.ts ln 356, Players bigNumber division is throwing it off
+      const winningNo = hwinner + sponsort;
       commit();
 
       each([Sponsor, Player], () => {
--- a/examples/raffle/index.rsh
+++ b/examples/raffle/index.rsh
@@ -153,3 +153,4 @@ export const main =
       .timeout(deadline, () => closeTo(Sponsor, () => {}));
       commit();
     });
+    
--- a/examples/raffle/index.rsh
+++ b/examples/raffle/index.rsh
@@ -95,7 +95,7 @@ export const main =
             require(randomMatches(player, ticket));
             ticketsM[player] = howManyReturned;
             delete randomsM[player];
-            return [ hwinner +  howMany,
+            return [ (hwinner + (ticket % howMany)) % howMany,
                      howManyReturned + 1 ];
           })
         )
@@ -122,9 +122,8 @@ export const main =
         const sponsort = declassify(_sponsort); });
       Sponsor.publish(sponsort);
       require(sponsortc == digest(sponsort));
-      // CFX does like doing division with UInts );
-      // shared_impl.ts ln 356, Players bigNumber division is throwing it off
-      const winningNo = hwinner + sponsort;
+
+      const winningNo = (hwinner + (sponsort % howManyReturned)) % howManyReturned;
       commit();
 
       each([Sponsor, Player], () => {
--- a/js/stdlib/ts/cfxers.ts
+++ b/js/stdlib/ts/cfxers.ts
@@ -79,7 +79,7 @@ interface IContract {
 export class Contract implements IContract {
   [k: string]: any
   _abi: any[]
-  _wallet: IWallet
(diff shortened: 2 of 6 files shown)


js/stdlib/ts/cfxers.ts
```diff
@@ -181,16 +177,13 @@ export class Contract implements IContract {
...
-            txn.gas = est.gasLimit;
+          if ( txn.gasLimit === undefined ) {
+            txn.gasLimit = est.gasLimit;
           }
           if ( txn.storageLimit === undefined ) {
             txn.storageLimit = est.storageCollateralized;
           }
         }
-        if ( txn.storageLimit === undefined || txn.storageLimit == 0 ) {
-          txn.storageLimit = 2048;
```
jeapostrophe, line 193: In my experience, the calculation of the storage limit estimate does not always put something in when it is needed and if it isn't needed, will get refunded, so we need this to use linear state.

js/stdlib/ts/cfxers.ts
```diff
@@ -181,16 +177,13 @@ export class Contract implements IContract {
         }
         debug(`cfxers:handler`, fname, {est, est_err});
         if ( est ) {
-          if ( txn.gas === undefined ) {
-            txn.gas = est.gasLimit;
+          if ( txn.gasLimit === undefined ) {
+            txn.gasLimit = est.gasLimit;
```
jeapostrophe, line 181: This --- https://confluxnetwork.gitbook.io/js-conflux-sdk/api/transaction#Transaction.js/Transaction/**constructor** --- says the field is named `gas`, not `gasLimit`

brohamgoham:
raffle was running fine on cfx locally but still failing on CI
And mint basic only timed out because ran out of gas

jeapostrophe:
> raffle was running fine on cfx locally but still failing on CI

In my experience, this can mean that there's a race condition where the CI machine running slower than our fast developer laptops behaves differently.

> And mint basic only timed out because ran out of gas

Does it work if the participants have more CFX in their accounts? If the gasLimit is increased? Another thing that can happen is that, at least on ETH, "out of gas" is kind of a catch all that just means "the transaction failed" which could mean a bigger problem (like trying to compete in a finished race)

examples/raffle/index.rsh
```diff
@@ -122,8 +122,9 @@ export const main =
         const sponsort = declassify(_sponsort); });
       Sponsor.publish(sponsort);
       require(sponsortc == digest(sponsort));
-
-      const winningNo = (hwinner + (sponsort % howManyReturned)) % howManyReturned;
+      // CFX does like doing division with UInts );
+      // shared_impl.ts ln 356, Players bigNumber division is throwing it off
+      const winningNo = hwinner + sponsort;
```
jeapostrophe, line 127: The change that you made is not correct. What you've changed it to will overflow

brohamgoham: The assertion of the original mod division remainder (sponsors % howManyReturned) % howmanyReturn 
 fails and points to a BigNumbers mod error, I kept thinking it was a issue w gas and storage but its how CFX handles bigNumber values im pretty sure

jeapostrophe: That is very valuable to learn. We can't change the program though. Potentially we need to update how the Reach compiler outputs OR we need to get a problem report to Conflux. IDEA: Could you write a little Solidity program by hand that has the same problem when run on Conflux, but not on Eth?

examples/raffle/index.rsh
```diff
@@ -152,4 +152,4 @@ export const main =
         }))
       .timeout(deadline, () => closeTo(Sponsor, () => {}));
       commit();
-    });
+    });
```
DanBurton, line 155: Nitpick: can you set your editor to keep a newline at the end of the file?

brohamgoham: i got you

js/stdlib/ts/cfxers.ts
```diff
@@ -423,13 +422,31 @@ export class Wallet implements IWallet {
...
+    let newStorageCost = format.big(estimate.storageCollateralized).times(storageRatio).toFixed(0);
+    
+    // Note: balace needs to cover value +  (gas + storageLimit)
+    let gasXstorage = format.big(newGasCost).plus(newStorageCost).toFixed(0);
+    let finalCost = format.big(txn.value).plus(gasXstorage).toFixed(0);
+    const final = BigInt(finalCost);
+    
+    debug(`SendTxn attempt, Final Cost of Tx is , ${final},  Balance of sender ${from} is ${balance}`);
+    if ( final > balance ) {
+      debug(`Checking: Account balanace of  ${from} is ${balance} and gasFee is, ${newGasCost}: Total TxValue is ${final}`)
```
jeapostrophe, line 441: Typo

js/stdlib/ts/cfxers.ts
```diff
@@ -423,13 +422,31 @@ export class Wallet implements IWallet {
...
+      throw Error(` INSUFFICIENT FUNDS GAS COST IS ${newGasCost},  TXN VALUE IS  ${final}, ACCOUNT ${from} ONLY HAS A BALANCE OF ${balance}`);
```
jeapostrophe, line 442: Still no test for this

jeapostrophe (changes requested):

PR #280, opened 2021-09-15
Add `muldiv` primitive

chrisnevers:
Add `muldiv: UInt -> UInt -> UInt` which will multiply the first two arguments and divide the result by the third argument. If a connector supports wide arithmetic operations, it will use them. The resulting quotient must be less than or equal to `UInt.max`.

### DIFF
--- a/examples/muldiv/index.mjs
+++ b/examples/muldiv/index.mjs
@@ -1,0 +1,27 @@
+import {loadStdlib} from '@reach-sh/stdlib';
+import * as backend from './build/index.main.mjs';
+const stdlib = loadStdlib(process.env);
+
+(async () => {
+  const startingBalance = stdlib.parseCurrency(100);
+
+  const [ accAlice ] =
+    await stdlib.newTestAccounts(1, startingBalance);
+  console.log('Hello, Alice!');
+
+  console.log('Launching...');
+  const ctcAlice = accAlice.deploy(backend);
+
+  console.log('Starting backends...');
+  await Promise.all([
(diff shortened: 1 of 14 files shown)


examples/muldiv/index.mjs
```diff
@@ -0,0 +1,27 @@
...
+  const ctcAlice = accAlice.deploy(backend);
+
+  console.log('Starting backends...');
+  await Promise.all([
+    backend.Alice(ctcAlice, {
+      ...stdlib.hasRandom,
+      log: (msg, x) => console.log(msg, x.toString()),
+      // This will be multiplied by 4 then divided by 5.
+      // x * 4 would normally overflow on Algorand
+      x: 9223372036854775807,
```
jeapostrophe, line 22: I think this should be a bignumberify because I'm scared of floating point

jeapostrophe: This example should also check the result by using big numbers

examples/muldiv/index.rsh
```diff
@@ -0,0 +1,25 @@
...
+  A.publish(x, z);
+
+  const y = muldiv(x, 4, 5);
+
+  require(y == z);
+
+  commit();
+
+  const a = muldiv(x, 4, 5);
+  A.interact.log("a", a);
```
jeapostrophe, line 23: This is almost a test but it is not checked

jeapostrophe (changes requested):
Changelog too. Great work!

jeapostrophe (approved):

PR #281, opened 2021-09-16
CORE-25 WIP: introduce config RC and offer to create it when missing

mattaudesse:


jeapostrophe:
There's a lot here. Can you help me understand the idea and why there's so much complexity? I don't understand every detail, but I see that there's something interesting going on with the config file contents... Why not just mount "~/.reach" as a volume and create it if it doesn't exist?

mattaudesse:
> There's a lot here. Can you help me understand the idea and why there's so much complexity? I don't understand every detail, but I see that there's something interesting going on with the config file contents... Why not just mount "~/.reach" as a volume and create it if it doesn't exist?

Sure, here's a jumble of my thought process:

Now that we've got `optparse-applicative` and can write stuff in Haskell, I think it's generally more natural and self-documenting (i.e. better UX) to add flags/arguments/etc than environment variables as user-tweakables.

```
Missing: COMMAND

reach 0.1.4 - Reach command-line tool

Usage: <interactive> [-c|--config CONFIG_YAML] COMMAND

Available options:
  -c,--config CONFIG_YAML  Override configuration file path (XDG-based default:
                           /home/matt/.config/reach/config.yaml)
  -h,--help                Show this help text

Available commands:
  compile                  Compile an app
  clean                    Delete 'build/$MODULE.$IDENT.mjs'
  init                     Set up source files for a simple app in the current
                           directory
  run                      Run a simple app
  down                     Halt all Dockerized Reach services and devnets
  scaffold                 Set up Docker scaffolding for a simple app in the
                           current directory
  react                    Run a simple React app
  rpc-server               Run a simple Reach RPC server
  rpc-run                  Run an RPC server + frontend with development
                           configuration
  devnet                   Run only the devnet
  upgrade                  Upgrade Reach
  update                   Update Reach Docker images
  docker-reset             Kill and remove all Docker containers
  version                  Display version
  hashes                   Display git hashes used to build each Docker image
  help                     Show usage
  ```

I can easily see cases where users want to set defaults that pertain to a given project rather than _all_ projects, so I thought it'd make sense to allow overriding the config file's location (that's what `-c|--config CONFIG_YAML` does). This made me wonder if we should consider the config-stacking behavior some tools use where project-specific config overrides more general settings stored "higher up", but without thinking about it too deeply I kinda think that'd be a mistake in our case. Obviously a future decision anyway.

I think you might say "they can override the connector with `REACH_CONNECTOR_MODE`", which of course is true, but what happens when we want to add new configuration knobs?

Maybe the answer to that is "just use a `~/.config/reach/defaultNetwork` file that only contains `ETH` and be done with it", but I think that delays answering some of these other questions and probably actually entails more effort over time than just making a Generic Thing first.

`REACH_CONFIG_SKIP` is provided as an escape-hatch for core devs/CI/other people who don't want to bother with config and don't want to see the first-run prompts Chris asked for every time they run `reach`.

Using the [XDG scheme](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) rather than dumping config directly into a user's `$HOME` directory is more flexible + arguably more polite and tidy.

Since `$HOME` doesn't fall within our `Docker` "context" we need to do something less direct to feed it into the `reach-cli` container.

Volume mounts have created a lot of problems and confusion for us and I think we're still not all on the same page RE: why/how (I just happen to see it more often because I'm working directly on these parts of the project). My goal was to avoid exacerbating with a new one unless you told me to proceed anyway. Off the top of my head I also don't think they're strictly necessary for anything we're doing - they just happen to be how we've built things before.

Remember this trick?
https://github.com/reach-sh/reach-lang/blob/master/rpc-client/py/Makefile#L12-L18
```Makefile
.PHONY: build
build: clean ./VERSION
	docker build --tag=reachsh/rpc-client-py:latest .
	@echo 'Bundling...' \
	  && CID="$$(docker run -itd --rm reachsh/rpc-client-py:latest)" \
	  && docker cp "$$CID:/lib/dist" ./dist \
	   ; docker container stop "$$CID"
```

Copying out of "build containers" is an alternative solution (not directly relevant to this discussion - just as an aside) that doesn't force the same tradeoffs (WRT running solely on one's local machine, or using a `linux` executor on CI, etc) we currently have to accept. I wouldn't be surprised if someday down the road we got frustrated enough with mounts to replace them, so here again I didn't want to complicate our future selves' lives unnecessarily.

The extra bits of `sh` that `base64` encode the dynamically-located config file and push it up as an argument aren't very elegant but give us some flexibility to build upon and don't bear the same cost as using a volume mount.

Another thing I wanted to avoid was bouncing back and forth between the `reach` wrapper script and `reach-cli` container a bunch of times, which is why you see a bit of extra argument parsing in `reach` and a little duplication in `Main.hs`. This is gross but worth it IMHO. It's also not something I think we'll need to do much more of provided we retain the config flexibility I'm proposing.

`base64 "$CF" | tr -d '\n'` is just a less buggy means to pass the config without newlines/escaping/etc.

https://github.com/reach-sh/reach-lang/pull/281/files#diff-bb894cf4646995c74b769a43368d84339d5611ccfe521b6265162ec14a2d8532R23-R26
```haskell
-- TODO friendlier layout for human-editing
data RC = RC
  { rcDefaultConnector :: Maybe ConnectorName
  } deriving (Eq, Show, Generic, ToJSON, FromJSON)
```

The friendliness I'm talking about here is swapping
```yaml
rcDefaultConnector: ETH
uglyFieldAccessor: "that won't make sense to end-users"
```
for something like
```yaml
connector:
  preferred: ETH
  new:
    fields:
      - as
      - they
      - become
      - necessary
```
that's easier for humans to grok and edit naturally.

I think the only other thing to mention is I've consolidated some stuff like the `Connector.hs` changes to be more DRY.

PR #286, opened 2021-09-22
Rotate dbg script

vinmazzi:


scripts/devnet-algo/rotate_dbg.sh
```diff
@@ -0,0 +1,10 @@
+cd /dbg
+for i in $(ls | grep [[:digit:]]-  | awk -F- '{print $1}' | uniq); do 
+  in_use=$(lsof | grep $i-[[:digit:]] ; echo $?)
```
jeapostrophe, line 3: I am scared of the race condition when the daemon has just closed `6-20` and is about to open `6-21`. We know it will never modify anything but the last one, so I think it is simpler to just only care about that

scripts/devnet-algo/rotate_dbg.sh
```diff
@@ -0,0 +1,10 @@
+cd /dbg
+for i in $(ls | grep [[:digit:]]-  | awk -F- '{print $1}' | uniq); do
```
jeapostrophe, line 2: I believe that `uniq` must occur after `sort`, because `ls` is not guaranteed to return sorted

scripts/devnet-algo/rotate_dbg.sh
```diff
@@ -0,0 +1,10 @@
+cd /dbg
+for i in $(ls | grep [[:digit:]]-  | awk -F- '{print $1}' | uniq); do 
+  in_use=$(lsof | grep $i-[[:digit:]] ; echo $?)
+  if [ $in_use -ne 0 ]; then
+    mkdir $i;
+    mv $i-* $i;
+    tar -czf $i.tgz $i;
+    rm -rf $i;
```
jeapostrophe, line 8: My guess is shellcheck complains about this code

scripts/devnet-algo/rotate_dbg.sh
vinmazzi: On this one I'll get a return like this from: ls | grep [[:digit:]]-  | awk -F- '{print $1}' | uniq:
15
16
17
18
It will basically iterate along the existing sections, it won't look file by file.
So if there's any 15- file in use, it skips it and moves to the next section so all 15-x.json files won't be gziped until 15 workflow is actually finished.

scripts/devnet-algo/rotate_dbg.sh
vinmazzi: The order here doesn't really matter actually, here I'm only concerned about the section groups (15-,16-,17-...).
I could still sort them if you'd like.

PR #288, opened 2021-09-22
examples/no-funds-example

DanBurton:
Was trying to write a "smoke test" for #269, ended up with this monstrosity, because promises and exceptions are kinda messy.

We should consider using a promise library that lets them be "cancelable", so that we can programmatically kill promise threads (from within js) that are hanging, or just taking longer than we want.

### DIFF
--- a/examples/no-funds-example/index.mjs
+++ b/examples/no-funds-example/index.mjs
@@ -1,0 +1,113 @@
+import {loadStdlib} from '@reach-sh/stdlib';
+import * as backend from './build/index.main.mjs';
+const stdlib = loadStdlib(process.env);
+
+const theLog = [];
+function log(m) {
+  theLog.push(m);
+  console.log(m);
+}
+
+function doTheThing(accDeployer, accAttacher) {
+  log('Hello, Deployer and Attacher!');
+
+  log('Launching...');
+  const ctcDeployer = accDeployer.deploy(backend);
+  const ctcAttacher = accAttacher.attach(backend, ctcDeployer.getInfo());
+
+  log('Starting backends...');
+
+  const pDeployer = backend.Deployer(ctcDeployer, {deadline: 10});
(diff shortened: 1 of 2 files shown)


DanBurton:
Hangs (?) on ALGO, as currently written. Works as expected on ETH and CFX when ran against code from #269

jeapostrophe:
I must be stupid. 

What is wrong with:
```
'reach 0.1';

  const A = Participant('A', {});
  deploy();
  A.publish();
  commit();
});
```
and
```
import {loadStdlib} from '@reach-sh/stdlib';
import * as backend from './build/index.main.mjs';
const stdlib = loadStdlib(process.env);

(async () => {
  const accAlice = await stdlib.newTestAccount(0);
  const ctcAlice = accAlice.deploy(backend);
  let err;
  try {
   await backend.Alice(ctcAlice, {});
  } catch (e) {
   err = e;
  }
  console.log(`Error was`, err);
  if ( ! err ) {
   throw new Error(`Expected error, but didn't get one`);
  }  
})();
```

Yea, maybe we could check that the `JSON.stringify(err).contains("balance")` or something, but... who cares for now?

DanBurton:
One of the things wrong with it is that `accAlice.deploy` spawns a promise that will fail in the background, and then it will cause the node program to hang or fail because nobody caught that particular promise. I'm pretty sure there's a simpler way to do things along those lines, but there's something tricky about making sure all the rejected promises get caught that is eluding me.

jeapostrophe:
I assumed that the exception from the background failed will be exposed when something `await`s on it and Alice will do that before going to the `recv` on the first message

DanBurton:
Interestingly, on CI, ALGO failed not due to an example timeout, but because:

> 'attacher0' didn't throw, but it should have.

I'm not sure how this is possible. I'll create a Jira issue to look into it.

PR #291, opened 2021-09-23
algo Wallet connect

brohamgoham:
wallet connect is working with stand alone reach.wz() function i made, but regarding walletFallback i still need to make some changes


brohamgoham:
might need to make a whole new doWalletFallback and also change getting providerEnv to bridge URL and QRcode import

jeapostrophe:
This needs to be rebased. There are not really 77 changed files. Also, anything that is not a new walletFallback option is not useful. What you have right now is a good experiment for you to learn about how WalletConnect works, but it is not applicable.

js/stdlib/ts/ALGO_WC.ts
```diff
@@ -0,0 +1,275 @@
+import WalletConnect from "@walletconnect/client";
+import algosdk from 'algosdk';
+import {  IQRCodeModal } from "@walletconnect/types";
+import { IClientMeta } from "@walletconnect/types";
+import { AlgorandTxn, ConnectionSettings, EncodedTransaction, SignedTx } from "@randlabs/myalgo-connect";
```
jeapostrophe, line 5: Why are you using stuff from myalgo? This code looks copied from https://github.com/randlabs/myalgo-connect/blob/d2d96ad22062994789c8604b44dd542b26cd89b2/lib/main.js rather than being based on Algorand's example --- https://developer.algorand.org/docs/features/walletconnect/#quick-start-for-dapp-devs

Your code is just so much more complicated and Algorand's example... why is it so much more stuff?

jeapostrophe: In particular... why does something like this not work? https://github.com/reach-sh/reach-lang/blob/23f379d207b2938031c64ede14c63fe709598643/js/stdlib/ts/ALGO_WalletConnect.ts

FWIW, my version does not actually work in my tests, but I don't understand why.

brohamgoham: I just cleaned up a lot of the imports but ALGO WC example wasn’t working when I try to use it with setWalletFallback

brohamgoham:
k

PR #293, opened 2021-09-24
CORE-25 WIP

mattaudesse:


hs/app/reach/Main.hs
```diff
@@ -241,9 +250,11 @@ mkVar = do
   reachEx <- lookupEnv "REACH_EX"
     >>= maybe (die "Unset `REACH_EX` environment variable") packed
   connectorMode <- do
-    rcm <- m "REACH_CONNECTOR_MODE" (pure "ETH-devnet") (pure . id)
+    let p = "REACH_CONNECTOR_MODE"
+    let d = "ETH-devnet"
```
jeapostrophe, line 254: Since we last talked I've convinced myself that no-default is a good idea. I don't think you need to do it in this PR, because I like incrementalism, but I think you should make a card and do it in this spring.

hs/app/reach/Main.hs
```diff
@@ -643,6 +654,37 @@ switchQuiet = switch
   $ long "quiet"
  <> help "Withhold progress messages"
 
+mkEnv :: IORef Effect -> Maybe FilePath -> Maybe Var -> IO (Parser Env)
+mkEnv eff mf mv = do
+  let strConfC = strOption (long "dir-config-container" <> value "/app/config" <> internal)
```
jeapostrophe, line 659: Do you use this or you're just being cool keeping track of everything?

hs/app/reach/Main.hs
```diff
@@ -223,11 +225,18 @@ truthyEnv = \case
...
   let packed = pure . pack
   let q e n = lookupEnv e >>= maybe n packed
   let m e n f = lookupEnv e >>= maybe n (\case "" -> n; j -> f j)
+  let def' p d = ifM (not <$> doesFileExist cnf) (pure d)
+        $ (filter (p' `L.isPrefixOf`) . lines <$> readFile cnf) >>= \case
+            [] -> pure d
+            cs -> pure $ case L.stripPrefix p' (last cs) of
+              Just c | length c > 0 -> c
+              _ -> d
+        where p' = p <> "="
```
jeapostrophe, line 239: This is a hardcore little parser.

mattaudesse: Yeah... it's kinda inflexible too, and I'm not sure if that's a good thing or not.

I had it in the back of my head that we may want to make it smarter later but this works in the minimal case.

jeapostrophe: As you know I like minimal things :) What you have makes the file "look like" something that could be `source`'d. That's cool. I think it is also feasible that some library we already use (yaml?) can serialize an arbitrary map and then we just load it. I bet (?) that would be less code and because it's someone else's problem, the parser is probably robustier

PR #295, opened 2021-09-27
CI: Test External Reach Repositories

chrisnevers:
Add a script that will create an `example` test case for external Reach repositories.

### Create New Example

```
./mk-ext-test.sh <name> <git-url>
```

This will produce a directory in `examples` and a `Makefile`.

### Test Example

When `make build` is ran, the repository will be cloned to `examples/ext-<name>`. We will only test if the repo can be compiled (via `one.sh`). 

The repository is cloned to `examples/ext-<name>` because `one.sh` expects the test case to be directly under `examples`. The repo cannot be cloned directly into `examples/<name>` because it may **also** specify a `Makefile`.


examples/one.sh
```diff
@@ -17,7 +17,7 @@ echo
 (
   cd "$e" || exit 1
   if [ -f Makefile ] && has_target ; then
-    make "$MODE"
+    make REACH=../../reach "$MODE"
```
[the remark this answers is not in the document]
jeapostrophe: Good catch

examples/duoswap-core/Makefile
```diff
@@ -0,0 +1,16 @@
...
+checkout:
+	if test -d ../ext-duoswap-core; then echo "Repo already cloned"; else git clone https://github.com/reach-sh/duoswap-core ../ext-duoswap-core; fi
+
+.PHONY: build
+build: checkout
+	cd ../ext-duoswap-core && ../one.sh build
+
+.PHONY: run
+run:
+	echo "external code will not be ran"
```
jeapostrophe, line 16: What you have is valuable... we're ensuring that we don't break their code's compilation, but it is a little unsatisfying that we don't run it either. You're scared it will take too long? You're scared that they will blow us up? Can't think of a good "protocol" to tell it to run?

chrisnevers: I was thinking it could either be long, interactive, or have no capability to run (i.e. just a reach library). But, I could have the script generate run commands. Then, it's up to the developer adding the test to check whether it makes sense and adjust if necessary?

jeapostrophe: Yea, those all make sense. In the case of duoswap, do you think it is worth running?

chrisnevers: Since it already runs in its own CI, I think having `reach-lang` check it can compile without long verification is fine. No strong feelings though

jeapostrophe (approved):

PR #297, opened 2021-09-28
CORE-620: Ensure `Participant` Can Make First Publication

chrisnevers:
Have the compiler ensure that the first publication of a Reach program is made by a unique `Participant` and update documentation.

### DIFF
--- a/hs/src/Reach/Eval/Core.hs
+++ b/hs/src/Reach/Eval/Core.hs
@@ -3853,8 +3853,10 @@ doToConsensus ks (ToConsensusRec {..}) = locAt slptc_at $ do
   ensure_live "to consensus"
   let st_pure = st {st_mode = SLM_ConsensusPure}
   let pdvs = st_pdvs st
-  allClasses <- and <$> traverse is_class (S.toList whos)
-  when (not (st_after_first st) && allClasses) $ do
+  isSoloSend <- case S.toList whos of
+                  [solo] -> not <$> is_class solo
+                  _ -> return False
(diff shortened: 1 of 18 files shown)


hs/src/Reach/Eval/Core.hs
```diff
@@ -3853,6 +3853,9 @@ doToConsensus ks (ToConsensusRec {..}) = locAt slptc_at $ do
   ensure_live "to consensus"
   let st_pure = st {st_mode = SLM_ConsensusPure}
   let pdvs = st_pdvs st
+  allClasses <- and <$> traverse is_class (S.toList whos)
+  when (not (st_after_first st) && allClasses) $ do
```
jeapostrophe, line 3857: There should also be exactly one

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
chrisnevers: So, `Anybody.publish()` and `race(Alice, Bob).publish()` are no longer legal first publications?

jeapostrophe: Correct, because that means that both `Alice` and `Bob` would both try to `deploy` and so they'd both create a new contract and not be able to share the contract info with each other

jeapostrophe: (It is literally possible to allow it, but I don't think the typical frontend would be correct with this backend.)

jeapostrophe (approved):

PR #298, opened 2021-09-28
CORE-615: remove `REACH_CONNECTOR_MODE` defaulting

mattaudesse:
**(WIP)**

@jeapostrophe I wanted to ask whether you agree with the following behavior before moving on to tweaking the compiler itself or making any docs changes about `REACH_CONNECTOR_MODE` becoming mandatory:

```bash
$ unset REACH_CONNECTOR_MODE
$ cd examples/nim

### The following needn't rely on `REACH_CONNECTOR_MODE`: ##########################
# (and I've tested each of them, BTW)
$ ../../reach help
$ ../../reach -h
$ ../../reach version
$ ../../reach version --help
$ ../../reach hashes
$ ../../reach down
$ ../../reach clean
$ ../../reach init
$ ../../reach scaffold
$ ../../reach upgrade
$ ../../reach update
$ ../../reach docker-reset

### ...But these will protest if it's absent or malformed: ########################
# compile | run | react | rpc-server | rpc-run | devnet

$ ../../reach compile
Unset `REACH_CONNECTOR_MODE` environment variable

$ REACH_CONNECTOR_MODE=X ../../reach compile
Invalid `REACH_CONNECTOR_MODE`: X

$ REACH_CONNECTOR_MODE=ETH ../../reach compile
  Verifying when ONLY "A" is honest
  Verifying when ONLY "B" is honest
Checked 86 theorems; No failures!

$ export REACH_CONNECTOR_MODE=X
$ ../../reach compile
Invalid `REACH_CONNECTOR_MODE`: X

$ export REACH_CONNECTOR_MODE=ALGO
$ ../../reach compile
  Verifying when ONLY "A" is honest
  Verifying when ONLY "B" is honest
Checked 86 theorems; No failures!
```


mattaudesse:
https://github.com/reach-sh/reach-lang/pull/293#discussion_r715644949

jeapostrophe:
I did not anticipate `compile` caring, because it doesn't read it all, but I get why you'd do that.

I think "Unset `REACH_CONNECTOR_MODE` environment variable" is a bad error message. I understand it, but I think it might be confusing, am I _supposed_ to unset it? I think it should say what it should do: "Set RCM to one of these options: ...."

--

When it comes to downloading images... what do you think about updating images that are present, but only downloading for the first time images needed by the RCM? This was an ALGOnut will never get the CFX devnet?

mattaudesse:
Thanks - I like the idea of improving the "unset" error message too, and I definitely think it seems kinda odd that `compile` would care about `REACH_CONNECTOR_MODE`... Maybe we should consider this to be an opportunity to relocate the [reporting code](https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Report.hs#L18)?

I've introduced the RCM to `compile` because CORE-25 asks that it be logged, but maybe the compiler shouldn't be the one handling that now(?) Relocating could also come later, I suppose.

RE: images I feel like selective-syncing is the feature people most care about from CORE-25, so if it makes them happier we should totally do it.

jeapostrophe:
> Thanks - I like the idea of improving the "unset" error message too, 

Good

> and I definitely think it seems kinda odd that `compile` would care about `REACH_CONNECTOR_MODE`... Maybe we should consider this to be an opportunity to relocate the [reporting code](https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Report.hs#L18)?

No, let's stay focused. The compile can report what RCM is, even if that is "unset". And of course, we expect 99% of users will be putting RCM in their .profile so it will actually see it.

> I've introduced the RCM to `compile` because CORE-25 asks that it be logged, but maybe the compiler shouldn't be the one handling that now(?) Relocating could also come later, I suppose.

We could ALSO report on other things than compile, you can make a Jira card for that?

> RE: images I feel like selective-syncing is the feature people most care about from CORE-25, so if it makes them happier we should totally do it.

mattaudesse:
> > Thanks - I like the idea of improving the "unset" error message too,
> 
> Good

How about this?
```bash
$ ../../reach run
Missing `REACH_CONNECTOR_MODE` environment variable - must be one of:
 * ALGO-devnet
 * ALGO-live
 * ALGO-browser
 * CFX-devnet
 * CFX-live
 * CFX-browser
 * ETH-devnet
 * ETH-live
 * ETH-browser
https://docs.reach.sh/ref-usage.html#%28env._.R.E.A.C.H_.C.O.N.N.E.C.T.O.R_.M.O.D.E%29
```

---
> No, let's stay focused. The compile can report what RCM is, even if that is "unset". And of course, we expect 99% of users will be putting RCM in their .profile so it will actually see it.

Cool - I've swapped `scriptWithConnectorMode` back to just `script` for `reach compile` now.

---
> We could ALSO report on other things than compile, you can make a Jira card for that?

Sure: CORE-621.

jeapostrophe:
Great, is this ready to go? I think that maybe we should delay merging (or at least releasing) until the config feature is also available, because it might be too painful. We should also audit the docs to make sure they don't imply there's a default. Oh, and the "you must set RCM", should really also refer to the config setting command.

mattaudesse:
> Great, is this ready to go? I think that maybe we should delay merging (or at least releasing) until the config feature is also available, because it might be too painful. We should also audit the docs to make sure they don't imply there's a default. Oh, and the "you must set RCM", should really also refer to the config setting command.

Yeah, I think this is incomplete and too unfriendly to users until we update the docs and guide their config. Delaying sounds right to me. It also doesn't actually log the RCM yet (although that seems pretty straightforward to add).

I could make this the "CORE-25 + CORE-615 PR" since they're somewhat reliant on each other, and just continue with both here unless you'd prefer to keep them separate - do you have a preference?

jeapostrophe:
That sounds right to me @mattaudesse ; I think you can safely move CORE-615 into In-PR and then just at CORE-25 to this PR.

hs/app/reach/Main.hs
```diff
@@ -319,18 +345,27 @@ swap a b src = T.replace ("${" <> a <> "}") b src
...
-  [ "reach"
-  , "reach-cli"
-  , "react-runner"
-  , "rpc-server"
-  , "runner"
-  , "devnet-algo"
-  , "devnet-cfx"
-  , "devnet-eth"
+-- Left: 3rd party; Right: Reach
+type Image = Either Text Text
```
jeapostrophe, line 349: Why does it matter? Just to know the `reachsh` at the front?

jeapostrophe (commented):
I believe where we're at is just waiting for the `config` command?

hs/app/reach/Main.hs
mattaudesse: That and tagging in `reach update` is handled differently, yup.

What I mean RE: tagging is that `reach update` uses `docker pull reachsh/${img}:${latest|x|x.y|x.y.z}` in the case of our own `reachsh` images, whereas we don't need to do that in the case of e.g. `postgres:11-alpine`.

mattaudesse:
> I believe where we're at is just waiting for the `config` command?

That's the main thing, yup. I'm working on it now.

I've got some other UX ideas/open questions floating around in my head but I don't think they need to be addressed right away.

PR #300, opened 2021-09-29
CORE-307: Warn When Transferring New Token On Algorand

chrisnevers:
Add an ALGO warning when transferring a token in the same consensus step it was created in.


hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1300,6 +1311,16 @@ checkTxnInit vTypeEnum ct_mcsend = do
   -- [ txn ]
   return ()
 
+checkTxnUsage :: CheckTxn -> App ()
+checkTxnUsage (CheckTxn {..}) = do
+  case ct_mtok of
+    Just tok -> do
+      newToks <- (liftIO . readIORef) =<< asks eNewToks
+      when (tok `S.member` newToks) $ do
+            bad $ LT.pack $ "Token cannot be transferred within the same consensus step it was created in (at: " <> show ct_at <> ")"
```
jeapostrophe, line 1320: I know it is harder, but I think this needs to have an error code with a link to a longer explanation of how to fix it and why it is a problem. I don't know how reusable that stuff is, but sometimes we have to make our lives painful to help the users.

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1588,6 +1615,7 @@ cloop which (C_Loop at svs vars body) = recordWhich which $ do
 ch :: LT.Text -> Int -> CHandler -> App ()
 ch _ _ (C_Loop {}) = return ()
 ch afterLab which (C_Handler at int from prev svs msg_ timev secsv body) = recordWhich which $ do
+  resetNewToks
```
jeapostrophe, line 1618: I am anxious about this style of one global box that gets reset at the start of a handler. I'd prefer if this was part of `recordWhich` (which uses local state for the which and duplicates the IORef of the txn count). I think it should fit there, maybe `dupeNewToks`?

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1300,6 +1311,16 @@ checkTxnInit vTypeEnum ct_mcsend = do
...
+  case ct_mtok of
+    Just tok -> do
+      newToks <- (liftIO . readIORef) =<< asks eNewToks
+      when (tok `S.member` newToks) $ do
+            bad $ LT.pack $ "Token cannot be transferred within the same consensus step it was created in (at: " <> show ct_at <> ")"
+    Nothing -> return ()
+
+
 checkTxn :: CheckTxn -> App ()
 checkTxn (CheckTxn {..}) = when (ct_always || not (staticZero ct_amt) ) $ do
```
jeapostrophe, line 1325: ct@

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1310,6 +1331,7 @@ checkTxn (CheckTxn {..}) = when (ct_always || not (staticZero ct_amt) ) $ do
           Just tok ->
             ("axfer", "AssetReceiver", "AssetAmount", "AssetCloseTo", textra)
             where textra = ca tok >> check1 "XferAsset"
+  checkTxnUsage $ CheckTxn {..}
```
jeapostrophe, line 1334: ct

hs/src/Reach/Connector/ALGO.hs
jeapostrophe: If you don't agree with that, at the least it should say why: ", because on Algorand, you must opt-in to a token to receive it, but it is impossible to opt-in to a token unless you know its id, and the id cannot be known until after this transaction"

jeapostrophe (changes requested):

jeapostrophe:
I made some small comments, sorry :(

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1300,8 +1329,18 @@ checkTxnInit vTypeEnum ct_mcsend = do
   -- [ txn ]
   return ()
 
+checkTxnUsage :: CheckTxn -> App ()
+checkTxnUsage (CheckTxn {..}) = do
+  case ct_mtok of
+    Just tok -> do
+      newToks <- (liftIO . readIORef) =<< asks eNewToks
+      when (tok `S.member` newToks) $ do
+            bad $ LT.pack $ runIdentity . flip runReaderT 3 $ getErrorMessage [] ct_at True Err_TransferNewToken
```
jeapostrophe, line 1338: That's a mouthful :)

jeapostrophe: I think you can just do `nest` and `show` or something?

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -34,9 +34,27 @@ import Reach.Util
...
+-- Errors for ALGO
+
+data AlgoError
+  = Err_TransferNewToken
+  deriving (Eq, ErrorMessageForJson, ErrorSuggestions, Generic)
+
+instance HasErrorCode AlgoError where
+  errPrefix = const "RA"
+  errIndex = \case
+    Err_TransferNewToken {} -> 0
```
jeapostrophe, line 51: The testing apparatus should ensure that these errors are covered in the docs too.

jeapostrophe (changes requested):
After these little things, ship it!

hs/src/Reach/Connector/ALGO.hs
chrisnevers: I believe the indentation will be incorrect if using the `nest`. LMK if you think it's still worth converting to though

jeapostrophe: Honestly, I'd rather have it look bad then expose these details of the Texty monad. `nest` twice would give 4 lines right? That's not so bad?

jeapostrophe:
Ship it!

PR #302, opened 2021-10-04
CORE-484: Disallow Mismatched Type `==` Comparisons

chrisnevers:
Do not allow comparing values for equality if they are not of the same type. If the values do not have runtime type representations, then the Haskell AST will be tested for equality.

Added a new primitive `unstrict : () -> a` that runs a thunk ignoring any usage of strict mode. 

Turned `polyNeq` into a primitive so it'll use the users program state to determine whether or not to evaluate in strict mode.

# 

`polyNeq` was defined in the stdlib in strict mode. So, a program not in strict mode would still receive this restriction on equality comparison. Inversely, if `polyNeq` used `unstrict` then programs in strict mode would not have the restriction on eq cmp.  Now, `==` and `!=` will behave similarly in a user's program.

### DIFF
--- a/hs/src/Reach/AST/SL.hs
+++ b/hs/src/Reach/AST/SL.hs
@@ -677,7 +677,7 @@ data SLPrimitive
   | SLPrim_Token_destroyed
   | SLPrim_muldiv
   | SLPrim_didPublish
-  | SLPrim_Unstrict
+  | SLPrim_unstrict
   | SLPrim_polyNeq
   deriving (Eq, Generic)
 
--- a/hs/src/Reach/AST/SL.hs
+++ b/hs/src/Reach/AST/SL.hs
@@ -677,6 +677,8 @@ data SLPrimitive
   | SLPrim_Token_destroyed
   | SLPrim_muldiv
   | SLPrim_didPublish
+  | SLPrim_Unstrict
+  | SLPrim_polyNeq
   deriving (Eq, Generic)
 
 instance Equiv SLPrimitive where
--- a/hs/src/Reach/Eval/Core.hs
+++ b/hs/src/Reach/Eval/Core.hs
@@ -2985,9 +2985,8 @@ evalPrim p sargs =
       zero_args
       evalPrim (SLPrim_fluid_read FV_didSend) []
     SLPrim_unstrict -> do
-      one_arg >>= \case
-        clo@SLV_Clo {} -> locUseUnstrict True $ evalApply clo []
-        ow -> expect_t ow Err_Eval_NotApplicable
+      x <- one_arg
+      locUseUnstrict True $ evalApply x []
     SLPrim_polyNeq -> do
       (x, y) <- two_args
       notFn <- unaryToPrim (JSUnaryOpNot JSNoAnnot)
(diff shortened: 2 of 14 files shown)


hs/src/Reach/AST/SL.hs
```diff
@@ -677,6 +677,8 @@ data SLPrimitive
   | SLPrim_Token_destroyed
   | SLPrim_muldiv
   | SLPrim_didPublish
+  | SLPrim_Unstrict
```
jeapostrophe, line 680: This is the tiniest annoying nitpick, but this should be lower-case `u`

hs/src/Reach/Eval/Core.hs
```diff
@@ -275,6 +280,9 @@ data SLScope = SLScope
   , -- One for the consensus
     sco_cenv :: SLEnv
   , sco_use_strict :: Bool
+  -- Deliberately ignore strict mode when evaluating code
+  -- declared in strict mode
+  , sco_use_unstrict :: Bool
```
jeapostrophe, line 285: I feel like there must be a good reason why `unstrict` doesn't just set `use_strict` to `False`?

hs/src/Reach/Eval/Core.hs
```diff
@@ -2958,6 +2984,20 @@ evalPrim p sargs =
       ensure_after_first
       zero_args
       evalPrim (SLPrim_fluid_read FV_didSend) []
+    SLPrim_Unstrict -> do
+      one_arg >>= \case
+        SLV_Clo clo_at Nothing clo@(SLClo _ clo_args _ _) -> do
+          at <- withAt id
+          unless (null clo_args) $
+            expect_thrown clo_at $
+              Err_Apply_ArgCount at 0 (length clo_args)
```
jeapostrophe, line 2993: I don't think you should bother with this. What's the harm in allowing someone to write `unstrict((x = 1) => { ... })` or to see an application error if they write `unstrict((x) => { ... })`?

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
chrisnevers: If we were to set `use_strict` to `False` then it could be overwritten by a `'use strict'` statement in the the body of a closure. For example, the below should succeed:

```javascript
const f = () => {
  'use strict';
  const x = 5;
  return true;
}

  deploy();
  assert(unstrict(f));
});
```

A similar situation happens for a library that's written in strict mode since the strict mode is captured in the closure environment

jeapostrophe: Ah, yes, that makes sense.

hs/src/Reach/Eval/Core.hs
```diff
@@ -2984,14 +2984,9 @@ evalPrim p sargs =
...
-    SLPrim_Unstrict -> do
+    SLPrim_unstrict -> do
       one_arg >>= \case
-        SLV_Clo clo_at Nothing clo@(SLClo _ clo_args _ _) -> do
-          at <- withAt id
-          unless (null clo_args) $
-            expect_thrown clo_at $
-              Err_Apply_ArgCount at 0 (length clo_args)
-          locUseUnstrict True $ evalApply (SLV_Clo clo_at Nothing clo) []
+        clo@SLV_Clo {} -> locUseUnstrict True $ evalApply clo []
```
jeapostrophe, line 2989: I don't even think you should do that... just

```
x <- one_arg
locUseUnstrict .... x ...
```

`evalApply` will error if it needs to

jeapostrophe (approved):
Approved after this little thing

PR #303, opened 2021-10-05
CORE-20: Add `Contract` Type

chrisnevers:
Add `Contract` type, use it in `remote` and `ctc.getInfo`. The CBR form is a string and can either be constructed from a number for ALGO or an string address

### DIFF
--- a/docs-src/ref-programs-compute.scrbl
+++ b/docs-src/ref-programs-compute.scrbl
@@ -397,7 +397,7 @@ Reach's @deftech{type}s are represented in programs by the following identifiers
   @item{
     @margin-note{Reach has different representations of @tech{contract}s across @tech{connector}s.
       For example, on Algorand a @reachin{Contract} is an Application ID, but on Ethereum it is an Address.}
-    @(mint-define! '("Contract")) @reachin{Contract}, which denotes the connection information for a @tech{contract}.}
+    @(mint-define! '("Contract")) @reachin{Contract}, which denotes the identifying information of a @tech{contract}.}
   @item{@(mint-define! '("Token")) @reachin{Token}, which denotes a @tech{non-network token}. @secref["ref-networks"]{} discusses how @reachin{Token}s are represented on specific networks.}
   @item{@(mint-define! '("Fun")) @reachin{Fun([Domain_0, ..., Domain_N], Range)}, which denotes a @deftech{function type}, when @reachin{Domain_i} and @reachin{Range} are types.
   The domain of a function is @tech{negative position}.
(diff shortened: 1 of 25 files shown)


docs-src/ref-programs-compute.scrbl
```diff
@@ -394,6 +394,10 @@ Reach's @deftech{type}s are represented in programs by the following identifiers
   Bytes of different lengths are not compatible; however the shorter bytes may be @tech{padded}.}
   @item{@(mint-define! '("Digest")) @reachin{Digest}, which denotes a @tech{digest}.}
   @item{@(mint-define! '("Address")) @reachin{Address}, which denotes an @tech{account} @tech{address}.}
+  @item{
+    @margin-note{Reach has different representations of @tech{contract}s across @tech{connector}s.
+      For example, on Algorand a @reachin{Contract} is an Application ID, but on Ethereum it is an Address.}
+    @(mint-define! '("Contract")) @reachin{Contract}, which denotes the connection information for a @tech{contract}.}
```
jeapostrophe, line 400: I wouldn't say "connection information". I dislike the word "connection" and would just leave it at "information" or maybe "identifying information"

js/stdlib/ts/ALGO_compiled.ts
```diff
@@ -144,6 +145,21 @@ export const T_Address: ALGO_Ty<CBR_Address> = {
   }
 };
 
+export const T_Contract: ALGO_Ty<CBR_Contract> = {
+  ...CBR.BT_Contract,
+  netSize: 8,
+  toNet: (bv: CBR_Contract): NV => {
```
jeapostrophe, line 151: Can we reuse the fields from T_UInt rather than copy the code? (Either by literally making `toNet: T_UInt.toNet` or by calling that function?

js/stdlib/ts/CBR.ts
```diff
@@ -142,6 +144,27 @@ export const BV_Address = (val: string): CBR_Address => {
   return BT_Address.canonicalize(val);
 }
 
+export const BT_Contract: BackendTy<CBR_Contract> = ({
+  name: 'Contract',
+  canonicalize: (val: unknown): CBR_Contract => {
```
jeapostrophe, line 149: I think that this canonicalize either needs to reuse existing code (like my comment above) or it needs to be custom per-connector and do the same (i.e. this file will not expose `BT_Contract`, but will allow the connectors to define it directly, potentially using a `make_T_Contract` function. I think it would actually be the simplest to simply make `ALGO` define `T_Contract` literally as `T_UInt` and `ETH` define it as `T_Address`. In other words, there's a new NAME but not a new VALUE.

jeapostrophe: You might think we can't do this because "We need CBR_Contract" to be one thing for all the networks, but here's a dirty secret: the CBRs aren't actually really enforced or used anywhere, because TypeScript isn't expressive enough to track it for us.

jeapostrophe (changes requested):

PR #306, opened 2021-10-06
CORE-171

vinmazzi:


.circleci/make.sh
```diff
@@ -6,32 +6,33 @@ PRE=config.pre.yml
...
-          - "build"
-END
-    cat >>"${END}" <<END
-          - "examples.${CONN}.${RANK}"
-END
-  done
-done
-
-cat "${PRE}" "${MID}" "${END}" > config.gen.yml
+cat "${PRE}" > config.gen.yml
```
jeapostrophe, line 37: I think you should move this line up to 9 and add `exit 0` so it is more clear that this is just an experiment.

hs/Dockerfile.haskell-build-artifacts
```diff
@@ -1,14 +1,50 @@
+FROM ethereum/solc:0.8.7-alpine  AS solc
```
jeapostrophe, line 1: I would like all of the `0.8.7`s through this and the other files consolidated into one place that can be sourced by all the various Makefiles/etc (like `/DEPS` or something)

hs/Dockerfile.haskell-build-artifacts
```diff
@@ -1,14 +1,50 @@
...
+ENV OPENZEP_URL=https://github.com/OpenZeppelin/openzeppelin-contracts/archive/refs/tags/v$OPENZEP.tar.gz
+RUN	cd sol && wget $OPENZEP_URL
+
+ENV ERC20="sol/openzeppelin-contracts-${OPENZEP}/contracts/token/ERC20"
+RUN cd sol && tar xzvf v$OPENZEP.tar.gz
+RUN touch $ERC20/ERC20.sol
+
+RUN	ERC20=$ERC20 sol/stdlib_openzep.sol.sh > sol/stdlib_openzep.sol
+
+RUN cat sol/stdlib_openzep.sol sol/stdlib_reach.sol > sol/stdlib.sol
```
jeapostrophe, line 42: Ditto

js/stdlib/Dockerfile
```diff
@@ -2,23 +2,48 @@ ARG NODE_VERSION
...
+FROM node:14.17.0 as sol
+RUN curl -sSL https://git.io/get-mo -o mo
+RUN chmod +x mo
+RUN mv mo /usr/local/bin
+WORKDIR /stdlib
+RUN wget https://github.com/ethereum/solidity/releases/download/v0.8.7/solc-static-linux
+RUN mv solc-static-linux solc
+RUN chmod +x /stdlib/solc
+COPY ts /stdlib/ts
+RUN mo ts/version.mo.ts > ts/version.ts
```
jeapostrophe, line 21: Same complaint about duplicating the `Makefile` content

jeapostrophe: Perhaps the answer is that it should be removed from the Makefiles, I'm not sure

.circleci/config.pre.yml
```diff
@@ -275,11 +317,13 @@ jobs:
...
         echo Go team!
-
 workflows:
   docs:
     jobs:
-    - "docs-render"
+    - "docs-render":
+        filters:
+          tags:
+            only: /[0-9]*\.[0-9]*\.[0-9]*/
```
jeapostrophe, line 326: We always want to do this

.circleci/config.pre.yml
```diff
@@ -288,13 +332,42 @@ workflows:
        - "docs-render"
   buildless:
     jobs:
-    - "sh-check"
+    - "sh-check":
+        filters:
+          tags:
+            only: /[0-9]*\.[0-9]*\.[0-9]*/
```
jeapostrophe, line 338: And this

.circleci/config.pre.yml
```diff
@@ -140,7 +166,21 @@ jobs:
               }
             }]
           }
-
+  "install-mo":
```
jeapostrophe, line 169: This isn't used any more

.circleci/config.yml
```diff
@@ -17,4 +17,7 @@ setup: true
 workflows:
   setup:
     jobs:
-      - setup
+      - setup:
+          filters:
+            tags:
+              only: /[0-9]*\.[0-9]*\.[0-9]*/
```
jeapostrophe, line 23: We always want to do this

.circleci/make.sh
```diff
@@ -2,20 +2,38 @@
...
 cat >"${MID}" </dev/null
 cat >"${END}" </dev/null
 cat >"${IEND}" </dev/null
+cat >"${DEND}" </dev/null
 
 cat >>"${IEND}" <<END
     - "build-sink":
+        filters:
+          tags:
+            only: /[0-9]*\.[0-9]*\.[0-9]*/
```
jeapostrophe, line 20: We always want this

.circleci/make.sh
```diff
@@ -2,20 +2,38 @@
...
         requires:
           - "hs-test"
 END
 
+cat >>"${DEND}" <<END
+    - "hold":
```
jeapostrophe, line 26: My intention was that you would put this directly into `config.pre.yml`  and it would just require `"build-sink"`

.circleci/make.sh
```diff
@@ -61,6 +85,9 @@ image "fake" "rpc-server" "runner"
 
 cat >>"${END}" <<END
     - "examples-sink":
+        filters:
+          tags:
+            only: /[0-9]*\.[0-9]*\.[0-9]*/
```
jeapostrophe, line 90: We always want this

.circleci/make.sh
```diff
@@ -78,6 +105,9 @@ for CONN in ETH ALGO CFX ; do
     NAME="examples.${CONN}.${RANK}"
     cat >>"${MID}" <<END
     - "examples":
+        filters:
+          tags:
+            only: /[0-9]*\.[0-9]*\.[0-9]*/
```
jeapostrophe, line 110: And this

hs/Makefile
```diff
@@ -48,6 +48,9 @@ build-reach:
 .PHONY: build
 build: build-haskell-build-artifacts build-reach-cli build-reach
 
+.PHONY: build
+build: build-haskell build-reach-cli build-reachc
```
jeapostrophe, line 52: Duplicate target

js/stdlib/Dockerfile
```diff
@@ -1,17 +1,23 @@
 ARG NODE_VERSION
 ARG NODE_IMAGE=node:$NODE_VERSION
-
 ARG REACH_VERSION
+ARG SOLC_IMAGE
 ARG JS_DEPS_IMAGE=reachsh/js-deps:$REACH_VERSION
 
+FROM reachsh/haskell-build-artifacts as haskell
+
+FROM ethereum/solc:${SOLC_IMAGE} as solc
```
jeapostrophe, line 9: These are unused

js/stdlib/Dockerfile
```diff
@@ -1,17 +1,23 @@
 ARG NODE_VERSION
 ARG NODE_IMAGE=node:$NODE_VERSION
-
 ARG REACH_VERSION
+ARG SOLC_IMAGE
```
jeapostrophe, line 4: Unused

js/stdlib/Makefile
```diff
@@ -22,8 +22,11 @@ build:
 	  --build-arg NODE_VERSION=$(NODE_VERSION) \
 	  --build-arg REACH_VERSION=$(VERSION) \
 	  --build-arg REACH_GIT_HASH=$(REACH_GIT_HASH) \
+	  --build-arg SOLC_IMAGE=${SOLC_IMAGE} \
```
jeapostrophe, line 25: Unused

js/stdlib/Makefile
```diff
@@ -22,8 +22,11 @@ build:
...
 	  .
 	TAG_ONLY=1 $(ROOT)/scripts/docker-push.sh $(IMAGE)
+push:
+	$(ROOT)/scripts/docker-push.sh $(IMAGE)
```
jeapostrophe, line 29: Duplicate target

jeapostrophe (changes requested):

vinmazzi:
Unfortunately we do need the filter tags on all jobs otherwise it never gets executed when the build is off of a tag:

Screen Shot 2021-11-05 at 7 44 58 AM

.circleci/.gitignore
```diff
@@ -1,4 +1,5 @@
 /config.gen.yml
 /config.mid.yml
 /config.end.yml
+/config.dend.yml
```
jeapostrophe, line 4: Not needed

jeapostrophe (changes requested):

jeapostrophe:
Re: the tag filter thing. I think I must be dumb. Doesn't that mean it *only* runs when it is tagged? It reads to me like "There's a filter. Only run in these conditions: there is a tag". Does it not work like that?

jeapostrophe:
This has moved to a different PR

PR #319, opened 2021-10-16
docs: change `run` to `compile` b/c `compile` gens text

hamirmahal:
https://docs.reach.sh/tut-5.html says,

"Before we had this line in the file, when we ran `./reach run`, it would print out the message:"

```
  Verifying when ONLY "Alice" is honest
  Verifying when ONLY "Bob" is honest
Checked 18 theorems; No failures!
```

But, `./reach compile` is actually the command that prints out that message, no?

At least, it is for me:

```
tut % ./reach compile
  Verifying when ONLY "Alice" is honest
  Verifying when ONLY "Bob" is honest
Checked 18 theorems; No failures!
```

### DIFF
--- a/docs-md/src/tut.md
+++ b/docs-md/src/tut.md
@@ -582,7 +582,7 @@ ${code("/examples/tut-5-attack/index.rsh", 34, 37)}
 + Line 36 conducts the proof by including an assert statement in the program.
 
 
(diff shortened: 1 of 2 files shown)


jeapostrophe:
Thanks!

hamirmahal:
You're welcome!

PR #320, opened 2021-10-18
start snippet at start of change to help beginners

hamirmahal:
I was wondering why I couldn't reproduce exactly the terminal message in https://docs.reach.sh/tut-5.html after reducing Alice's payout.

It turns out, it was because I didn't undo the earlier changes in this lesson where we made Bob always win.

To help beginners, we can start [the code snippet that precedes the failed verification text](https://github.com/reach-sh/reach-lang/blob/master/examples/tut-5-attack/index-bad.rsh#L34-L41) at the start of the block that contains the first change since [the last code snippet](https://github.com/reach-sh/reach-lang/blob/master/examples/tut-5-attack/index.rsh#L34-L37), which I believe is line 27.

https://www.diffchecker.com/Ar0vXjwm

### DIFF
--- a/docs-src/tut.scrbl
+++ b/docs-src/tut.scrbl
@@ -628,8 +628,6 @@ Reach includes some such assertions automatically in every program.
 That's why every version of @|RPS| has said that a number of theorems were checked.
 We can see what these theorems do by deliberately inserting an error into the program.
 
-Let's change the computation of the payout and make it so that if Alice wins, then she only gets her wager back, not Bob's.
-
 Let's also undo the changes we made earlier by changing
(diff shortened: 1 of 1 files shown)


jeapostrophe:
I think that the problem you identify is real, but I don't think your solution is the best. I think the text needs to be more clear that we're undoing what we did before and going back to a different version, rather than just showing more code with less prose.

jeapostrophe:
Re: previewing --- The documentation is compiled and you need to compile it to see what it looks like.

Re: docs-md --- You should not be looking at this directory at all.

PR #352, opened 2021-11-04
CORE 759 add compiler err JSON type for safer dev

hamirmahal:
Since we know the ["shape" of the compiler error message JSON](https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/AST/Base.hs#L84), we can add that as a type to `vsce/server/src/server.ts`, to make development in `vsce/server/src/server.ts` less likely to break things.

Warning: This change accesses an error code from the compiler that's only available on release 09c295c6 and later. This shouldn't be a problem, since we want all users to be on version 09c295c6 or later, anyway.

Since we're assuming access to error codes from the Reach compiler is available, we can rewrite the conditional that checks for the error that has odd highlighting behavior by checking for its error code, `"RE0013"`, instead of by looking at its error message.

### DIFF
--- a/vsce/server/src/server.ts
+++ b/vsce/server/src/server.ts
@@ -350,7 +350,7 @@ export interface ErrorLocation {
 
 // Based on
 // https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/AST/Base.hs#L84
-type ReachCompilerErrorJSON {
+type ReachCompilerErrorJSON = {
 	ce_position: number[];
 	ce_suggestions: string[];
 	ce_errorMessage: string;
--- a/vsce/server/src/server.ts
+++ b/vsce/server/src/server.ts
@@ -388,7 +388,7 @@ CallStack (from HasCallStack):
 		const suggestions = errorJson.ce_suggestions;
 		const actualMessage = errorJson.ce_errorMessage;
 		const offendingToken = errorJson.ce_offendingToken;
-		const REACH_ERROR_CODE = errorJson.ce_errorCode;
+		const reachCompilerErrorCode = errorJson.ce_errorCode;
 
 		const start ={ line: linePos, character: charPos };
 		const end = offendingToken ?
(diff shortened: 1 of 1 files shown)


vsce/server/src/server.ts
```diff
@@ -369,27 +380,36 @@ CallStack (from HasCallStack):
...
-		const errorJson = JSON.parse(m[0].substring(7));
+		const errorJson: ReachCompilerErrorJSON = JSON.parse(m[0].substring(7));
 
 		//connection.console.log(`Tokens: ` + tokens);
 		const linePos = errorJson.ce_position[0] - 1;
 		const charPos = errorJson.ce_position[1] - 1;
 		const suggestions = errorJson.ce_suggestions;
 		const actualMessage = errorJson.ce_errorMessage;
 		const offendingToken = errorJson.ce_offendingToken;
+		const REACH_ERROR_CODE = errorJson.ce_errorCode;
```
jeapostrophe, line 391: My personal taste is that all caps is reserved for global things

jeapostrophe (dismissed):

vsce/server/src/server.ts
```diff
@@ -342,11 +342,22 @@ connection.onDidChangeWatchedFiles(_change => {
...
 	suggestions: string[]; // e.g. literally this whole thing: "declassify","array","assert","assume","closeTo"
 }
 
+// Based on
+// https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/AST/Base.hs#L84
+type ReachCompilerErrorJSON {
+	ce_position: number[];
+	ce_suggestions: string[];
+	ce_errorMessage: string;
+	ce_offendingToken?: string;
```
chrisnevers, line 357:
```suggestion
	ce_offendingToken: string | null;
```
This field will always be in the error JSON, and it is nullable.

hamirmahal: I'm glad you're a reviewer. Thank you!

vsce/server/src/server.ts
hamirmahal: I see. I personally prefer it for all constants, but I think it's important for a codebase to have consistent code style.

Right now, my constant declaration is the one that's breaking consistency, so I renamed `REACH_ERROR_CODE` out of `CONSTANT_CASE` with f9ef7dd096ee871dc5d647c88632e36743964397.

jeapostrophe (approved):

PR #355, opened 2021-11-05
CORE-149: Collect Info Needed for API functions

chrisnevers:
Collect the info needed for API functions and store it in `CPProg`

### DIFF
--- a/hs/src/Reach/APICut.hs
+++ b/hs/src/Reach/APICut.hs
@@ -98,7 +98,6 @@ instance Contains DLStmt where
     DL_Only _ (Right _) t -> has q t
     DL_Only {} -> False
     DL_MapReduce {} -> False
-    DL_setApiDetails {} -> False
 
 instance Contains DLTail where
   has q = \case
--- a/hs/src/Reach/APICut.hs
+++ b/hs/src/Reach/APICut.hs
@@ -98,6 +98,7 @@ instance Contains DLStmt where
     DL_Only _ (Right _) t -> has q t
     DL_Only {} -> False
     DL_MapReduce {} -> False
+    DL_setApiDetails {} -> False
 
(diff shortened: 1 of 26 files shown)


hs/src/Reach/APICut.hs
```diff
@@ -98,6 +98,7 @@ instance Contains DLStmt where
     DL_Only _ (Right _) t -> has q t
     DL_Only {} -> False
     DL_MapReduce {} -> False
+    DL_setApiDetails {} -> False
```
jeapostrophe, line 101: Adding statements is a major pain. Why is this not a new expression evaluated for effect? I find that easier to deal with because most of the passes ignore them

hs/src/Reach/AST/PL.hs
```diff
@@ -246,13 +246,16 @@ type ViewInfos = M.Map Int ViewInfo
 
 type CPViews = DLViews
 
+type ApiInfo = M.Map SLPart ([DLType], (Maybe String), Int)
```
jeapostrophe, line 249: I'd like the rhs of this to be a record type with names for the fields

hs/src/Reach/EPP.hs
```diff
@@ -171,6 +171,8 @@ data BEnv = BEnv
   , be_view_setsr :: IORef ViewSet
   , be_inConsensus :: Bool
   , be_counter :: Counter
+  , be_which :: Int
+  , be_api_info :: IORef (M.Map SLPart ([DLType], (Maybe String), Int))
```
jeapostrophe, line 175: Here's a place where that named typed will come in handy. I think it would live in DLBase

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #358, opened 2021-11-07
CORE-660 extract hard-coded keyword documentation

hamirmahal:
We eventually want the compiler to output data files that Reach's language server can use.

This change lays the foundation for that by extracting the hard-coded keyword documentation we have currently into a separate directory that Reach's compiler will eventually populate.

`set` originally mapped to two documentation strings: [it mapped to documentation for `Participant.set` on line 612](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/server.ts#L612), yet [it would also map to documentation for `Array.set` on line 636](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/server.ts#L636).

We can't have a `Duplicate object key json(520)`, so I just mapped `set` to the documentation for `Participant.set`. That's the behavior that it had originally, anyway, since that conditional would execute first in the `else if` statements.


vsce/server/src/server.ts
```diff
@@ -39,6 +39,17 @@ import {
...
 
+// Do this import differently so we can add types, to avoid a
+// "No index signature with a parameter of type 'string' was found... ts(7053)"
+// error later. See
+// https://stackoverflow.com/questions/42986950/how-to-define-import-variable-type
+// Also, if we do this import this way, we don't have to add
+// "resolveJsonModule": true,
+// to tsconfig.json.
+const KEYWORD_TO_DOCUMENTATION: { [ keyword: string ] : string } = require(
+	'../../compiler_outputted_files/keywordToDocumentation.json'
```
jeapostrophe, line 50: In a future push (when you have something else to add also), I think you should name this directory `data` rather than the long name or just put it in the `/vsce/`

jeapostrophe (approved):

vsce/server/src/server.ts
hamirmahal: Just to be explicit, the path on line 50 will read

`'../../data/keywordToDocumentation.json'?

hamirmahal: Just to be explicit, the path on line 50 will read

`'../../data/keywordToDocumentation.json'`?

hamirmahal: #360 should solve this. Thanks for the suggestion!

PR #359, opened 2021-11-08
Added Algolia Search and miscellaneous

hagenhaus:
1. Added Algolia Search so the plumbing is in place. Currently, there are only 7 records in the search index. Adding more records, implementing advanced search features, and refining search results are pretty easy. The most important effort now is the establish content organization (urls) and get the content in place.
2. Added For Teachers section in the Essentials book. 
3. Other small changes.


docs/dev/assets.in/scripts.js
```diff
@@ -1,3 +1,7 @@
+import algoliasearch from 'https://cdn.jsdelivr.net/npm/algoliasearch@4/dist/algoliasearch-lite.esm.browser.js';
+const client = algoliasearch('WTEWG2I8PH', 'cdf7770e846ff4e95802411b91afbba6');
```
jeapostrophe, line 2: I believe that Chris can hook you up with the account he set up for this.

docs/dev/bak/search.json
```diff
@@ -0,0 +1,113 @@
+[
```
jeapostrophe, line 1: I assume this is the actual search content? And we'll have to have the generator create this?

jeapostrophe (approved):

docs/dev/assets.in/scripts.js
hagenhaus: Yes, if you'll accept this PR, I'll get access to Chris' Algolia account, update that index, get the creds, update script.js, test, etc. Thx

docs/dev/bak/search.json
hagenhaus: Yes, this is real search content. Initially, I am glad to create a record for each RDP page in the search index. It doesn't take long. This will get us started. We don't need to update the search index every time someone submits a PR with a few changes, so, if we do hook the search-index-generation process into the generator, which I think is a good idea, then I think kicking off the process needs to be conditional in some way, either by a human-set command-line flag, or by some programmatic condition that I can't think of.

docs/dev/assets.in/scripts.js
jeapostrophe: It's accepted, you can merge it

PR #360, opened 2021-11-08
rename directory for brevity; devs can see README

hamirmahal:
This change just renames the directory of the files that the compiler will eventually output from `compiler_outputted_files/` to `data/`.

I added a `README.md` to this directory previously, anyway, so anyone curious what the purpose of this directory is can just read that.


jeapostrophe (changes requested):
The use of the file in the code would need to change too

hamirmahal:
Ah! Thank you. I forgot about that.

2eab2e50ca08875a482047321d1177b8e651f5ff should address that.

hamirmahal:
Ah! Thank you. I forgot about that.

2eab2e50ca08875a482047321d1177b8e651f5ff should address that.

jeapostrophe (approved):

PR #363, opened 2021-11-09
CORE-662 extract hard-coded data for compiler

hamirmahal:
This change extracts hard-coded array methods and keywords from `server/src/keywordCompletion.ts` and places them in `data/arrayMethods.json`, and `data/keywordToItemKind.json`, respectively.

We hard-code the type of the import that specifies part of the `CompletionItemKind` with

```
{
  [keyword: string] : "Text" | "Method" | "Function" |
    "Constructor" | "Field" | "Variable" | "Class" |
    "Interface" | "Module" | "Property" | "Unit" | "Value" |
    "Enum" | "Keyword" | "Snippet" | "Color" | "File" |
    "Reference" | "Folder" | "EnumMember" | "Constant" |
    "Struct" | "Event" | "Operator" | "TypeParameter"
}
```

because that avoids an error from TypeScript, later, that complains about invalid index signatures.


vsce/server/src/keywordCompletion.ts
```diff
@@ -1,160 +1,53 @@
 import { CompletionItemKind } from 'vscode-languageserver';
 
+// Use a `require` statement instead of an `import`
+// statement to avoid the following error from TypeScript:
+// File '../../data/arrayMethods.json' is not under 'rootDir'
+// 'rootDir' is expected to contain all source files.ts(6059)
+const ARRAY_METHODS: string[] = require(
+  "../../data/arrayMethods.json"
```
jeapostrophe, line 8: I don't think you should make this a separate file. You should just have one.

jeapostrophe (changes requested):

vsce/server/src/keywordCompletion.ts
hamirmahal: Hmm.... This could be done easily if we require the compiler to do the prefixing.

In other words, if we make the compiler output

```
  "Array.all": "Method",
  "Array.and": "Method",
  "Array.any": "Method",
  "Array.average": "Method",
  "Array.concat": "Method",
  "Array.count": "Method",
  "Array.empty": "Method",
  "Array.findIndex": "Method",
  "Array.forEach": "Method",
...
```
instead of
```
  "all",
  "and",
  "any",
  "average",
  "concat",
  "count",
  "empty",
  "findIndex",
  "forEach",
...
```
then this change is really easy, and I can just add

```
  "Array.all": "Method",
  "Array.and": "Method",
  "Array.any": "Method",
  "Array.average": "Method",
  "Array.concat": "Method",
  "Array.count": "Method",
  "Array.empty": "Method",
  "Array.findIndex": "Method",
  "Array.forEach": "Method",
...
```

to `data/keywordToItemKind.json` and delete the `addPrefix` method altogether. But, then we lose
```
  "all",
  "and",
  "any",
  "average",
  "concat",
  "count",
  "empty",
  "findIndex",
  "forEach",
...
```
as keywords, unless we do more finagling. And, by that, I mean we'd have to do a `keyword.split('.')[1]` on each of
```
  "Array.all": "Method",
  "Array.and": "Method",
  "Array.any": "Method",
  "Array.average": "Method",
  "Array.concat": "Method",
  "Array.count": "Method",
  "Array.empty": "Method",
  "Array.findIndex": "Method",
  "Array.forEach": "Method",
...
``` 
to keep
```
  "all",
  "and",
  "any",
  "average",
  "concat",
  "count",
  "empty",
  "findIndex",
  "forEach",
...
```
as keywords. Do we want to do that?

jeapostrophe: You should make it so that this code is as dumb as possible and it is the compiler's job to output a complete list of all the keywords, with the `Array.` if they exist and without it if they don't. This will be slightly annoying for you right now as you make it by hand, but it will be easy for the compiler (actually, easier than the current way)

hamirmahal: I see. To clarify, the compiler will eventually output
```
{
  "Array.all": "Method",
  "Array.and": "Method",
  "Array.any": "Method",
  "Array.average": "Method",
  "Array.concat": "Method",
  "Array.count": "Method",
  "Array.empty": "Method",
  "Array.findIndex": "Method",
  "Array.forEach": "Method",
...
  "all": "Method",
  "and": "Method",
  "any": "Method",
  "average": "Method",
  "concat": "Method",
  "count": "Method",
  "empty": "Method",
  "findIndex": "Method",
  "forEach": "Method",
...
}
```
in `data/keywordToItemKind.json`?

If this is what we're getting from the compiler, there are a lot of things we can erase in `server/src/keywordCompletion.ts`, making that file much cleaner and simpler.

As I'm implementing this locally, I see that we're getting that `Duplicate object key json(520)` error again, except this time for

1. `"concat"`
2. `"forEach"`
3. `"length"`
4. `"map"`
5. `"reduce"`
6. `"set"`
7. `"zip"`

in `data/keywordToItemKind.json`.

Indeed, in [`vsce/server/src/keywordCompletion.ts`](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts), we define

1. `"concat"` as a [`CompletionItemKind.Method` on line 10](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L10) and as a [`CompletionItemKind.Function` on line 56](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L56) 
2. `"forEach"` as [`  'forEach': CompletionItemKind.Method,` on line 14](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L14) and [`  'forEach': CompletionItemKind.Function,` on line 70](https://github.com/reach-sh/reach-lang/blob/6bbb2ddc762f5daaf5c3b64e090e6fec47dcb801/vsce/server/src/keywordCompletion.ts#L70)
3. `"length"` as [`  'length': CompletionItemKind.Method,` on line 18](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L18) and [`  'length': CompletionItemKind.Function,` on line 119](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L119)
4. `"map"` as [`  'map': CompletionItemKind.Method,` on line 19](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L19) and [`	'map': CompletionItemKind.Function,` on line 122](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L122)
5. `"reduce"` as [`  'reduce': CompletionItemKind.Method,` on line 24](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L24) and [`  'reduce': CompletionItemKind.Function,` on line 138](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L138)
6. `"set"` as [`  'set': CompletionItemKind.Method,` on line 26](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L26) and [`  'set': CompletionItemKind.Function,` on line 141](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L141)
7. `"zip"` as [`  'zip': CompletionItemKind.Method,` on line 28](https://github.com/reach-sh/reach-lang/blob/6bbb2ddc762f5daaf5c3b64e090e6fec47dcb801/vsce/server/src/keywordCompletion.ts#L28) and [`  'zip': CompletionItemKind.Function,` on line 157](https://github.com/reach-sh/reach-lang/blob/master/vsce/server/src/keywordCompletion.ts#L157)

So, for each of those 7 duplicate keys, we're defining a keyword once as a `CompletionItemKind.Method`, and once as a `CompletionItemKind.Function`.

I'm going to arbitrarily have all 7 of these keywords be a `CompletionItemKind.Function`, unless you have any objections.

jeapostrophe: Exactly. All good.

hamirmahal: Okay, thank you for the clarification!

eab9e01 should address all the points we made in this discussion.

jeapostrophe (approved):

PR #364, opened 2021-11-09
CORE-632: add additional analytics

mattaudesse:


hs/src/Reach/Report.hs
```diff
@@ -16,8 +16,18 @@ import Reach.Version
...
-startReport mwho = do
+data Initiator
+  = LogCompile
+  | LogDevnetCreate
+  | LogDevnetDaily
+  | LogReact
+  | LogRpcRun
+  | LogRpcServer
+  | LogRun
+  deriving (Read, Show)
```
jeapostrophe, line 27: This whole change is great and a good example of how nice this new Haskell setup is. My only comment is on this line right here. I feel like it is unnecessarily complicated to mandate that the initiators come from a fixed vocabulary. Perhaps it makes it so we're unlikely to have a typo and write `LogDevntCreate`, but I'm not super worried about this. This is a "light" recommendation

mattaudesse: Yeah... Using `Read` is ugly too.

Shall I just make this stringly-typed? Another option would be to create a `LogSomethingElseTagged Text` constructor and fall back to that when `readMay` gives `Nothing`.

mattaudesse: Another thought I had while doing all this: I dislike the way I've replicated `--disable-reporting` among individual sub-commands rather than making it work like `reach --disable-reporting compile ...`.

Since it's undocumented I think we can punt on moving it until later (assuming you agree it should be consolidated). Do you have an opinion in either direction?

jeapostrophe: My preference is Initiator=String

I care about the issue of consolidation for code-cleanliness. If making it consolidated is less code, than pay for that work now rather than later when it spreads

mattaudesse: Thanks 👍🏼

jeapostrophe (approved):

PR #367, opened 2021-11-10
Skip `remote-view` on CFX

chrisnevers:


examples/remote-view/index.mjs
```diff
@@ -7,7 +7,7 @@ import * as fs from 'fs';
 const stdlib = loadStdlib(process.env);
 
 (async () => {
-  if ( stdlib.connector === 'ALGO' ) { return; }
+  if ( ['ALGO', 'CFX'].includes(stdlib.connector) ) { return; }
```
jeapostrophe, line 10: I think my preference is to continue running it on CFX, but add it to xfail, so we don't forget about the error by losing it in some old Jira card in the backlog.

examples/remote-view/index.mjs
```diff
@@ -37,8 +42,9 @@ const stdlib = loadStdlib(process.env);
       keepGoing: async () => {
         const kg = count < 5;
         const ctcAddr = await ctcAlice.getContractAddress();
-        const x = await contract.callStatic["callView"](ctcAddr);
-        console.log(`Compare: `, x.toNumber(), count);
+        const f = (stdlib.connector == 'CFX') ? contract : contract.callStatic;
```
jeapostrophe, line 45: Perhaps this means that we should change `cfxers` to support `callStatic`?

chrisnevers: I'll make a ticket for that

examples/remote-view/index.mjs
chrisnevers: Looks like it was added to xfail two days ago.. I'll close this PR

PR #368, opened 2021-11-10
CORE-792 add documentation for `Bool`

hamirmahal:
I somehow missed this in the migration. I believe it's the only keyword I missed. There's code [on the Jira ticket for this](https://reachsh.atlassian.net/browse/CORE-792) that one can run to verify this.

[`Bool` has the same markdown documentation as `Null`](https://github.com/reach-sh/reach-lang/pull/358/files#diff-153b25dbdb8a302038da1305689c4002a5ebbadb38382d52c1d2b3ae73a45c63L704), so we just add an extra entry in `data/keywordToDocumentation.json` with the same value as `Null`, but a different key: `Bool`.

Without this change:

Screen Shot 2021-11-10 at 11 05 17 AM

With this change:

Screen Shot 2021-11-10 at 11 58 27 AM


hamirmahal:
This is a part of #358.

jeapostrophe (approved):

PR #370, opened 2021-11-11
UPDTE: RAPI0002 docs with example snippet

AwolDes:
## Summary

Had a confusing error today. In an `API_CONSENSUS_EXPR` block there was the following snippet:
```javascript
.api(User.mintTok, (algoAmount, apiReturn) => {
    require(algoAmount > 0);
    require(oraclePrice > 0);
    const mint = algoAmount * oraclePrice;
    if(balance(tok) > mint) {
      transfer(mint, tok).to(this);
      apiReturn(true);
    }
    apiReturn(false);
  }
})
```

Calling `apiReturn` twice was the problem, however the error in this section of the docs was slightly misleading to begin with, and we were asking the question "How can User.mintTok be called twice?". This update to the docs clarifies what `RAPI0002` might mean.

### DIFF
--- a/docs/scrbl/ref-error-codes.scrbl
+++ b/docs/scrbl/ref-error-codes.scrbl
@@ -2466,7 +2466,7 @@ The error means that you use an @tech{API} in two places in your program, which 
 
 This might look like the following in the API_CONSENSUS_EXPR
 @reach{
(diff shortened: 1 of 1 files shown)


jeapostrophe:
Thank you!

PR #377, opened 2021-11-12
CORE-758 wait for comp to finish before comp again

hamirmahal:
This change seeks to fix a bug users experience where Reach's language server automatically compiles a Reach program on every keystroke.

This fix attempts to solve that problem by using a flag that we reset once compilation has finished. While that flag is set, i.e., while a compilation is already in process, future compilations will not occur.

We can be sure that compilation has finished when the callback for `exec` exectutes, because the documentation for `child_process.exec()` says that it [spawns a shell and runs a command within that shell, passing the stdout and stderr to a callback function **when complete**.](https://nodejs.org/api/child_process.html#child-process)

This approach might be preferable to setting a timeout that waits for the user to finish typing, since different users have different devices, and those with faster devices may not need to wait as long as those with slower ones for compilation completion.

### DIFF
--- a/vsce/server/src/server.ts
+++ b/vsce/server/src/server.ts
@@ -229,7 +229,8 @@ documents.onDidChangeContent(change => {
 	validateTextDocument(change.document);
 });
 
-let theCompilerIsCompiling = false;
+let	theCompilerIsCompiling 	= false,
+	weNeedToCompileAgain 	= false;
 
 async function validateTextDocument(textDocument: TextDocument): Promise<void> {
 
@@ -290,8 +291,9 @@ async function validateTextDocument(textDocument: TextDocument): Promise<void> {
 		: exeLoc;
 
 	if (theCompilerIsCompiling) {
+		weNeedToCompileAgain = true;
 		console.debug(
-			"Compilation is already in process! Returning...",
+			"Compilation already in process; will recompile",
 			new Date().toLocaleTimeString()
 		);
 		return;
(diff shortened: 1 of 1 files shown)


vsce/server/src/server.ts
```diff
@@ -286,7 +288,34 @@ async function validateTextDocument(textDocument: TextDocument): Promise<void> {
 	const reachPath = (exeLoc == '' || exeLoc == './reach')
 		? path.join(workspaceFolder, "reach")
 		: exeLoc;
+
+	if (theCompilerIsCompiling) {
+		console.debug(
```
jeapostrophe, line 293: My comment in the call was that you could do `doAnotherCompile = true` here

vsce/server/src/server.ts
```diff
@@ -286,7 +288,34 @@ async function validateTextDocument(textDocument: TextDocument): Promise<void> {
...
+		// when this compilation command has finished.
+		// "child_process.exec(): spawns a shell...
+		// passing the stdout and stderr to a callback
+		// function when complete". See
+		// https://nodejs.org/api/child_process.html#child-process
+		console.debug(
+			"Compilation should now have finished.",
+			new Date().toLocaleTimeString()
+		);
+		theCompilerIsCompiling = false;
```
jeapostrophe, line 317: And then here, do `if ( doAnotherCompile ) { goAgain(); }`

jeapostrophe (approved):

vsce/server/src/server.ts
hamirmahal: I see. 1304276 should address this.

Thank you for adding the suggestions to a specific line: that makes things a lot easier for me.

vsce/server/src/server.ts
hamirmahal: Thank you for the suggestion, and specifically for pointing out the line to which to add the suggestion: that's helpful.

1304276 should address this.

jeapostrophe (approved):

PR #378, opened 2021-11-12
[WIP] CORE-149: Add External Functions To Call API

chrisnevers:
Still need to add an example that tests this out


hs/src/Reach/AST/DLBase.hs
```diff
@@ -530,8 +534,13 @@ data DLExpr
   -- * the either is whether it is a generated one or a prescribed one
   -- * the type is the type
   -- * the dlarg is the value being logged
-  | DLE_EmitLog SrcLoc String DLVar
-  | DLE_setApiDetails SrcLoc SLPart [DLType] (Maybe String)
+  | DLE_EmitLog SrcLoc String (Maybe String) DLVar
+  | DLE_setApiDetails {
+    sed_at :: SrcLoc,
```
jeapostrophe, line 539: What does `sed` stand for?

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1269,6 +1394,7 @@ solPLProg (PLProg _ plo dli _ _ (CPProg at (vs, vi) _ai hs)) = do
   ctxt_ints <- newIORef mempty
   ctxt_outputs <- newIORef mempty
   ctxt_tlfuns <- newIORef mempty
+  ctxt_api_rngs <- newIORef $ if M.null ai then Nothing else Just mempty
```
jeapostrophe, line 1397: Why a maybe?

hs/src/Reach/EPP.hs
```diff
@@ -352,10 +352,10 @@ be_m = \case
     case de of
       DLE_Remote {} -> recordOutputVar mdv
       DLE_EmitLog {} -> fg_use de
-      DLE_setApiDetails _ p tys mc -> do
-        which <- asks be_which
+      DLE_setApiDetails _ p tys mc isf -> do
+        (which, msg) <- asks be_which_msg
         api_info <- asks be_api_info
-        liftIO $ modifyIORef api_info $ M.insert p $ ApiInfo tys mc which
+        liftIO $ modifyIORef api_info $ M.insert p $ ApiInfo tys mc which isf msg
```
jeapostrophe, line 358: Why does the API need them?

hs/src/Reach/Eval/Core.hs
```diff
@@ -4351,7 +4369,7 @@ doApiCall lhs (ApiCallRec{..}) = do
   let returnVal = jidg "rng"
   let returnLVal = jidg "rngl"
   let interactOut = jsCall a (mkIdDot a ["interact", "out"]) [dom, returnLVal]
-  let doLog = jsCall a (jid ".emitLog") [ jidg "rng" ]
+  let doLog = jsCall a (jid ".emitLog") [ jidg "rng", who_str ]
```
jeapostrophe, line 4372: Why do logs need to know the name of the participant?

jeapostrophe: (i.e. you could just allocate a new variable and associate it with the participant)

hs/src/Reach/Optimize.hs
```diff
@@ -336,8 +336,8 @@ instance Optimize DLExpr where
     DLE_TimeOrder at tos -> DLE_TimeOrder at <$> opt tos
     DLE_GetContract at -> return $ DLE_GetContract at
     DLE_GetAddress at -> return $ DLE_GetAddress at
-    DLE_EmitLog at m a -> DLE_EmitLog at m <$> opt a
-    DLE_setApiDetails at w t c -> return $ DLE_setApiDetails at w t c
+    DLE_EmitLog at m ma a -> DLE_EmitLog at m ma <$> opt a
+    DLE_setApiDetails {..} -> return $ DLE_setApiDetails {..}
```
jeapostrophe, line 340: You should write it out so that if we ever add more we will be forced to think about if they need to be freshened

hs/src/Reach/Subst.hs
```diff
@@ -100,8 +100,8 @@ instance Subst DLExpr where
     DLE_TimeOrder at tos -> DLE_TimeOrder at <$> subst tos
     DLE_GetContract at -> return $ DLE_GetContract at
     DLE_GetAddress at -> return $ DLE_GetAddress at
-    DLE_EmitLog at m x -> DLE_EmitLog at m <$> subst x
-    DLE_setApiDetails at who ts ci -> return $ DLE_setApiDetails at who ts ci
+    DLE_EmitLog at m ma x -> DLE_EmitLog at m ma <$> subst x
+    DLE_setApiDetails {..} -> return $ DLE_setApiDetails {..}
```
jeapostrophe, line 104: Ditto

hs/src/Reach/AST/DLBase.hs
```diff
@@ -486,16 +486,20 @@ type DLTimeArg = Either DLArg DLArg
 data ApiInfo = ApiInfo
   { ai_msg_tys :: [DLType]
   , ai_mcase_id :: Maybe String
-  , ai_which :: Int }
+  , ai_which :: Int
+  , ai_is_fork :: Bool
```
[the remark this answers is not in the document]
chrisnevers: Okay, I'll look into this

hs/src/Reach/CollectTypes.hs
```diff
@@ -132,7 +132,7 @@ instance CollectsTypes DLExpr where
     DLE_TimeOrder _ tos -> cts tos
     DLE_GetContract _ -> mempty
     DLE_GetAddress _ -> mempty
-    DLE_EmitLog _ _ a -> cts a
+    DLE_EmitLog _ _ _ a -> cts a
```
hamirmahal, line 135: Do the `_ _ _` here denote wildcards?

jeapostrophe: Yea

hamirmahal: I see. Thank you for the clarification.

hs/src/Reach/EPP.hs
```diff
@@ -171,7 +171,7 @@ data BEnv = BEnv
   , be_view_setsr :: IORef ViewSet
   , be_inConsensus :: Bool
   , be_counter :: Counter
-  , be_which :: Int
+  , be_which_msg :: (Int, [DLVar])
```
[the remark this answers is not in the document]
chrisnevers: that makes sense, I'll update this

jeapostrophe (commented):
I think your example should ALSO have two Reach programs talk to each other.

hs/src/Reach/Eval/Core.hs
chrisnevers: makes sense, Ill add this

jeapostrophe (commented):
This is cool!

hs/src/Reach/Eval/Core.hs
chrisnevers: I'm actually not sure I understand because if you have a map of `int - slval` you'd still have to expand it in `doEmitLog` and expect a participant. Unless you mean, handle the expansion in solidity generation, so the map would have to be passed through the `program` field?

jeapostrophe: I don't know if this is a good idea.

What I mean is that `EmitLog` already has a `DLVar`, which is the one that it is outputting. For example, it says `DLE_EmitLog _ _ v22` and that `v22` is tied to this log entry. We use it for the event value in Solidity and in the ALGO log header. What you could do is make it so that `doEmitLog` (in `Core`) associates `v22` with `Flower_girl` (i.e. each API has an attribute of "What is the variable that my output comes from?")

Then, every logged value gets put inside the memory table for potential returning.

Why could this be better? It is a more "general" feature rather than tying things specifically to API names.

chrisnevers: I think the issue with this approach is that every `emitLog` used by APIs is generated via JavaScript and we cannot know the DLVar associated with the log entry when generating the JavaScript. So, any association made between `v22` and `Flower_girl` would have to be as a result of passing arguments to the `emitLog` function. If I'm still not understanding this, maybe you can provide an example of something that would utilize the generalness of what you're describing so I can better reason about the compilation path..

jeapostrophe: I don't know what you mean by generated via JavaScript... these emit logs are generated inside of Core and they get compiled to Solidity & JS & TEAL

chrisnevers: Sorry,  I meant we generate Javascript, `.emitLog(..)`, in Core that is later evaluated

jeapostrophe (approved):

jeapostrophe:
Conflicts :'(

PR #396, opened 2021-11-19
CORE-762: Warn about using Objects in Views and APIs

chrisnevers:


docs/scrbl/ref-error-codes.scrbl
```diff
@@ -2548,3 +2548,9 @@ This warning indicates that your program does not contain any publications.
 
 You can fix this issue by making sure at least one @reachin{Participant} performs a @reachin{publish}.
 
+@error{RW0005}
+
+This warning indicates that a @reachin{View} or @reachin{API} produces or consumes an @reachin{Object},
+which is a type internal to Reach.
```
jeapostrophe, line 2554: I would expand this out and say what this mean: it has an opaque and unspecified representation that can only be consumed by other Reach programs, so it is probably a bad choice for general purpose interfaces.

chrisnevers: I updated the docs

jeapostrophe (approved):

PR #401, opened 2021-11-20
CORE-4: RPC views and exports

mattaudesse:
~WIP~


examples/rpc-kitchen-sink/index.rsh
```diff
@@ -0,0 +1,30 @@
...
+  const [ v1, v2 ] = [ View('Main', v), View(v) ];
+
+  deploy(); A.publish(); commit();
+
+  runView.vbytes(A, v1, Bytes(10).pad('Main/who'), Bytes(10).pad('Main/meta'));
+  runView.vbytes(A, v2, Bytes(10).pad('who'),      Bytes(10).pad('meta'));
+
+  runView.fun(A, v1, Bytes(10).pad('Main/fun'));
+  runView.fun(A, v2, Bytes(10).pad('fun'));
+
```
jeapostrophe, line 28: I'd put another `A.publish()` here so that the last step of the `runView` isn't in an empty context

examples/rpc-kitchen-sink/Makefile
```diff
@@ -0,0 +1,50 @@
...
+
+.PHONY: build
+build: client-lib-js build/index.main.mjs
+	docker build \
+	  --tag=$(CIMG) \
+	  --build-arg NODE_VERSION=$(NODE_VERSION) \
+	  .
+
+run:
+	@# Start fresh RPC server to ensure server-log is populated correctly
```
jeapostrophe, line 36: Why does `reach rpc-run` work for this example? I am anxious about duplicating the way that works yet again

examples/rpc-kitchen-sink/index.mjs
```diff
@@ -0,0 +1,53 @@
...
+
+  const checkViewBytes = async (l1, l2, e) => {
+    await assertEq(`ctc.v     > ${l1} + ${l2}:`, e, [ await rpc(`/ctc/v/${l1}`,     ctcAlice),
+                                                      await rpc(`/ctc/v/${l2}`,     ctcAlice) ]);
+    await assertEq(`ctc.views > ${l1} + ${l2}:`, e, [ await rpc(`/ctc/views/${l1}`, ctcAlice),
+                                                      await rpc(`/ctc/views/${l2}`, ctcAlice) ]);
+  };
+
+  /* Exports */
+  // TODO: how should we access nested elements in arrays?
```
jeapostrophe, line 37: I don't understand what this means

jeapostrophe: Maybe you mean that we currently have a way of turning paths like `/a/b/c` into object references like `a.b.c` and you're wondering about how to do `a.b[c]`?

examples/rpc-kitchen-sink/tls/reach-server.crt
```diff
@@ -0,0 +1,30 @@
+-----BEGIN CERTIFICATE-----
```
jeapostrophe, line 1: These files wouldn't be needed if you use `reach rpc-run`

examples/rpc-kitchen-sink/index.rsh
```diff
@@ -0,0 +1,29 @@
...
+  const A = Participant('Alice', {
+    ...runView.pii,
+  });
+
+  const [ v1, v2 ] = [ View('Main', runView.T), View(runView.T) ];
+
+  setOptions({ verifyArithmetic: true, verifyPerConnector: true });
+  deploy(); A.publish(); commit();
+
+  // Temporarily disable until ETH gas estimation + arbitrary view-checking order are fixed
```
jeapostrophe, line 20: What does this mean?

examples/rpc-kitchen-sink/Makefile
mattaudesse: I can probably make `rpc-run` work here. The main idea was just to ensure we spit out a detailed `server-log` in case of failure, and we didn't really have an off-the-shelf answer to that problem before.

examples/rpc-kitchen-sink/index.mjs
mattaudesse: Yup, we've now got a way to do `/a/b/c` for `a.b.c` if it's just a nested object of objects, but if `b = [ 'c', { d: 'e' } ]`, how should we pluck out `a.b[0]` or `a.b[1]`? (Or `d`?)

examples/rpc-kitchen-sink/index.rsh
mattaudesse: The gas part is referring to the ETH CI failure I mentioned:
https://77248-205868997-gh.circle-artifacts.com/0/tmp/artifacts/ETH.rpc-kitchen-sink.gz

The arbitrary order part is what I was saying about reshuffling the order in which `runView.vbytes` and `runView.fun` are executed. I feel like I should be able to run them in any order but I must be missing something to allow that to succeed.

examples/rpc-kitchen-sink/index.mjs
jeapostrophe: I think you need to make up something or look for precedent like XPath, JsonPath, or the Go template syntax, etc. FWIW, I think JS weirdness means that `/a/b/1/d` will "just work"

examples/rpc-kitchen-sink/index.rsh
jeapostrophe: First, that error means that the program is too big. There's a maximum size of ETH programs and currently our view code is not as efficient as it could be.

Second, what error do you get when you switch the order?

mattaudesse: (Using `ALGO`)

If I move the `runView.fun` lines from 24 and 25 above the `runView.vbytes` lines I get:
```
backend.getExports > app.timesTwoPlusThree(app.five): {"type":"BigNumber","hex":"0x0d"} === {"type":"BigNumber","hex":"0x0d"}
backend.getExports > app.a.b.e: [{"type":"BigNumber","hex":"0x05"},{"type":"BigNumber","hex":"0x06"}] === [{"type":"BigNumber","hex":"0x05"},{"type":"BigNumber","hex":"0x06"}]
backend.getExports > app.a.b.c: "d" === "d"
backend.getExports > app.a.z.c: null === null
backend.getExports > app.nonExistent: null === null
backend.getExports > app.nonExistent.a: null === null
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > Main/fun: *** Mismatch! ***
  Expected: ["Some",true]
    Actual: ["None",null]
```

If I run just the `runView.fun` lines and comment both `runView.vbytes` lines it succeeds:
```
backend.getExports > app.timesTwoPlusThree(app.five): {"type":"BigNumber","hex":"0x0d"} === {"type":"BigNumber","hex":"0x0d"}
backend.getExports > app.a.b.e: [{"type":"BigNumber","hex":"0x05"},{"type":"BigNumber","hex":"0x06"}] === [{"type":"BigNumber","hex":"0x05"},{"type":"BigNumber","hex":"0x06"}]
backend.getExports > app.a.b.c: "d" === "d"
backend.getExports > app.a.z.c: null === null
backend.getExports > app.nonExistent: null === null
backend.getExports > app.nonExistent.a: null === null
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > Main/fun: ["Some",true] === ["Some",true]
ctc.views > Main/fun: ["Some",true] === ["Some",true]
ctc.v     > Main/fun: ["Some",false] === ["Some",false]
ctc.views > Main/fun: ["Some",false] === ["Some",false]
ctc.v     > Main/fun: ["Some",false] === ["Some",false]
ctc.views > Main/fun: ["Some",false] === ["Some",false]
ctc.v     > Main/fun: ["Some",false] === ["Some",false]
ctc.views > Main/fun: ["Some",false] === ["Some",false]
ctc.v     > Main/fun: ["Some",true] === ["Some",true]
ctc.views > Main/fun: ["Some",true] === ["Some",true]
ctc.v     > Main/fun: ["Some",true] === ["Some",true]
ctc.views > Main/fun: ["Some",true] === ["Some",true]
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > Main/fun: ["None",null] === ["None",null]
ctc.views > Main/fun: ["None",null] === ["None",null]
ctc.v     > fun: ["None",null] === ["None",null]
ctc.views > fun: ["None",null] === ["None",null]
ctc.v     > fun: ["None",null] === ["None",null]
ctc.views > fun: ["None",null] === ["None",null]
ctc.v     > fun: ["None",null] === ["None",null]
ctc.views > fun: ["None",null] === ["None",null]
ctc.v     > fun: ["Some",true] === ["Some",true]
ctc.views > fun: ["Some",true] === ["Some",true]
ctc.v     > fun: ["Some",false] === ["Some",false]
ctc.views > fun: ["Some",false] === ["Some",false]
ctc.v     > fun: ["Some",false] === ["Some",false]
ctc.views > fun: ["Some",false] === ["Some",false]
ctc.v     > fun: ["Some",false] === ["Some",false]
ctc.views > fun: ["Some",false] === ["Some",false]
ctc.v     > fun: ["Some",true] === ["Some",true]
ctc.views > fun: ["Some",true] === ["Some",true]
ctc.v     > fun: ["Some",true] === ["Some",true]
ctc.views > fun: ["Some",true] === ["Some",true]
ctc.v     > fun: ["None",null] === ["None",null]
ctc.views > fun: ["None",null] === ["None",null]
ctc.v     > fun: ["None",null] === ["None",null]
ctc.views > fun: ["None",null] === ["None",null]
ctc.v     > fun: ["None",null] === ["None",null]
ctc.views > fun: ["None",null] === ["None",null]
```

So this demonstrates what I was talking about with "subsequent phases" on Slack: I'm surprised the equality assertion would fail where it did (if it even fails at all).

It makes even less sense to me when I put `runView.fun` first and disable lines _inside_ of `runView.vbytes` instead of commenting it out entirely because that causes the assertion to fail elsewhere, but still in `runView.fun`. I also spent a good amount of time considering whether I'd messed up my consensus steps somehow but didn't make much progress on that front.

It feels like the broader takeaway is I need to crunch this down into much smaller bites, though (even if I'm just confused about how to make the above bits work)... We probably want separate contracts running under one test umbrella, eh?

jeapostrophe: First, your last comment, yes, I think it will be simpler to make a bunch of small applications. They can all be in the same file and Reach will compile them all and your single frontend (and single RPC server) will use them all.

Second, the program is too abstract for me to track what is really happening. I don't understand what the two `runView` functions do nor what the `f` function inside them does. That said, I think that you could label the `f` calls so you'd know which ones are failing. The next thing that is odd to me is that each `runView` functions starts "unsynchronized", because there is a bare `f` call without knowing what the state is, so you don't know it will be `None`.

examples/rpc-kitchen-sink/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+.PHONY: build
+build: client-lib-js build/index.main.mjs
+	docker build \
+	  --tag=$(CIMG) \
+	  --build-arg NODE_VERSION=$(NODE_VERSION) \
+	  .
+
+run:
+	$(RKEY) $(PORT) $(REACH) rpc-run docker run --rm \
```
jeapostrophe, line 36: Part of what I like about the simplicity of `rpc-run` is that you don't need to have hard-coded the keys and ports and stuff. Similarly, my assumption, which I realize now was weird, is that you'd `$(REACH) rpc-run $(REACH) run`, which is definitely strange :)

examples/crpc-common.mjs
```diff
@@ -3,14 +3,11 @@ import chalk from 'chalk';
 export const mkAssertEq = rpc => async (l, e, a) => {
   const f = x => JSON.stringify(x).replace(/\\u0000/g, '');
   const [ ej, aj ] = [ f(e), f(a) ];
-  if (ej === aj) {
+  if (ej === aj)
```
jeapostrophe, line 6: I hate `if`s and `elses` without `{}`. I have made the mistake so many times of adding an expression

jeapostrophe (approved):

PR #409, opened 2021-11-23
make diagnostic `code` use err code added in #335

hamirmahal:
On [line 89 of `hs/src/Reach/AST/Base.hs`, which Chris added in #335](https://github.com/reach-sh/reach-lang/pull/335/files#diff-27c5e2605d727a1d825bfc17e9fc740d8f4b6149ae0efa9835b327b57811c6eaR89) (thank you, Chris!), we gave what is basically Reach's language server access to an error code from Reach's compiler.

This change lets us utilize that error code, so that diagnostics can use that code from Reach's compiler instead of a generic "Compile Error" code, which is what we have currently.

Without this change:

image

With this change:

image

Also, since this change just uses the code from Reach's compiler, now, we can remove a constant that had the sole purpose of holding the `string` `'Compile Error'`.

### DIFF
--- a/vsce/server/src/server.ts
+++ b/vsce/server/src/server.ts
@@ -64,8 +64,6 @@ let hasDiagnosticRelatedInformationCapability: boolean = false;
 
 const NAME: string = 'Reach IDE';
 
-const DIAGNOSTIC_TYPE_COMPILE_ERROR: string = 'Compile Error';
-
 const DID_YOU_MEAN_PREFIX = 'Did you mean: ';
 
 let reachTempIndexFile: string;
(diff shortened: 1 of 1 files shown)


vsce/server/src/server.ts
```diff
@@ -337,7 +335,7 @@ async function validateTextDocument(textDocument: TextDocument): Promise<void> {
 						`${err.errorMessage}`,
 						'Reach compilation encountered an error.',
 						DiagnosticSeverity.Error,
-						DIAGNOSTIC_TYPE_COMPILE_ERROR,
+						err.code,
```
jeapostrophe, line 338: I'm accepting this. You should also generate a link

jeapostrophe (approved):

vsce/server/src/server.ts
hamirmahal: To do that, we need to update the dependencies, which I've already done locally, and will now push shortly.

hamirmahal: Update: #410 does this.

PR #410, opened 2021-11-23
CORE-823 update dependencies, add link to err code

hamirmahal:
To get the error code to have a link, we have to update the dependencies, which is why this change also has relatively large changes to `package.json` and `package-lock.json` files.

### DIFF
--- a/vsce/server/src/server.ts
+++ b/vsce/server/src/server.ts
@@ -363,7 +363,7 @@ async function validateTextDocument(textDocument: TextDocument): Promise<void> {
 	connection.sendDiagnostics({ uri: textDocument.uri, diagnostics });
 
 	function addDiagnostic(element: ErrorLocation, message: string, details: string, severity: DiagnosticSeverity, code: string | undefined, suggestions: string[]) {
-		const href = `https://docs.reach.sh/${code}`;
+		const href = `https://docs.reach.sh/${code}.html`;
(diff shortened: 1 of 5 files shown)


hamirmahal:
This is a screen recording I took after checking out the changes on another device, running `npm install`, and pressing `F5`.

https://user-images.githubusercontent.com/43425812/143142057-d4d5b625-82c1-4b33-8e67-62930ec3777e.mov

vsce/server/src/server.ts
```diff
@@ -356,12 +363,16 @@ async function validateTextDocument(textDocument: TextDocument): Promise<void> {
 	connection.sendDiagnostics({ uri: textDocument.uri, diagnostics });
 
 	function addDiagnostic(element: ErrorLocation, message: string, details: string, severity: DiagnosticSeverity, code: string | undefined, suggestions: string[]) {
+		const href = `https://docs.reach.sh/${code}`;
```
jeapostrophe, line 366: I don't know why you don't need `.html`?

hamirmahal: That's interesting. It appears `.html` isn't required.

hamirmahal: 5e5061c adds it just in case we need it.

jeapostrophe (approved):

PR #411, opened 2021-11-23
CORE-791: Handle Latest `language-javascript`

chrisnevers:
The new version of `language-javascript` treats lambda bodies as 
```Haskell
data ConciseBody
  | JSConciseExprBody JSExpression
  | JSConciseFunBody JSBlock
```
instead of `JSStatement`

### DIFF
--- a/hs/src/Reach/Connector/ETH_Solidity.hs
+++ b/hs/src/Reach/Connector/ETH_Solidity.hs
@@ -1202,6 +1202,8 @@ apiDef who ApiInfo{..} = do
           _ -> do
             tc_args <-
               case ai_msg_vs of
+                [DLVar _ _ (T_Tuple []) _] ->
+                  return $ [ "false" ]
                 [v] -> do
                   tc' <- solType_ $ varType v
                   return $ [ solApply tc' args ]
--- a/hs/stack.yaml
+++ b/hs/stack.yaml
@@ -32,7 +32,7 @@ extra-deps:
     commit: ca66b8dcfe22d1c1afbe81c83274dbc27115bf68
     # ^ from branch "reach-fork-2"
   - git: https://github.com/reach-sh/language-javascript.git
-    commit: 2c41d77fdcfbe6ac5f9e9d23ef9fc5b890323d24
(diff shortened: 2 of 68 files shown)


hs/stack.yaml
```diff
@@ -32,7 +32,7 @@ extra-deps:
     commit: ca66b8dcfe22d1c1afbe81c83274dbc27115bf68
     # ^ from branch "reach-fork-2"
   - git: https://github.com/reach-sh/language-javascript.git
-    commit: a9bde32d618f0f6a2098ff256001b48ff5a9a511
+    commit: bb807bd75fd45a8cce355f4be4bc413f9541e276
```
chrisnevers, line 35: We'd need to merge `language-javascript` pr first: https://github.com/reach-sh/language-javascript/pull/2

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1202,6 +1202,8 @@ apiDef who ApiInfo{..} = do
           _ -> do
             tc_args <-
               case ai_msg_vs of
+                [DLVar _ _ (T_Tuple []) _] ->
+                  return $ [ "false" ]
```
jeapostrophe, line 1206: What is going on here?

chrisnevers: Found a small issue with the lifted API call. Usually we want to construct the type of the argument like `T22(..)` when calling the consensus function, but when the domain is empty, the function expects a boolean. So, this is outputs `bool` instead of `bool()`.

jeapostrophe: You should make a note to have an example that demonstrates that problem.

hs/stack.yaml
jeapostrophe: I accidentally made a merge rather than a rebase, so the commit works but is not perfect. Sorry. I made a rule in the other repo to prevent myself from making the same mistake again.

jeapostrophe (approved):

PR #416, opened 2021-11-27
make source of compiler diagnostics more precise

hamirmahal:
When we publish compiler errors, we don't have to say "Reach IDE". We can be more precise, since we know Reach's compiler is reporting the error messages, and say "Reach Compiler".

Without this change:

image

With this change:

image


jeapostrophe (approved):
My taste is to just say "Reach"

hamirmahal:
I was going to protest, saying "Reach" is less specific than "Reach IDE", but it does seem to be the convention to have the name of the language be the `source` of a `Diagnostic`.

image

image

2ac3ae3 makes this change.

jeapostrophe (approved):

PR #422, opened 2021-12-01
[WIP] CORE-148: Add Events

chrisnevers:


hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1412,7 +1412,10 @@ ce = \case
   DLE_TimeOrder {} -> impossible "timeorder"
   DLE_GetContract _ -> code "txn" ["ApplicationID"]
   DLE_GetAddress _ -> cContractAddr
-  DLE_EmitLog at _ _ v@(DLVar _ _ _ n)  -> do
+  DLE_EmitLog _ _ _ (L_Event vs) -> do
+    clog $ map DLA_Var vs
```
chrisnevers, line 1416: @jeapostrophe  I assume this will need to be tagged with the label, similar to how we tag internal logs with the DLVar int, right? Should we make a map of labels to ints so the tag can be an int and the runtime will know to look at the mapping?

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -243,7 +243,7 @@ data SolCtxt = SolCtxt
   , ctxt_plo :: PLOpts
   , ctxt_intidx :: Counter
   , ctxt_ints :: IORef (M.Map Int Doc)
-  , ctxt_outputs :: IORef (M.Map DLVar Doc)
+  , ctxt_outputs :: IORef (M.Map String Doc)
```
chrisnevers, line 246: Use the outputEvent name which is either `oe${DLVar}` or the Event label. This prevents multiple event declarations from being created when emitting an Event multiple times in a Reach program.

jeapostrophe (changes requested):

js/stdlib/ts/shared_impl.ts
```diff
@@ -200,6 +209,28 @@ export type ISetupView<ContractInfo, VerifyResult, ConnectorTy extends AnyBacken
...
+}
+
+export type EventStream<T> = {
+  // mvp
+  seek: (t:Time) => void,
+  next: () => Promise<Event<T>>
+  // additional
+  seekNow: () => void,
+  lastTime: () => Time,
+  // why can't TS handle a function type as arg
```
jeapostrophe, line 230: You just need lots of parens

js/stdlib/ts/shared_impl.ts
```diff
@@ -200,6 +209,28 @@ export type ISetupView<ContractInfo, VerifyResult, ConnectorTy extends AnyBacken
   getView1: ((views:IBackendViewsInfo<ConnectorTy>, v:string, k:string|undefined, vi:IBackendViewInfo<ConnectorTy>, isSafe: boolean) => ViewVal)
 };
 
+export type ISetupEvent<ContractInfo, VerifyResult, ConnectorTy extends AnyBackendTy> = (args:ISetupEventArgs<ContractInfo, VerifyResult>, bin: IBackend<ConnectorTy>) => {
+  getEvent: (name: string, ty: any) => Promise<Event<any>>;
+}
+
+export type Time = number;
```
jeapostrophe, line 216: It will PROBABLY be BigNumber

examples/event/index.rsh
```diff
@@ -5,7 +5,7 @@ export const main = Reach.App(() => {
   const A = Participant('A', {
     getX: Fun([], UInt),
   });
-  const E = Event('x_event', {
+  const E = Events('x_event', {
```
jeapostrophe, line 8: FYI, you're going to have to support "unscoped" Events just like APIs and Views

js/stdlib/ts/shared_impl.ts
hamirmahal: `monitor: (fn: () => void) => void` is valid TypeScript. Depending on the function signature of `monitor`, you might want to change that.

For instance, if `monitor` takes a sum function that takes two `number` parameters `a` and `b`, and this sum function returns their sum, also a `number`, and `monitor` returns `void`, we could write that as `monitor: (fn: (a: number, b: number) => number) => void`.

hamirmahal: `monitor: (fn: () => void) => void` is valid TypeScript. Depending on the function signature of `monitor`, you might want to change that.

For instance, if `monitor` takes a sum function that takes two `number` parameters `a` and `b`, and this sum function returns their sum, also a `number`, and `monitor` returns `void`, we could write that as

```
monitor: (fn: (a: number, b: number) => number) => void
```

js/stdlib/ts/ALGO.ts
```diff
@@ -1652,15 +1656,55 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
...
-        void evt;
-        void tys;
-        return { when: 0, what: undefined };
-      }
-      return { getEvent };
+    const setupEvents = (a: SetupEventArgs) => {
+      const eventCache = new EventCache();
+      const getC = makeGetC(a, eventCache);
+      const createEventStream = (event: string, tys: AnyALGO_Ty[]) => {
+        let time: BigNumber | null = null;
```
jeapostrophe, line 1663: I think this time might be part of the same state as the eventCache (i.e. out one level of scope)

js/stdlib/ts/ALGO.ts
```diff
@@ -1652,15 +1656,55 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
...
+          // @ts-ignore
+          const parsedLog = T_Tuple([T_Bytes(4)].concat(tys)).fromNet(reNetify(log));
+          const blockTime = bigNumberify(res.txn['confirmed-round']);
+          debug(dhead + ` parsed log`, parsedLog, blockTime);
+          parsedLog.shift(); // Remove tag
+          return { when: blockTime, what: parsedLog };
+        }
+        return { seek, next };
+      };
+      return createEventStream;
```
jeapostrophe, line 1707: You should make the return an object w/ named things

js/stdlib/ts/ETH_like.ts
```diff
@@ -834,15 +851,42 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account>
...
+      let creation_block = bigNumberify(0);
+      const getC = makeGetC(setupArgs, eventCache, (cb) => {
+        creation_block = bigNumberify(cb);
+      });
+      const createEventStream = (event: string, tys: any[]) => {
         void tys;
-          return { when: 0, what: undefined };
-      }
-      return { getEvent };
+        let time = creation_block;
```
jeapostrophe, line 862: Ditto

js/stdlib/ts/shared_impl.ts
```diff
@@ -209,11 +209,13 @@ export type ISetupView<ContractInfo, VerifyResult, ConnectorTy extends AnyBacken
   getView1: ((views:IBackendViewsInfo<ConnectorTy>, v:string, k:string|undefined, vi:IBackendViewInfo<ConnectorTy>, isSafe: boolean) => ViewVal)
 };
 
-export type ISetupEvent<ContractInfo, VerifyResult, ConnectorTy extends AnyBackendTy> = (args:ISetupEventArgs<ContractInfo, VerifyResult>, bin: IBackend<ConnectorTy>) => {
-  getEvent: (name: string, ty: any) => Promise<Event<any>>;
-}
+export type ISetupEvent<ContractInfo, VerifyResult> =
+  (args:ISetupEventArgs<ContractInfo, VerifyResult>) =>
+      (event: string, tys: any[]) =>
```
jeapostrophe, line 214: This is the thing that should be an object

jeapostrophe (changes requested):

js/stdlib/ts/ALGO.ts
```diff
@@ -1671,40 +1671,73 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
...
           parsedLog.shift(); // Remove tag
           return { when: blockTime, what: parsedLog };
         }
-        return { seek, next };
+
+        const seekNow = async () => {
+          eventCache.currentRound = (await getNetworkTime()).toNumber() - 1;
+        }
+
+        const lastTime = async () => {
```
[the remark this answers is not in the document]
chrisnevers: I was concerned about it having fresh accurate data. But, maybe I can make `next()` store a state variable: `latestLog: {when, what}`. And run `next()` within this function and return the `when` field?

js/stdlib/ts/ALGO.ts
```diff
@@ -1671,40 +1671,73 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
...
+            const logs: string[] = (txn['logs'] || []).slice(logIdx);
             const good = logs.some((log) => log.startsWith(base64Hash));
             return good;
           };
-          const res = await eventCache.query(dhead, ApplicationID, { minRound: time.toNumber() }, pred);
-          if (!res.succ) {
-            throw Error(`Event not found`);
+          let res: QueryResult = { succ: false, round: 0  };
+          while (!res.succ) {
+            res = await eventCache.query(dhead, ApplicationID, { minRound: time.toNumber() }, true, pred);
```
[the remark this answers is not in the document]
chrisnevers: I'm not sure what `busy way` means

js/stdlib/ts/ALGO.ts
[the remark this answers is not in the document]
jeapostrophe: Yea. Do that

js/stdlib/ts/ALGO.ts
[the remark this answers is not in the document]
jeapostrophe: I meant "busy wait"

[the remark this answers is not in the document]
chrisnevers: This may be occurring. When I put a `wait` in the example contracts, docker uses ~200% CPU, as opposed to the normal 150% it uses. What would be a technique to avoid busy wait? Putting a sleep timer?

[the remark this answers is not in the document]
jeapostrophe: Yea, using Timeout. You know that a new block can't happen in less than like 5 seconds, so you can wait 5s

docs/scrbl/ref-model.scrbl
```diff
@@ -51,6 +51,7 @@ In addition to @tech{values}, @tech{consensus state} may contain a fixed number
 These @tech{mappings} are referred to as "@deftech{linear state}" because their size is linear in the number of @tech{participants} in the @tech{contract}.
 Furthermore, a @tech{contract} may provide @deftech{views} of its @tech{consensus state}, which are labeled functions and values which may be hierarchically organized, such as @litchar{NFT.owner} or @litchar{scoreOfPlayer}.
 These @tech{views} are visible in sub-trees of the computation graph.
+A @tech{contract} may also emit @deftech{event}s, which store the given arguments in the transaction logs of a @tech{consensus network}.
```
jeapostrophe, line 54: This is not sufficiently "model-y". For example, it shouldn't talk about "arguments" or "transaction logs" or anything like that. It should be something like ", which are externally observable values that are persistently available."

js/stdlib/ts/ETH_like.ts
```diff
@@ -882,6 +883,7 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account>
           let res: QueryResult = { succ: false, block: 0 } ;
           while (!res.succ) {
             res = await eventCache.query(dhead, getC, event, { fromBlock: time.toNumber(), isEventStream: true }, getMinBlockWithLogIndex(logIndex));
+            if (!res.succ) { sleep(5000); }
```
jeapostrophe, line 886: We use `await Timeout.set(X);` everywhere else

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #429, opened 2021-12-05
safer operations for CI

chikeabuah:


hs/src/Reach/Simulator/Core.hs
```diff
@@ -694,3 +694,8 @@ while bl cons = do
       _ <- interp cons
       while bl cons
     _ -> impossible "unexpected error"
+
+saferIndex :: Int -> [a] -> a
```
jeapostrophe, line 698: I'll accept this for now, but I'm pretty sure this is in the `Safe` module and we already use it. It's called `atMay` I think

jeapostrophe (approved):

PR #437, opened 2021-12-07
implement only; utilize lcl state; identify several todos

chikeabuah:
conditionally opens sim server port, begins to utilize local state, add initial tests


hs/src/Reach/Simulator/Core.hs
```diff
@@ -154,6 +154,9 @@ type Token = Integer
 
 data Ledger = Ledger
   { nw_ledger :: M.Map Account (M.Map Token Balance)
+  -- QUESTION: do you still want nw_next_acc here as
+  -- opposed to in `Global`?
+  -- I would rather not nest data unless for a good reason :)
```
jeapostrophe, line 159: I don't think it really matters. I can imagine `Ledger` going completely away and going into `Global`. I can also imagine some sense of abstraction purity where you have everything "about the ledger" together.

hs/src/Reach/Simulator/Core.hs
```diff
@@ -631,28 +634,63 @@ instance Interp LLStep where
...
-            Just part -> do
-              let m = saferMapRef "LLS_ToConsensus1" $ M.lookup (bpack part) tc_send
-              case m of
-                DLSend {..} -> do
+      let sends = M.mapKeys bunpack tc_send
+      v <- suspend $ PS_Suspend (A_TieBreak mid $ M.keys sends)
+      case tc_mtime of
+        Just (dltimearg, step) -> case dltimearg of
+          Left dlarg -> do
+            ev <- interp dlarg
```
jeapostrophe, line 646: I think you need some helper functions like `vUInt` that throws the "unexpected" error to keep this code nicer: `vUInt <$> interp dlarg`

hs/src/Reach/Simulator/Core.hs
```diff
@@ -631,28 +634,63 @@ instance Interp LLStep where
...
+    V_UInt n -> do
+      let locals = e_locals l
+      let lclsv = saferMapRef "LLS_ToConsensus" $ M.lookup (fromIntegral n) locals
+      let lclst = l_who $ lclsv
+      case lclst of
+        Nothing -> impossible "expected participant id, received consensus id"
+        Just part -> do
+          let dls = saferMapRef "LLS_ToConsensus1" $ M.lookup part sends
+          case dls of
+            DLSend {..} -> do
```
jeapostrophe, line 677: There's only one case, so you want `let dls@DLSend {..} = safer`

examples/simulator-1/index.rsh
```diff
@@ -1,5 +1,7 @@
 'reach 0.1';
 
+// rps-2-rps
```
jeapostrophe, line 3: Consider changing the Makefile to do `cp ../rps-2-rps/index.rsh .` as part of the `build` process

examples/simulator-2/Makefile
```diff
@@ -0,0 +1,16 @@
+REACH = ../../reach
+NODE = node
+NPM = npm
+
+.PHONY: build
+build:
+	$(NPM) install
```
jeapostrophe, line 7: Heresy!

hs/app/reach/Main.hs
```diff
@@ -896,6 +896,7 @@ compile = command "compile" $ info f d where
             -e REACH_CONNECTOR_MODE \
             -e "REACHC_ID=$${ID}" \
             -e "CI=$ci'" \
+            -p 3001:3001 \
```
jeapostrophe, line 899: This is very dangerous. For now, make it so this only happens when you do `--sim` and make a note to do it better in the future

hs/src/Reach/Compiler.hs
```diff
@@ -69,15 +69,15 @@ compile env (CompilerOpts {..}) = do
...
-          startServer ol
         let vo_out = woutnMay
         let vo_mvcs = doIf connectors dlo_verifyPerConnector
         let vo_timeout = co_verifyTimeout
         let vo_dir = dirDotReach'
         verify (VerifyOpts {..}) ol >>= maybeDie
         el <- erase_logic ol
         showp "el" el
+        unless (not co_sim) $ do
+          startServer el
```
jeapostrophe, line 80: One danger of this is that I have a future dream of integrating the simulator logic with the SMT solver so you can do something like "return a symbolic value" rather than 42 for a given interact and then get a SMT context. I don't think you should change anything, but it's something I think about

hs/src/Reach/Simulator/Core.hs
```diff
@@ -199,6 +177,7 @@ data Action
   | A_ChangeActor Int
   | A_AdvanceTime Integer
   | A_AdvanceSeconds Integer
+  | A_InteractV String String DLType
```
jeapostrophe, line 180: There should be some protection that these are always the same value. I imagined that when you created a participant, you would have to specify all of the interact values up front

hs/src/Reach/Simulator/Core.hs
```diff
@@ -245,6 +224,19 @@ addToStore x v = do
...
+  (g, _) <- getState
+  let record = reverse $ e_record g
+  return $ checkRecord' record
+  where
+    checkRecord' [] = Nothing
+    checkRecord' (a:as) = do
+      let store = t_store a
+      case M.lookup x store of
+        Nothing -> checkRecord' as
+        Just v -> Just v
```
jeapostrophe, line 238: What's going on here? What is this function? It seems like a poor-man's `Map.lookup`?

hs/src/Reach/Simulator/Core.hs
```diff
@@ -327,18 +318,30 @@ instance Interp DLArg where
...
-      case M.lookup aid locals of
-        Nothing -> possible "addToStore: no local store"
-        Just lst -> do
+      case (M.lookup aid locals, M.lookup (fromIntegral consensusId) locals) of
+        (Nothing,_) -> possible "DLA_Var: no local store"
+        (Just lst,Just conslst) -> do
           let st = l_store lst
-          return $ saferMapRef "DLA_Var" $ M.lookup dlvar st
+          let cst = l_store conslst
+          case M.lookup dlvar (M.union st cst) of
```
jeapostrophe, line 326: What is going on here?

hs/src/Reach/Simulator/Core.hs
```diff
@@ -451,7 +454,7 @@ instance Interp DLExpr where
           V_UInt n -> do
             suspend $ PS_Suspend (A_AdvanceSeconds n)
           _ -> impossible "unexpected error"
-    DLE_PartSet _at _slpart dlarg -> interp dlarg
+    DLE_PartSet _at _slpart dlarg -> interp dlarg --TODO
```
jeapostrophe, line 457: This doesn't do anything else

hs/src/Reach/Simulator/Core.hs
```diff
@@ -563,7 +565,19 @@ instance Interp DLStmt where
           addToStore switch_binding v
           interp dltail
         _ -> impossible "unexpected error"
-    DL_Only _at _either_part dltail -> interp dltail
+    DL_Only _at either_part dltail -> do
+      (g,l) <- getState
+      case either_part of
+        Left slpart -> do
```
jeapostrophe, line 571: You should only run this if you actually are this part

hs/src/Reach/Simulator/Core.hs
```diff
@@ -563,7 +565,19 @@ instance Interp DLStmt where
...
+          let pacts = e_partacts g
+          let actid = saferMapRef "DL_Only" $ M.lookup (bunpack slpart) pacts
+          let restore_id = e_curr_actorid l
+          setLocal $ l {e_curr_actorid = actid}
+          r <- interp dltail
+          l' <- getLocal
+          setLocal $ l' {e_curr_actorid = restore_id}
+          return r
+        Right _b -> interp dltail
```
jeapostrophe, line 580: You should never see a `Right`

hs/src/Reach/Simulator/Core.hs
```diff
@@ -632,27 +646,60 @@ instance Interp LLStep where
...
+              case tc_mtime of
+                Just (dltimearg, step) -> case dltimearg of
+                  Left dlarg -> do
+                    n' <- vUInt <$> interp dlarg
+                    let t = e_nwtime g
+                    f t n' step
+                  Right dlarg -> do
+                    n' <- vUInt <$> interp dlarg
+                    let t = e_nwtime g
+                    f t n' step
```
jeapostrophe, line 676: This doesn't make any sense. The only way you can run the timeout is if the time advanced. But you have this logic inside of the "tie breaking", but really there would be another action called "Run Timeout" but this could only be used (if it would only not error) if the time is past

hs/src/Reach/Simulator/Core.hs
```diff
@@ -632,27 +646,60 @@ instance Interp LLStep where
...
         _ -> impossible "unexpected client answer"
 
+consensusBody :: Integer -> Global -> Local -> DLSend -> DLRecv LLConsensus -> App DLVal
+consensusBody n g l (DLSend {..}) (DLRecv {..}) = do
+  let locals = e_locals l
+  let lclsv = saferMapRef "LLS_ToConsensus" $ M.lookup (fromIntegral n) locals
+  addToRecord (fromIntegral n) dr_time $ V_UInt (e_nwtime g)
+  addToRecord (fromIntegral n) dr_secs $ V_UInt (e_nwsecs g)
+  addToRecord (fromIntegral n) dr_didSend $ V_Bool True
+  addToRecord (fromIntegral n) dr_from $ V_Address $ l_acct lclsv
```
jeapostrophe, line 687: I don't understand what this "record" abstraction is about

hs/src/Reach/Simulator/Core.hs
```diff
@@ -632,27 +646,60 @@ instance Interp LLStep where
...
+                Nothing -> consensusBody n g l dls recv
         _ -> impossible "unexpected client answer"
 
+consensusBody :: Integer -> Global -> Local -> DLSend -> DLRecv LLConsensus -> App DLVal
+consensusBody n g l (DLSend {..}) (DLRecv {..}) = do
+  let locals = e_locals l
+  let lclsv = saferMapRef "LLS_ToConsensus" $ M.lookup (fromIntegral n) locals
+  addToRecord (fromIntegral n) dr_time $ V_UInt (e_nwtime g)
+  addToRecord (fromIntegral n) dr_secs $ V_UInt (e_nwsecs g)
+  addToRecord (fromIntegral n) dr_didSend $ V_Bool True
```
jeapostrophe, line 686: This is only true for the actual sender. It is false for everyone else

hs/src/Reach/Simulator/Core.hs
```diff
@@ -632,27 +646,60 @@ instance Interp LLStep where
...
+  addToRecord (fromIntegral n) dr_from $ V_Address $ l_acct lclsv
+  let (DLPayAmt {..}) = ds_pay
+  let restore_id = e_curr_actorid l
+  setLocal $ l {e_curr_actorid = fromIntegral n}
+  net <- interp pa_net
```
jeapostrophe, line 691: It looks like you are ignoring the token amounts

jeapostrophe (changes requested):

examples/simulator-1/client.mjs
```diff
@@ -1,88 +1,24 @@
...
-
 const main = async () => {
-
+  const port = 3001
+  console.log("Init Testing!")
   const params = {
     port: port
   };
-
   await waitPort(params)
```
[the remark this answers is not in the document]
chikeabuah: okay

examples/simulator-1/index.rsh
chikeabuah: i'm not sure yet if i will change the original programs

chikeabuah: also, not doing that makes local testing more convenient

jeapostrophe (approved):

PR #442, opened 2021-12-10
backgrounding server is actually necessary

chikeabuah:
at some point...i thought `&` was unnecessary and removed it. as it turns out, this was the worst idea i've had all year :(


jeapostrophe (dismissed):

jeapostrophe (approved):
The real solution is to make a script for `run` that starts the simulator, runs the client, then kills the simulator. Also, this is going to break if they end up being put in the same CI job, since they are all using the same port

PR #445, opened 2021-12-13
CORE-885 track VS ext usage with environment var

hamirmahal:
Because we run `reach` commands both client-side, in `vsce/client/src/extension.ts`, and server-side, in `vsce/server/src/server.ts`, we need to modify the environment variable in both of these files to properly track usage of our Visual Studio extension.

### DIFF
--- a/hs/src/Reach/Report.hs
+++ b/hs/src/Reach/Report.hs
@@ -20,6 +20,8 @@ startReport :: Maybe String -> String -> IO (Report -> IO ())
 startReport mwho i = do
   startTime <- getCurrentTime
   cm <- lookupEnv "REACH_CONNECTOR_MODE" >>= maybe (pure "") pure
+  --- check env variable to track extension usage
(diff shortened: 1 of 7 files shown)


hs/src/Reach/Report.hs
```diff
@@ -20,6 +20,8 @@ startReport :: Maybe String -> String -> IO (Report -> IO ())
 startReport mwho i = do
   startTime <- getCurrentTime
   cm <- lookupEnv "REACH_CONNECTOR_MODE" >>= maybe (pure "") pure
+  --- check env variable to track extension usage
+  vse <- lookupEnv "REACH_IDE" >>= maybe (pure "") pure
```
jeapostrophe, line 24: @mattaudesse Can you add this to the CLI in this branch?

mattaudesse: Yup - will do.

jeapostrophe (approved):

PR #454, opened 2021-12-17
CORE-904 add `QuickFix` "auto-correct" for header

hamirmahal:
https://user-images.githubusercontent.com/43425812/146607442-d61ab8b7-069b-4ea9-a9be-b8b4f5913897.mp4

### DIFF
--- a/vsce/server/src/server.ts
+++ b/vsce/server/src/server.ts
@@ -520,6 +520,26 @@ async function getCodeActions(diagnostics: Diagnostic[], textDocument: TextDocum
 			suggestions.forEach(suggestion => {
 				codeActions.push(getQuickFix(diagnostic, labelPrefix + suggestion, range, suggestion, textDocument));
 			});
+		} else if (diagnostic.code === "RE0048") {
+			const title = "Add Reach program header.";
+			const newText: string = "'reach 0.1';\n\n";
+			const textEdit: TextEdit = {
+				newText,
+				range
+			};
+			const changes = {
+				[ textDocument.uri ]: [ textEdit ]
+			};
+			const edit: WorkspaceEdit = {
+				changes
+			};
+			const codeAction: CodeAction = {
+				edit,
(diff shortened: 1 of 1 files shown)


jeapostrophe (approved):

vsce/server/src/server.ts
```diff
@@ -520,6 +520,26 @@ async function getCodeActions(diagnostics: Diagnostic[], textDocument: TextDocum
 			suggestions.forEach(suggestion => {
 				codeActions.push(getQuickFix(diagnostic, labelPrefix + suggestion, range, suggestion, textDocument));
 			});
+		} else if (diagnostic.code === "RE0048") {
+			const title = "Add Reach program header.";
+			const newText: string = "'reach 0.1';\n\n";
```
jeapostrophe, line 525: Ideally it will get 0.1 from the `VERSION` file

hamirmahal: It would only be able to do that in development, and not production, so maybe we keep some `version.ts` file in `vsce/server/src`?

hamirmahal: Or, we can maybe grab it from the error message, since the error message gives a suggested header, anyway.

Screen Shot 2021-12-19 at 5 03 30 PM

hamirmahal: Actually, after thinking about it a bit more, I think it really does make sense for the "Quick Fix", or, "auto-correct", to match whatever header Reach's compiler suggests in its error message.

We want the "auto-correct" to fix the compiler error: it doesn't matter what `../../VERSION` just so happened to be in the repository at the time we published the extension.

I am assuming that the compiler error message will have the correct Reach header.

jeapostrophe: Good idea

hamirmahal: Thanks.

hamirmahal: Thanks.

PR #455, opened 2021-12-18
implement and test actor threads

chikeabuah:


hs/src/Reach/Simulator/Core.hs
```diff
@@ -660,70 +648,127 @@ instance Interp LLStep where
...
       (g, l) <- getState
-      let mid = e_nmsgid g
-      incrMessageId
+      let actId = l_curr_actor_id l
+      phId <- getPhaseId actId
+      msgs' <- case M.lookup phId $ e_messages g of
+            Just m -> return m
+            Nothing -> do
+              let msg = NotFixedYet mempty
+              setGlobal g { e_messages = M.insert phId msg $ e_messages g }
```
jeapostrophe, line 659: This is an ugly way to initialize for my taste. I think a `fromMaybe` that returns the default is more elegant

hs/src/Reach/Simulator/Core.hs
```diff
@@ -660,70 +648,127 @@ instance Interp LLStep where
...
+                      interp $ dr_k
+                    Fixed (actId',_msg) -> do
+                      winner dlr actId' phId
+                      interp $ dr_k
+                Just (DLSend {..}) -> do
+                  case msgs' of
+                    NotFixedYet msgs'' -> do
+                      ds_msg' <- mapM interp ds_msg
+                      let sto = M.fromList $ zip dr_msg ds_msg'
+                      setGlobal g { e_messages = M.insert phId (NotFixedYet $ M.insert actId sto msgs'') (e_messages g) }
```
jeapostrophe, line 683: Because you're just going to be setting it down hear anyways

hs/src/Reach/Simulator/Core.hs
```diff
@@ -660,70 +648,127 @@ instance Interp LLStep where
...
+                      _ <- suspend $ PS_Suspend (A_Contest phId)
+                      winner dlr actId phId
+                      interp $ dr_k
```
jeapostrophe, line 686: This pattern is repeated four times in this code

hs/src/Reach/Simulator/Core.hs
chikeabuah: fair

hs/src/Reach/Simulator/Core.hs
```diff
@@ -504,10 +504,9 @@ instance Interp DLExpr where
       let new_nw_ledger = M.insert 0 (M.delete ev m) map_ledger
       setGlobal $ e {e_ledger = new_nw_ledger }
       return V_Null
-    -- TODO: new stuff
     DLE_TimeOrder _at _assoc_maybe_arg_vars -> return V_Null
-    DLE_GetContract _at -> impossible "undefined"
-    DLE_GetAddress _at -> impossible "undefined"
+    DLE_GetContract _at -> V_Address <$> l_acct <$> getMyLocalInfo
```
jeapostrophe, line 508: Contract are not necessarily addresses and it might be a good idea to introduce a new value, `V_Contract`, for them in the future to ensure that they can't be mixed at run time

jeapostrophe (approved):

hs/src/Reach/Simulator/Core.hs
chikeabuah: noted

jeapostrophe (approved):

PR #457, opened 2021-12-20
Api nft dump

CharaD7:
Implemented API calls for nft-dumb.

I could not successfully run it on my local system. Kindly test and let me know there are any mishaps.

### DIFF
--- a/examples/nft-dumb/index.rsh
+++ b/examples/nft-dumb/index.rsh
@@ -7,9 +7,9 @@ export const main =
     const Creator = Participant('Creator', {
         getId: Fun([], UInt) });
 
-    const Owner = ParticipantClass('Owner', {
-        newOwner: Fun([], Address),
-        showOwner: Fun([UInt, Address], Null) });
+    const Owner = API('Owner', {
+        newOwner: Fun([Address], Null),
+        showOwner: Fun([], Tuple(UInt, Address)) });
 
     const vNFT = View('NFT', {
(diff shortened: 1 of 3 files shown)


examples/nft-dumb/index.rsh
```diff
@@ -26,28 +26,20 @@ export const main =
     while ( true ) {
       commit();
 
-      Owner.only(() => {
```
MrJackdaw, line 29: This doesn't quite capture what the original contract does. If we follow the old logic:

```
NFT Creator deploys contract
NFT Creator passes id of NFT into contract

[ =    set variable to track new owner    = ]
[ =    start while loop                               = ]

NFT Creator joins contract as Owner ParticipantClass

Owner (NFT Creator) publishes new owner
  FIRST LOOP:
     NFT Creator will specify the address for the new Owner
  ALL OTHER TIMES:
     Owner will be the ParticipantClass member that won the race to join

Contract sets new Owner of NFT to variable

[ =    end while loop                               = ]
```

We can jam on this after all the meetingz™ if you like.

examples/nft-dumb/reach
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
```
jeapostrophe, line 1: Copies of this should never be committed

jeapostrophe:
I'm going to close this. Feel free to continue and resubmit.

PR #467, opened 2021-12-22
Fixing parseCurrency and formatCurrency on Algorand

keeganthomp:
## Summary

`parseCurrency` and `formatCurrency` are currently not working well with tokens/assets that have any decimal value other than `6` on the Algorand network. This PR resolves that issue.

`formatCurrency` is currently taking a `decimal` argument, but the returned value was still getting split at 6 decimals.
I.e.  If I have 1 TokenA with 0 decimals, it would be shown as `.000001`. this PR resolves that

`parseCurrency` was not taking any decimal argument which is causing it to not parse amounts correctly that have varying decimal values.
I.e. if I pass 1 of TokenA with 0 decimals to `parseCurrency`, it would parse it as `1000000` as every value was getting converted to microalgos (`10 ** 6`)

#### Testing
I made a simple CLI program locally that parsed and formatted various values and decimals by calling `formatCurrency` and `parseCurrency` and was checking the expected output. I also implemented a version of this in Humbleswap UI that has been tested by me and team as well.

#### Docs/Changelog
Will need to update docs as `parseCurrency` and `formatCurrency` are now standard to/from atomic unit of the token instead of the network on Algorand


DanBurton (changes requested):
I am a strong -1 on changing the current implementation in a way that breaks backwards compatibility. The current function has two args, and was never intended to work with anything other than the standard unit:

* amt
* (optional) how many decimals to display

If the functionality in this PR is desired then a new function should be created that has 3 arguments:
* amt
* how many decimals are in the token/currency
* (optional) how many decimals to display

keeganthomp:
>If the functionality in this PR is desired then a new function should be created that has 3 arguments:

Are you suggesting a new, additional, function for both `formatCurrency` and `parseCurrency` (I.e. `formatFromDecimals` and `parseFromDecimals`)?

DanBurton:
I worry less about `parseCurrency` because the change to that function is backwards compatible.

You _could_ change `formatCurrency` by making the _third_ argument the number of decimals it has, but that argument order would probably feel wrong to people and lead to mistakes.

keeganthomp:
ok cool. that makes sense to me. I'll just make a new `formatWithDecimals` function due to your explanation.

js/stdlib/ts/ALGO.ts
```diff
@@ -1898,32 +1903,47 @@ function ldrop(str: string, char: string) {
...
  * @param decimals  up to how many decimal places to display in the {@link standardUnit}.
  *   Trailing zeros will be omitted. Excess decimal places will be truncated (not rounded).
  *   This argument defaults to maximum precision.
- * @returns  a string representation of that amount in the {@link standardUnit} for that network.
+ * @returns  a string representation of that amount in the {@link standardUnit} for that network or token.
  * @example  formatCurrency(bigNumberify('100000000')); // => '100'
  * @example  formatCurrency(bigNumberify('9999998799987000')); // => '9999998799.987'
  */
-export function formatCurrency(amt: any, decimals: number = 6): string {
+function handleFormat(amt: any, decimals: number, isNetwork: boolean): string {
```
jeapostrophe, line 1915: My taste is to take the `splitValue` as an argument explicitly; do you, @keeganthomp , want to write the docs for this too?

keeganthomp: no problem, I can update this to reflect that.

And for the docs, I sure can. how do I go about doing that?

jeapostrophe: Edit this file around here --- https://github.com/reach-sh/reach-lang/blob/master/docs/src/frontend/index.md?plain=1#L1032-L1052

jeapostrophe (approved):

PR #469, opened 2021-12-23
cypress tests

DanBurton:
currently WIP; docker not working yet

The meat of it is `js/browser-tests/cypress/integration/spec.js`.

If you have `yarn` installed locally (tangentially, trying yarn instead of npm):

```
make js-install
make js-run      # run headless tests
make js-ui       # run tests interactively
```

`rps-9-web` must be running on `localhost:3000` using `patch index.js cypress.patch`.
(`REACH_CONNECTOR_MODE=ALGO` recommended since fallback wallet is not implemented on other connectors.)


js/browser-tests/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
jeapostrophe, line 1: Since this is empty do we need it?

js/browser-tests/cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
...
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
+ */
+// eslint-disable-next-line no-unused-vars
+module.exports = (on, config) => {
```
jeapostrophe, line 19: Since this is empty do we need it?

jeapostrophe (approved):

DanBurton:
We don't need the empty files but cypress re-creates them every time it runs. Committing them seems better than gitignoring them to avoid confusion later if we decide to actually put something in there in the future.

PR #470, opened 2021-12-23
CORE-924: Add `verifyMuldiv`

chrisnevers:
Add `verifyMuldiv` that is a context-sensitive claim that the result of `muldiv(x, y, z)` will not overflow. This expression will either add an  `assume` in local steps, `require` in consensus steps, or `assert` otherwise. 

In short, I think this is the easiest implementation (bundling the assertion within the expression) because otherwise EraseLogic would need to erase assume/require depending on whether they contain a `verifyMuldiv`. That info is not easily discoverable in EraseLogic, which iterates backwards through the computation.

### DIFF
--- a/docs/src/changelog/index.md
+++ b/docs/src/changelog/index.md
@@ -7,6 +7,7 @@ Versions and changes-within-versions are listed in reverse-chronological order: 
 
 Version 0.1.7 is the current Reach release candidate version.
 
++ 2021/12/23: Added `{!rsh} verifyMuldiv`.
 + 2021/12/23: `{!rsh} deploy` was renamed to `{!rsh} init`.
 + 2021/12/21: The backend interface to deployed contracts was updated, so old contracts will not work with this version.
 + 2021/12/21: The backend interface to the compiled objects was updated, so you'll need to recompile for this release.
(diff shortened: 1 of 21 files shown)


jeapostrophe (changes requested):
First, you need docs. Second, I dislike the name, I feel like it is vague what is being verified... `checkMuldivDoesntOverflow` is really bad. I also don't like the use "verify" in this context... it is more like "check", but `checkMuldiv` doesn't make it clear that there's an `assert` behind the scenes... so maybe `verifyMuldiv` is the best option.

chrisnevers:
@jeapostrophe Yeah, I'll do that now. I was holding off on docs until I confirmed that this approach looked good. Regarding the name, I'll try to think of a more fitting, concise one.

jeapostrophe:
I agree with you that it is ugly, not horribly so, but intelligently expedient

chrisnevers:
Added docs but I'm unable to think of a better name that's not a mouthful 😕

jeapostrophe (approved):

jeapostrophe:
Whoops, I snipped you!

jeapostrophe (approved):

PR #480, opened 2022-01-03
Have the Web UI record its sequence of interactions into a serializable JSON object that can be saved and restored

chikeabuah:


examples/programmer-art/src/index.mjs
```diff
@@ -45,12 +45,14 @@ const redraw = async () => {
 }
 
 const log = document.querySelector("#output")
+var jsonLog = []
```
jeapostrophe, line 48: You should use `let`

jeapostrophe: Although, I don't think you actually mutate it

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #485, opened 2022-01-04
restructure layout and display src code with prism

chikeabuah:


examples/programmer-art/src/index.mjs
```diff
@@ -1,10 +1,52 @@
 import * as c from '@reach-sh/simulator-client';
 import cytoscape from 'cytoscape';
 import klay from 'cytoscape-klay';
+import Prism from 'prismjs';
```
jeapostrophe, line 4: You gotta use Shiki, that what we use in the docs --- https://github.com/reach-sh/reach-lang/tree/master/docs/dev

chikeabuah: i would rather not use this. this library seems to be in a terribly broken state. just trying to run the basic demo i get half a dozen errors including these issues that have been open for months : 

- https://github.com/shikijs/shiki/issues/216
- https://github.com/shikijs/shiki/issues/246

jeapostrophe: We really can't maintain multiple different formatting engines for Reach though. I'll let this slide, but make a note to revisit.

jeapostrophe (approved):

PR #488, opened 2022-01-05
implement remote object call

chikeabuah:


hs/src/Reach/Simulator/Core.hs
```diff
@@ -209,6 +209,7 @@ data Action
   | A_InteractV String String DLType
   | A_Interact SrcLoc [SLCtxtFrame] String String DLType [DLVal]
   | A_Contest PhaseId
+  | A_Remote SrcLoc [SLCtxtFrame] String [DLVal] [DLVal]
```
jeapostrophe, line 212: This is like Interact and needs to know the type that should be returned

jeapostrophe (approved):

PR #496, opened 2022-01-07
Actually run cypress tests in circle

DanBurton:


DanBurton:
Check out the video artifact:
https://app.circleci.com/pipelines/github/reach-sh/reach-lang/5101/workflows/ec2da89b-b97c-4ea9-b8ba-576d8a02c325/jobs/98329/artifacts

At this point it's clearly doing something and at least getting part of the way through. I think it's worth merging now, though there may be further improvements to do since the video doesn't show it running to completion.

Is circle just cutting off our ability to view the video arbitrarily? Or is it really not running to completion? Not sure.

jeapostrophe (approved):
I agree that it is cool and should be merged. There's lots I don't like though... I don't like that it doesn't finish and you don't know why. I assume it does when you do it locally? I don't like the `sleep 60`. You should do some grepping or polling of the port.

PR #497, opened 2022-01-07
Z3 version

vinmazzi:
## Summary
Make Z3_VERSION Argument work during the dockerbuild.


jeapostrophe (approved):
My guess is that the ARG has to be after the FROM?

jeapostrophe:
Oh, it conflicts.

jeapostrophe:
I believe this is actually merged

PR #504, opened 2022-01-12
docs: update Set docs

TheChronicMonster:
Updating Set documentation based on [Discord conversation](https://discord.com/channels/628402598663290882/749639931399241792/930533886721687592).

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -562,12 +562,12 @@ bidders.member(Alice); // false
 ```
 
 
-A `{!rsh} Set` is another container for linear state. It is simply a type alias of `{!rsh} Map(Null)`;
-it is only useful for tracking `{!rsh} Address`es. Because a `{!rsh} Set` is internally a `{!rsh} Map`, it may
-only be constructed in a consensus step.
+A `{!rsh} Set` is another container for linear state. It is simply a type alias of `{!rsh} Map(Null)`; it is only useful for tracking `{!rsh} Address`es. Because a `{!rsh} Set` is internally a `{!rsh} Map`, it may only be constructed in a consensus step.
 
-A `{!rsh} Set` may be modified by writing `{!rsh} s.insert(ADDRESS)` to install `{!rsh} ADDRESS` in the
-set, `{!rsh} s`, or `{!rsh} s.remove(ADDRESS)` to remove the `{!rsh} ADDRESS` from the set.
-Such modifications may only occur in a consensus step.
+As a use case example, a `{!rsh} Set` can whitelist an unbounded amount of addresses after the contract exits. Alternatively, an [array](https://docs.reach.sh/rsh/compute/#ref-programs-arrays) stores a maximum of 256 addresses (roughly 8k worth of data).
+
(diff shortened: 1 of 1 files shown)


docs/src/rsh/consensus/index.md
```diff
@@ -562,12 +562,12 @@ bidders.member(Alice); // false
 ```
 
 
-A `{!rsh} Set` is another container for linear state. It is simply a type alias of `{!rsh} Map(Null)`;
-it is only useful for tracking `{!rsh} Address`es. Because a `{!rsh} Set` is internally a `{!rsh} Map`, it may
-only be constructed in a consensus step.
+A `{!rsh} Set` is another container for linear state. It is simply a type alias of `{!rsh} Map(Null)`; it is only useful for tracking `{!rsh} Address`es. Because a `{!rsh} Set` is internally a `{!rsh} Map`, it may only be constructed in a consensus step.
```
jeapostrophe, line 565: What is changed on this line? One thing that I aspire to in the docs is semantic linefeeds --- https://rhodesmill.org/brandon/2012/one-sentence-per-line/ --- and this change makes it worse

docs/src/rsh/consensus/index.md
```diff
@@ -562,12 +562,12 @@ bidders.member(Alice); // false
...
 
-A `{!rsh} Set` may be modified by writing `{!rsh} s.insert(ADDRESS)` to install `{!rsh} ADDRESS` in the
-set, `{!rsh} s`, or `{!rsh} s.remove(ADDRESS)` to remove the `{!rsh} ADDRESS` from the set.
-Such modifications may only occur in a consensus step.
+As a use case example, a `{!rsh} Set` can whitelist an unbounded amount of addresses after the contract exits. Alternatively, an [array](https://docs.reach.sh/rsh/compute/#ref-programs-arrays) stores a maximum of 256 addresses (roughly 8k worth of data).
```
jeapostrophe, line 567: This is not true in general and thus not appropriate to talk about here. That limitation is only present on Algorand, not on Ethereum. Also, you should never link to a section like that. You should use `##ref-programs-arrays`

docs/src/rsh/consensus/index.md
```diff
@@ -562,12 +562,12 @@ bidders.member(Alice); // false
...
+
+One consideration when using `{!rsh} Set` is that address owners pay for storage. Participants are given the opportunity to opt-in or decline before payments are extracted.
```
jeapostrophe, line 569: This only true on Algorand, and should not be mentioned here.

docs/src/rsh/consensus/index.md
```diff
@@ -562,12 +562,12 @@ bidders.member(Alice); // false
...
+
+A `{!rsh} Set` may be modified by writing `{!rsh} s.insert(ADDRESS)` to install `{!rsh} ADDRESS` in the set, `{!rsh} s`, or `{!rsh} s.remove(ADDRESS)` to remove the `{!rsh} ADDRESS` from the set. Such modifications may only occur in a consensus step.
```
jeapostrophe, line 571: This is worse of semantic linefeeds.

PR #519, opened 2022-01-20
add min balance func

temptemp3:
## Summary

In ALGO, a portion of accounts is made not available based on account usage of apps and storage on the network. The `stdlib.minBalance(acc)` function computes amount that an account may use in a future transaction.

Min balance calculations use the following as a reference:
https://developer.algorand.org/docs/get-details/dapps/smart-contracts/apps/#minimum-balance-requirement-for-a-smart-contract

Function is accompanied by an `minBalance` in `examples`.

Yes, I believe it should be added to the change long and docs as a ALGO specific feature.

### DIFF
--- a/js/stdlib/ts/ALGO.ts
+++ b/js/stdlib/ts/ALGO.ts
@@ -172,10 +172,16 @@ type AppState = {
   'id': number,
   'key-value': AppStateKVs,
 };
+type AppSchema = {
+  "num-byte-slice": number,
+  "num-uint": number
+}
 type AccountInfo = {
-  'assets'?: Array<AccountAssetInfo>,
+  'assets': Array<AccountAssetInfo>,
   'amount': number,
-  'apps-local-state'?: Array<AppState>
+  'apps-local-state': Array<AppState>
+  'apps-total-schema': AppSchema,
+  'created-apps': Array<AppInfo>
 };
 type IndexerAccountInfoRes = {
   'current-round': number,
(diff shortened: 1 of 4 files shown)


js/stdlib/ts/ALGO.ts
```diff
@@ -171,10 +171,16 @@ type AppState = {
...
 };
+type AppSchema = {
+  "num-byte-slice": number,
+  "num-uint": number
+}
 type AccountInfo = {
-  'assets'?: Array<AccountAssetInfo>,
+  'assets': Array<AccountAssetInfo>,
   'amount': number,
-  'apps-local-state'?: Array<AppState>
```
jeapostrophe, line 177: Ditto

js/stdlib/ts/ALGO.ts
```diff
@@ -171,10 +171,16 @@ type AppState = {
...
+  'apps-local-state': Array<AppState>
+  'apps-total-schema': AppSchema,
+  'created-apps': Array<AppInfo>
```
jeapostrophe, line 183: These are all `?`

js/stdlib/ts/ALGO.ts
```diff
@@ -1748,6 +1754,22 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
   return stdAccount({ networkAccount, getAddress: selfAddress, stdlib, setDebugLabel, tokenAccepted, tokenAccept, tokenMetadata, contract });
 };
 
+export const minBalance = async (acc: Account): Promise<BigNumber> => {
+  const addr = extractAddr(acc);
+  const ai = await getAccountInfo(addr);
+  let createdApps = ai['created-apps']
```
jeapostrophe, line 1760: You should use `const` because you don't mutate them

js/stdlib/ts/ALGO.ts
```diff
@@ -1764,6 +1786,8 @@ const balanceOfM = async (acc: Account, token: Token|false = false): Promise<Big
   }
 };
 
+
+
```
jeapostrophe, line 1790: We don't want extra spaces

jeapostrophe (changes requested):
If you don't want to make these changes, just say so and I will

js/stdlib/ts/ALGO.ts
```diff
@@ -171,10 +171,16 @@ type AppState = {
   'id': number,
   'key-value': AppStateKVs,
 };
+type AppSchema = {
+  "num-byte-slice": number,
+  "num-uint": number
+}
 type AccountInfo = {
-  'assets'?: Array<AccountAssetInfo>,
```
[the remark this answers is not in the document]
temptemp3: You're right. I'll fix these.

temptemp3:
@jeapostrophe Made some changes. Hope it works. Thx!

AwolDes:
Excited for this, great work @temptemp3

jeapostrophe:
I went to merge this today, but because you are doing merges rather than rebasing, I can't. Can you either rebase or squash and update the PR? If you don't want to do either, just tell me and I'll manually make the same changes.

temptemp3:
@jeapostrophe let's not try to merge this just yet. Fixing an issue discovered earlier today. About the merges, you're right. I'll try to stay away from resolve conflicts in github!

temptemp3:
@jeapostrophe it took some time but I think I got it ; )

jeapostrophe:
Merged, thanks!

PR #533, opened 2022-01-25
Merge Tool files into a single page

Craig-Elre:
Merge Usage and Installation into a single page, but retain the links.


docs/src/tool/index.md
```diff
@@ -3,4 +3,363 @@
...
+
+Tip: try entering your desired year in the filter box to skip other tag types.
++ The identifer "`stable`".
+
+Reach will interpret this to mean the most recent stable [major](##guide-versions) version.
+
+When using the semantic versioning form of `REACH_VERSION` the preceding `v` character is optional.
+In other words, `v0.1.6` is equivalent to `0.1.6`.
+
+## {#ref-usage-compile} `reach compile`
```
jeapostrophe, line 69: Minor thing, I think it will be nice if these are `#`, not `##`

docs/src/tool/index.md
```diff
@@ -3,4 +3,363 @@
...
+
+in your project repository.
+You can copy this file to other repositories or move it to a directory in your `PATH`, like `~/bin`.
+(`PATH` is a UNIX environment variable listing each of the directories that contain programs you can run in a shell session.)
+
+:::note
+If you're using Windows, consult [the guide to using Reach on Windows](##guide-windows).
+:::
+
+# {#ref-usage} Usage
```
jeapostrophe, line 33: I feel like the name "Usage" is bad now, especially, because this is actually about the version. I think this should be something like
```
# {#ref-usage} Selecting a version
```
or something like that

docs/src/tool/index.md
```diff
@@ -3,4 +3,363 @@
 This document describes the `reach` tool: how to install it and how to use it.
 
 + @{seclink("ref-install")}
-+ @{seclink("ref-usage")}
\ No newline at end of file
++ @{seclink("ref-usage")}
```
jeapostrophe: I don't think these look good anymore and can just be removed

jeapostrophe (changes requested):

docs/src/tool/index.md
Craig-Elre: I thought about that, but then thought it might be handy to have for the mobile version. I was looking at it on my phone and there is no way to access the left-side navigation, and scrolling down can be a bit tedious. I can remove them, though. It might be interesting to check web stats every so often to see how people are accessing the documentation for insight into design. What are your thoughts?

docs/src/tool/index.md
Craig-Elre: There might come a time when we want to reuse the reference link "usage", I think now would be a good time to modify the ref as well, if you don't object.

jeapostrophe: The name `#ref-usage` has to stay because it is used to generate backwards compatible links with the old site

docs/src/tool/index.md
jeapostrophe: I trust your judgment. However, you can get the navigation... there should be hamburger at the top of the page that you can click on and get

jeapostrophe (approved):

PR #535, opened 2022-01-26
Doc hyperlink

willghatch:
## Summary

This adds cross reference hyperlinking to the documentation.

## Some questions:

* It's been... a while since I've done anything serious in Javascript.  What is our preferred method of returning maybe something?  In this I returned an `undefined`, but I am doubtful that it's the best option.  Maybe `false` would be good, but `false == 0` and similar punning makes me less confident using false in JS than I would in scheme.  Any thoughts?
* I didn't see anything in the onboarding docs about how I should go about submitting PRs.  Is this good?

### DIFF
--- a/docs/dev/generator.js
+++ b/docs/dev/generator.js
@@ -53,10 +53,43 @@ const fail = (...args) => {
   if ( forReal ) { hasError = true; }
 };
 
+
+const languages = [
+  ...shiki.BUNDLED_LANGUAGES.map((l) => {
+    if ( l.id === 'javascript' ) {
+      l.aliases.push('mjs');
+    }
+    return l;
+  }),
+  {
+    id: 'reach',
+    scopeName: 'source.js',
+    // XXX customize this
+    path: 'languages/javascript.tmLanguage.json',
+    aliases: ['rsh'],
+  },
+];
+const languageNormalizeDict = {};
+for (const l of languages) {
+  languageNormalizeDict[l.id] = l.id;
+  for (const alias of (l.aliases || [])) {
+    languageNormalizeDict[alias] = l.id;
+  }
+}
+const documentScopes = ['h', 'term'];
+const normalizeScope = (s) => {
+  const ns = (documentScopes.includes(s) && s) || languageNormalizeDict[s];
+  if (!ns) {
+    fail(`Bad xref scope:`, s);
+  }
+  return ns;
+};
+
 const xrefs = {};
 const xrefPut = (s, t, v) => {
-  if ( ! xrefs[s] ) { xrefs[s] = {}; }
-  const e = xrefs[s][t];
+  const ns = normalizeScope(s);
+  if ( ! xrefs[ns] ) { xrefs[ns] = {}; }
+  const e = xrefs[ns][t];
   if ( e !== undefined ) {
     const es = JSON.stringify(e);
     const vs = JSON.stringify(v);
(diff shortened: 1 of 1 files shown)


docs/dev/generator.js
```diff
@@ -536,6 +541,17 @@ const processMd = async ({baseConfig, relDir, in_folder, iPath, oPath}) => {
   theader.remove();
 
   // Process code snippets.
+  const linkifySpan = (s, language) => {
+    const text = s.textContent
+    const ref = xrefGetMaybe(language == "reach" ? "rsh" : language, text);
```
jeapostrophe, line 546: You should make a normalization dictionary and error if there's no normalization and/or consult the `aliases` field of Shiki's bundled languages (see the `sher` defn)

docs/dev/generator.js
```diff
@@ -653,6 +672,9 @@ const processMd = async ({baseConfig, relDir, in_folder, iPath, oPath}) => {
     const s = doc.createElement('span');
     s.classList.add('snip');
     s.appendChild(mkEl(hc));
+    for (const is of s.querySelectorAll('span')) {
```
jeapostrophe, line 675: Shame! No copying code!

jeapostrophe (changes requested):
`undefined` is great.

Does it look nice? Can you include a screenshot? I'm surprised you don't need to change the CSS to not get underlines and boldness

jeapostrophe:
This PR is fine. One thing though is that the CI will run if you commit to a branch on `reach-sh:master` rather than your own fork

willghatch:
So in that screenshot, the upper code snippet in Reach has clickable `const`, `export`, `App`, `Participant`, etc, and the `interact` inline code snippet between the two blocks is clickable.  However, the lower code block is in Javascript, and none of the keywords have defined refs, so nothing in that block is clickable.

jeapostrophe:
That's cool, I like it.

docs/dev/generator.js
```diff
@@ -536,6 +552,22 @@ const processMd = async ({baseConfig, relDir, in_folder, iPath, oPath}) => {
   theader.remove();
 
   // Process code snippets.
+  const linkifySpans = (elem, language) => {
+    const linkifySpan = (s, language) => {
+      const text = s.textContent
+      const ref = xrefGetMaybe(language == "reach" ? "rsh" : language, text);
```
jeapostrophe, line 558: I don't think this `?:` is needed because of the normalization

docs/dev/generator.js
```diff
@@ -53,21 +53,51 @@ const fail = (...args) => {
...
+  },
+];
+const languageNormalizeDict = {};
+for (const l of languages) {
+  languageNormalizeDict[l.id] = l.id;
+  for (const alias of (l.aliases || [])) {
+    languageNormalizeDict[alias] = l.id;
+  }
+}
+const normalizeScope = (s) => languageNormalizeDict[s] || s;
```
jeapostrophe, line 79: I think you should error if there's no normalization and we'd fix it

willghatch:
After figuring out `scrot`'s options, here is a screenshot showing my cursor hovering over a link:

doc-hyperlink-screenshot-2

jeapostrophe (changes requested):
Does any JS get linked? What about go and python? (See the RPC pages)

docs/dev/generator.js
willghatch: Oh, yeah.

docs/dev/generator.js
willghatch: So one thing that that allows through is that a bunch of things are added with a literal 'h' scope.  Also with literal 'term' scope.  These are not languages that Shiki knows about.  So this cross reference stuff is used for both code cross references and document cross references.

willghatch:
> Does any JS get linked? What about go and python? (See the RPC pages)

Yes, though there aren't many JS, go, or python references defined.  The RPC pages each have a couple links.

docs/dev/generator.js
willghatch: Shall we add a list of literals like 'h' and 'term' to the normalizer, and error if it's not a language name in the shiki list or a known document scope literal?

jeapostrophe: I see. Yea, I think you should add them.

willghatch:
Ok, it now fails for scopes that can't be normalized (by being document scopes or in the language list).

jeapostrophe (approved):

PR #539, opened 2022-01-27
doc cross references - allow dependent identifiers like fork.case

willghatch:
## Summary

Here is a screenshot:

doc-hyperlink-fork case

I added a reference anchor for `fork.case` to add the demo, but otherwise I'm not sure which identifiers we want this with, so currently this only does anything with the one identifier.  But when more are added they will work, and it works for arbitrary depth of `foo.bar.baz.quux` nesting.  But note that it's not very accurate -- if a code snippet includes `fork` and then later includes a `case` that's not really related to fork, it will mis-link the `case` identifier to `fork.case`.

### DIFF
--- a/docs/dev/generator.js
+++ b/docs/dev/generator.js
@@ -86,10 +86,11 @@ const normalizeScope = (s) => {
 };
 
 const xrefs = {};
+const freshXrefScopeDict = () => ({"xrefs": {}, "prefixes": {}});
 const xrefPut = (s, t, v) => {
   const ns = normalizeScope(s);
-  if ( ! xrefs[ns] ) { xrefs[ns] = {}; }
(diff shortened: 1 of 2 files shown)


jeapostrophe (approved):
OOoooooh, this is awesome.

Now, how painful would it be for you to go through the `docs/src/rsh` directory and at all the things like this? It will give you the "opportunity" ;) to learn more about Reach by reading the docs closely?

willghatch:
There is a bootcamp for Reach that starts tomorrow that I asked to join, in which I imagine I'll do some exercises to start learning Reach.  How about I try to find and add all of these as I'm going through the documentation anyway as I'm learning reach?

jeapostrophe:
Sounds great to me! <3

PR #540, opened 2022-01-27
Squelch unused binding warning of `many` in APICut.hs

mattaudesse:


hs/src/Reach/APICut.hs
```diff
@@ -312,8 +312,8 @@ slurp = \case
 many_ :: [Maybe a] -> Maybe a
 many_ = getFirst . mconcat . map First
 
-many :: (a -> App (Maybe a)) -> [a] -> App (Maybe a)
-many f l = many_ <$> mapM f l
+_many :: (a -> App (Maybe a)) -> [a] -> App (Maybe a)
+_many f l = many_ <$> mapM f l
```
hamirmahal, line 316: The `_` just tells whatever's complaining about the error that it's a readonly variable/it's okay if the variable is unused, correct?

jeapostrophe: Yea, it's a signal to Haskell that we're ignoring it on purpose

hamirmahal: Ah, I see. I believe we can send a similar signal to TypeScript in the same way. Thanks for clarifying.

PR #541, opened 2022-01-27
add "Quick Fix" to autocorrect `let` to `var`

hamirmahal:
https://user-images.githubusercontent.com/43425812/151452154-18570ca8-ea10-4c20-bb58-c6a0cf6dbc0a.mov

### DIFF
--- a/vsce/server/src/server.ts
+++ b/vsce/server/src/server.ts
@@ -562,6 +562,46 @@ async function getCodeActions(diagnostics: Diagnostic[], textDocument: TextDocum
 				title
 			};
 			codeActions.push(codeAction);
+		} else if (diagnostic.code === "RE0002") {
+			// Grab the general starting area of
+			// the diagnostic.
+			const { start } = range;
+
+			// Then, start analyzing text from
+			// the *beginning* of the line
+			// containing the Diagnostic, since
+			// the Diagnostic may start in the
+			// middle of a declaration, i.e., it
+			// may not contain "let", especially
+			// if indented, like "	let ...".
+			start.character = 0;
+
+			const text = textDocument.getText(range);
+
+			// This edit does not handle all possible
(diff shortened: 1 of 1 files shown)


jeapostrophe (approved):

jeapostrophe:
Cooool!

vsce/server/src/server.ts
```diff
@@ -562,6 +562,46 @@ async function getCodeActions(diagnostics: Diagnostic[], textDocument: TextDocum
...
+			// if indented, like "	let ...".
+			start.character = 0;
+
+			const text = textDocument.getText(range);
+
+			// This edit does not handle all possible
+			// causes of RE0002! It only handles what
+			// I imagine could be a common error where
+			// developers assume "let" is a keyword.
+			const newText = text.replace("let", "var");
```
DanBurton, line 585: This seems slightly dangerous. Does `range` always end before anything else on the line? Is there a risk of `let letter = 'a'` turning into `var varter = 'a'`?

hamirmahal: Thanks for taking a look at this!

> Does `range` always end before anything else on the line?

I actually get `range` from the compiler. I tried tricking the compiler with some weird indentation, and starting the "`textToReplace`" at the beginning of whatever line our compiler says the error is on, seems to work around any inconsistent `range`s the compiler gives us.

> Is there a risk of `let letter = 'a'` turning into `var varter = 'a'`?

[The MDN page for `replace()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#parameters) says

> Only the first occurrence will be replaced.

Also, I tried this with the below example code and we get the output we want.

```js
const text = "let letter = 'a'";
const newText = text.replace("let", "var");
console.log(newText); // "var letter = 'a'"
```

PR #542, opened 2022-01-28
Remove docker save to workspace

vinmazzi:
## Summary
Remove the docker save to circleci workspace and pull the cache from docker registry instead.

Testing was done using this branch on circleCI: https://app.circleci.com/pipelines/github/reach-sh/reach-lang?branch=save-to-workspace-test&filter=all

### DIFF
--- a/.circleci/config.pre.yml
+++ b/.circleci/config.pre.yml
@@ -72,15 +72,6 @@ commands:
               docker tag reachsh/$img:circleci--${CIRCLE_BRANCH} reachsh/$img:$VERSION
             fi
           done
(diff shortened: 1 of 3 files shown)


jeapostrophe (changes requested):
I don't think this is safe with concurrent builds. I think we need to incorporate something like the CIRCLE_SHA1 into the tag --- https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables --- when you download a dependency, but only the branch, when we restore our own

jeapostrophe (approved):

PR #545, opened 2022-01-29
increase visibility of `ask.ts`'s `export`s' types

hamirmahal:


js/stdlib/ts/index.ts
```diff
@@ -1,3 +1,4 @@
+export {ask, done, yesno} from './ask';
```
jeapostrophe, line 1: Does `* as ask` work as well?

hamirmahal: I believe changing the above line to `export * as ask from './ask';` would make it so that developers would be typing things like `ask.ask`, `ask.done` and `ask.yesno`.

image

I'm simulating this with changes to the `node_modules` directory in my tutorial workspace, since I believe we don't have an easy way to see the behavior caused by this change other than a standard library publish, based on what Dan told me.

jeapostrophe: Okay, @DanBurton please review :)

DanBurton: I am in favor of `export * as ask from './ask'`, because I do want those things namespaced like that.

DanBurton (changes requested):

DanBurton (approved):

PR #552, opened 2022-02-01
demo of v0.3 (redesigned interactVs) to Nina + labs reps

chikeabuah:


hs/src/Reach/Simulator/Core.hs
```diff
@@ -793,7 +788,7 @@ bindConsensusMeta (DLRecv {..}) actorId accId = do
 
 instance Interp LLProg where
   interp (LLProg _at _llo slparts _dli _dex _dvs _apis _evts step) = do
-    registerParts $ M.keys $ sps_ies slparts
+    registerParts $ M.toList $ sps_ies slparts
```
jeapostrophe, line 791: I try to always use `toAscList` because I am nervous about depending on the order and it changing someday

hs/src/Reach/Simulator/Server.hs
```diff
@@ -270,6 +270,25 @@ getProgState sid = do
...
 
+checkIfValIType :: IType -> Bool
+checkIfValIType = \case
+  IT_Val _ -> True
+  IT_Fun _ _ -> False
+  IT_UDFun _ -> False
+
+initVals :: InteractEnv -> WebM (M.Map String String)
+initVals (InteractEnv iv'') = do
+  let iv' = M.toList iv''
```
jeapostrophe, line 281: Ditto

hs/src/Reach/Simulator/Server.hs
```diff
@@ -270,6 +270,25 @@ getProgState sid = do
...
+  let iv = filter (\(_,itype) -> checkIfValIType itype) iv'
+  return $ M.fromList $ map (\(var,itype) -> (var, show itype)) iv
```
jeapostrophe, line 283: This function could be implemented entirely inside `M` with `M.filter checkIfValIType` and `M.map show`, which should be more efficient than going in and out of lists

jeapostrophe (approved):

PR #553, opened 2022-02-01
`import` url instead of `require`-ing it for types

hamirmahal:
Without this change:

image

With this change:

image


jeapostrophe (approved):
Do it for everything!

hamirmahal:
Will do!

PR #556, opened 2022-02-02
CORE-308: Track Tokens Dynamically

chrisnevers:


examples/abstract-tok/index.rsh
```diff
@@ -0,0 +1,44 @@
+'reach 0.1';
+
+const MT = Maybe(Token);
```
jeapostrophe, line 3: I think another good test would be making it so the HumbleSwap code is actually the same but with a `Maybe(Token)` parameter for the A side

hs/src/Reach/Eval/Core.hs
```diff
@@ -5465,6 +5479,9 @@ evalStmt = \case
             dv' <- ctxt_mkvar $ DLVar at_c (Just (at_c, de_v)) vt
             let vv = case vt of
                   T_Null -> SLV_Null at_c "case"
+                  T_Token -> case de_val of
+                      SLV_Data _ _ _ (SLV_DLTok (DLToken _ i)) -> SLV_DLTok (DLToken dv' i)
```
jeapostrophe, line 5483: This line shows the brittleness of this technique. I think if you write more test scenarios, you'll see problems with it. For example... a tuple of two tokens, an array of tokens, a function that returns a token based on some condition, a token sent in to the program that is identical to an earlier token (asserted with a require)

I think you could generalize and improve your technique a little bit by making the `ST_Token` type have a `(Maybe Int)` parameter which is used to identify which it is and if it is `Nothing`, then that means "any token". I'm not sure how our type machinery will work with that, but I hypothesize it will be easier. I don't think that the DL types would change at all, but maybe I'm wrong because of the way balance assertions work.

I think the most powerful technique will be "valued based". Remember, the reason we need to keep track of these things is so that we can (a) track the balance adjustments and (b) track the balances in loops with the fluid variable capture. The dumbest thing that could possibly work would be (1) representing the balances as `Map`-like type in DL, (2) using SMT's maps for that during verification, and (3) compiling it to an array on a platform like ALGO that doesn't have maps (and maybe it would be good to be an array on others when we know the identity of tokens statically.) This would basically be like adding a new `T_TokenBalances` type.

Maybe my two ideas are too heavy, but I don't think we can evaluate your idea without more examples of what it allows and doesn't allow.

jeapostrophe (changes requested):

examples/abstract-tok/index.rsh
chrisnevers: I don't think it's possible still because `paySpec` requires a hardcoded array of identifiers

hs/src/Reach/Eval/Core.hs
chrisnevers: I get why this seems brittle, but in this case, we know that if the unboxed value being bound by a pattern match is of type `T_Token`, it can only be `SLV_Data _ _ _ (SLV_DLTok (DLToken _ i))`.

I will update the example to demonstrate how it works with arrays, functions etc. so long as it's statically computable. When a token sent in to the program is identical to an earlier token, it errors as expected.

I think the DL `T_Token` type would require the `Maybe (Int)` information for looking up the balances. In my opinion, this approach could get ugly later in the compiler when we have Map keys that are `T_Token`, so we'd probably have to intercept lookups to treat `lookup (T_Token (Just ..)) map` as `lookup (T_Token Nothing) map`

I kind of get the gist of the value-based approach, but I don't think I understand it well. For

> (1) representing the balances as Map-like type in DL

Would Tokens be `DLVar`s with type: `T_TokenBalances (Map DLVar Int)`?

examples/abstract-tok/index.mjs
```diff
@@ -11,12 +11,14 @@ const startingBalance = pc(100);
 
   const zmd = await stdlib.launchToken(accAlice, "Zorkmid", "ZMD");
   const gil = await stdlib.launchToken(accAlice, "Gil", "GIL");
+  const gol = await stdlib.launchToken(accAlice, "Gold", "AU");
```
jeapostrophe, line 14: I think gold would be GP ;)

examples/abstract-tok/index.rsh
```diff
@@ -29,16 +30,32 @@ export const main = Reach.App(() => {
...
+      Some: ([ t1, t2 ]) => {
+        transfer(balance(t1), t1).to(A);
+      },
+      None: () => {}
+    });
   }
 
-  go(MS(token1));
+  const ta = array(Token, [ getTok(4), token2 ]);
+  go(MS(ta));
```
jeapostrophe, line 45: This is better

examples/abstract-tok/index.rsh
jeapostrophe: Could you write `.paySpec( mTokA.match({None: xyz, Some: abc}) )`? I thought that was an evaluated position?

hs/src/Reach/Eval/Core.hs
jeapostrophe: Let's break this conversation up into two pieces.

First, I want more systematic and small tests (okay if they are just in `hs/t` and compile-time only) that show what is possible _now_, with *this* implementation, and what is still not possible.

The way I'd describe in words is that today, tokens must appear at the top-level of a publish and must remain completely un-abstracted in every operation, thus they must be compile-time static values the entire time. So, you can't publish a Maybe Token and expect it to be visible, you have to publish the actual token. Inside one consensus step, you can't stick two tokens into a container and then pull them out and then be able to use them afterwards. What about across a consensus step... does that matter?

With this change, I think that token identities are preserved in DLArgs, but I don't have a good sense of what that allows or doesn't.

--

Second, the different implementations.

About the type idea... you have introduced an argument `DLToken DLVar Int` with the invariant that `DLVar` is always `DLVar src lab T_Token id`. My point is just that `DLToken (DLVar src lab T_Token id) which` is equivalent to storing the `which` in the token type `DLVar src lab (T_Token (Just which)) id` and is like the common dependent type concept of an indexed type, so I think it would fit in better if you did it that way. (You need the `Maybe`, because in your version some `DLVar`s that are `T_Token` aren't wrapped in a `DLToken`.)

About the map idea... what's the point of keeping track of tokens? It is so that when you do `transfer(x, tok)` we can do `(assert (> (balance tok) x))` and `(fluid-set (balance tok) (- (balance tok) x)`. Everything about tokens is all about that little snippet. Right now, we do that by implementing `(balance tok)` with a fixed array that is referenced by position which requires us to know which index `tok` is at. It's like its type is `Array(UInt, HowManyTokensInProgram)`. We could instead represent it with `Array(Tuple(Token, UInt), HowManyTokensInProgram)` and then implement (balance tok)` with a simple search. In SMT, we could compile that "search" with an actual map. With some clever compilation, we may also be able to get the exact position a lot of the time. I have an idea for how to implement this smoothly that I can explain also, but I think I'd rather speak it than write it.

examples/abstract-tok/index.rsh
chrisnevers: No, it's not evaluated. We use the `JSExpression` directly to generate code in `doFork`

jeapostrophe: Well, maybe we should evaluate it and then synthesize something new in its place :) If not, then we should at least write down the examples that fail/are-not-allowed because it is not evaluated

hs/t/n/Err_InspectAbstractToken.rsh
```diff
@@ -0,0 +1,18 @@
...
+  });
+  init();
+  A.only(() => {
+    const [ token ] = declassify(interact.params);
+    const mt = Maybe(Token).Some(token);
+  });
+  A.publish(mt);
+  commit();
+
+  A.pay([ maybe(mt, 0, (tok) => [ 5, tok ]) ]);
```
jeapostrophe, line 15: What does this program do without your other changes? My assumption is that your code makes it so I can't even write
```
const x = fromMaybe(maybeTok, otherTok);
view.set(x);
```
because I would "inspecting" a token from a container. But the actual problem is that we aren't allowed to be paid in or pay with a token that we learned this way.

hs/t/y/abs_tok_container.rsh
```diff
@@ -0,0 +1,26 @@
+'reach 0.1';
```
jeapostrophe, line 1: This fails w/o your patch?

hs/t/y/abs_tok_fun.rsh
```diff
@@ -0,0 +1,33 @@
+'reach 0.1';
```
jeapostrophe, line 1: Ditto?

hs/src/Reach/Verify/SMT.hs
```diff
@@ -711,13 +712,20 @@ pathAddUnbound at_dv (Just dv) msmte = do
...
+        smtAssert (smtApply "<=" [Atom v, rhs])
+
 pathAddBound :: SrcLoc -> Maybe DLVar -> Maybe SMTExpr -> SExpr -> SMTCat -> App ()
 pathAddBound _ Nothing _ _ _ = mempty
 pathAddBound at_dv (Just dv) de se sc = do
   let DLVar _ _ t _ = dv
   v <- smtVar dv
   let smlet = Just . SMTLet at_dv dv (DLV_Let DVC_Once dv) sc =<< de
   smtDeclare_v v t smlet
+  assertInvariants at_dv t v
```
chrisnevers, line 728: @jeapostrophe  Any harm in this? Otherwise, making claims about balances in loop invariants require `balance(token) <= UInt.max`

hs/src/Reach/AST/DLBase.hs
```diff
@@ -639,6 +638,7 @@ data DLExpr
   | DLE_GetUntrackedFunds SrcLoc (Maybe DLArg) DLArg
   | DLE_FromSome SrcLoc DLArg DLArg
   -- Maybe try to generalize FromSome into a Match
+  | DLE_BalanceInit Int DLVar
```
chrisnevers, line 641: This is for the initial creation of the balance array. At the start of the program, we don't know how big the array will be yet, so this will transform to an actual array after evaluation. DLVar is `<None null>` which will be used to initialize every spot in the array.

hs/src/Reach/AST/DLBase.hs
```diff
@@ -55,12 +55,16 @@ data DLType
   | T_Object (M.Map SLVar DLType)
   | T_Data (M.Map SLVar DLType)
   | T_Struct [(SLVar, DLType)]
+  | T_TokenBalances Integer
```
chrisnevers, line 58: Don't know the size of the array until after evaluation; this will transform to the actual array type

hs/src/Reach/Linearize.hs
```diff
@@ -338,15 +338,23 @@ data DFEnv = DFEnv
...
   }
 
 df_allocVar :: DFApp Int
 df_allocVar = allocVar eCounter_df
 
+adjustFluids :: FluidVar -> DFApp FluidVar
+adjustFluids = \case
+  FV_balances _ _ -> do
+    eBals <- asks eBals
+    return $ FV_balances eBals mempty
```
chrisnevers, line 351: I temporarily did all the work of tweaking the balance values/types to the correct array size in Linearize just to see it work. I think I want to make a new pass immediately after evaluation

hs/rsh/stdlib.rsh
```diff
@@ -166,6 +166,9 @@ export const Array_indexOf =
 export const Array_indexOf1 =
   (arr) => (x) => Array_indexOf(arr, x);
 
+export const Array_set =
+  (x, y, z) => Array.set(x, y, z)
```
jeapostrophe, line 170: Why?

hs/src/Reach/AST/DLBase.hs
jeapostrophe: What's the integer?

hs/src/Reach/AST/DLBase.hs
```diff
@@ -224,6 +229,23 @@ litTypeOf = \case
   DLL_Bool _ -> T_Bool
   DLL_Int {} -> T_UInt
 
+data DLToken = DLToken DLVar Int
```
jeapostrophe, line 232: Don't you not need this anymore?

hs/src/Reach/AST/DLBase.hs
jeapostrophe: Why are they maybes, rather than just UInts? And what's the number?

hs/src/Reach/AST/DLBase.hs
```diff
@@ -1058,7 +1088,7 @@ instance Pretty a => Pretty (DLRecv a) where
       <> render_nest (pretty dr_k)
 
 data FluidVar
-  = FV_balance Int
+  = FV_balances Integer (M.Map DLArg Int)
```
jeapostrophe, line 1091: What's the integer and the map? I thought this was going to just be `FV_balances` (no parameter)

hs/src/Reach/AST/DLBase.hs
```diff
@@ -1083,9 +1113,12 @@ instance Pretty FluidVar where
     FV_baseWaitSecs -> "baseWaitSecs"
     FV_didSend -> "didPublish"
 
+maybeTokenMap :: M.Map SLVar DLType
+maybeTokenMap = M.fromList [("None", T_Null), ("Some", T_Token)]
+
 fluidVarType :: FluidVar -> DLType
 fluidVarType = \case
-  FV_balance _ -> T_UInt
+  FV_balances i _ -> T_TokenBalances (i + 1)
```
jeapostrophe, line 1121: I thought that `FV_balances` would be of type `T_TokenBalances` (no parameter) or we could manage to avoid looking at it until later and thus never need to see the type, so this would be `impossible`

hs/src/Reach/Eval/Core.hs
```diff
@@ -2184,13 +2184,117 @@ doBalanceInit_ fv mtok amta = do
...
+  bal <- doArrRef_ tup (SLV_Int at 1)
+  compileToVar bal
+
+getBalsEnv :: FluidVar -> M.Map DLArg Int
+getBalsEnv = \case
+  FV_balances _ m -> m
+  _ -> impossible "getBalsEnv: Expected FV_balances"
+
+networkBal :: SrcLoc -> SLVal
+networkBal at = SLV_Data at maybeTokenMap "None" $ SLV_Null at "networkBalance"
```
jeapostrophe, line 2217: Ah, I understand why you have the maybe now. I think it will be simpler to just have two paths... one for the network and one for the non-network. All these maybes are complex

hs/src/Reach/Eval/Core.hs
```diff
@@ -2184,13 +2184,117 @@ doBalanceInit_ fv mtok amta = do
...
+            indexOf <- lookStdlib "Array_findIndex"
+            let v = public $ SLV_Data at maybeTokenMap "Some" $ SLV_DLVar dv
+            checkKeyF <- evalApplyVals' (snd checkKey) [v]
+            mIdx <- evalApplyVals' indexOf $ [b, checkKeyF]
+            fromSome <- lookStdlib "fromSome"
+            r <- evalApplyVals' fromSome $ [mIdx, public $ SLV_Int at 0]
+            return r
+    Just ow -> impossible $ "getBalanceIndex: tsv received: " <> show ow
+
+setBalance :: Maybe DLArg -> DLArg -> ReaderT Env IO DLArg
```
jeapostrophe, line 2250: Use `App`

hs/src/Reach/Eval/Core.hs
```diff
@@ -2184,13 +2184,117 @@ doBalanceInit_ fv mtok amta = do
...
+  bals <- lookupBalances
+  case mtok of
+    Nothing -> do
+      return $ public $ SLV_Int at 0
+    Just (DLA_Var dv) -> do
+        let env = getBalsEnv bals
+        case join $ fmap (flip M.lookup env) mtok of
+          Just i -> return $ public $ SLV_Int at $ fromIntegral i
+          Nothing -> do
+            b <- doFluidRef bals
```
jeapostrophe, line 2239: Ah, so this expands to an array of at least as big as the real thing at this point in the program

hs/src/Reach/Eval/Core.hs
```diff
@@ -2184,13 +2184,117 @@ doBalanceInit_ fv mtok amta = do
...
+getBalanceIndex mtok = do
+  at <- withAt id
+  bals <- lookupBalances
+  case mtok of
+    Nothing -> do
+      return $ public $ SLV_Int at 0
+    Just (DLA_Var dv) -> do
+        let env = getBalsEnv bals
+        case join $ fmap (flip M.lookup env) mtok of
+          Just i -> return $ public $ SLV_Int at $ fromIntegral i
```
jeapostrophe, line 2237: And this path is looking for when it is fixed and happens to be static

hs/src/Reach/Eval/Core.hs
```diff
@@ -2205,7 +2309,22 @@ doBalanceAssert_ fv mtok lhs op msg = do
...
+    b <- lookupBalances
+    -- Find the current balance of mtok
+    bsv <- getBalanceOf mtok
+    -- Apply op to value
+    bv' <- evalApplyVals' up_rator [bsv]
+    bva <- compileCheckType T_UInt $ snd bv'
+    -- Set value/fluid ref with updated array
+    b' <- setBalance mtok bva
+    doFluidSet_ b b'
+
```
jeapostrophe, line 2327: This whole approach is interesting and different than I was thinking. I was thinking there would be `DLE_BalanceGet` and `DLE_BalanceSet` and `T_Balances` and then `FV_balances` of type `T_Balances` and then we'd transform them. But, thinking now, we might even get away with making `DLS_BalanceInit SrcLoc`, `DLS_BalanceGet SrcLoc DLVar DLArg`, and `DLS_BalanceSet SrcLoc DLVar DLArg DLArg` and then expand those into the array of the right size and uses of fluid set and ref. Then the rest of the compiler doesn't need to think about them at all and it happens before Linearize and we can discover the correct size right away.

hs/src/Reach/Eval/Error.hs
```diff
@@ -727,5 +729,7 @@ instance Show EvalError where
       "The " <> intercalate ", " (map bunpack $ S.toList apis) <> " `API`(s) cannot explicitly make a publication."
     Err_ForkNoCases ->
       "A `fork`/`parallelReduce` statement must include a `.case` or `.api` component"
+    Err_InspectAbstractToken ->
+      "Cannot inspect `Token` that was published inside of a container."
```
jeapostrophe, line 733: Do you still throw this error? I don't think so?

hs/src/Reach/Verify/SMT.hs
jeapostrophe: No harm

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
chrisnevers: I will try this approach/what you described in slack

hs/src/Reach/AST/DLBase.hs
```diff
@@ -193,19 +197,26 @@ instance Pretty DLInit where
 
 data DLConstant
   = DLC_UInt_max
-  deriving (Eq, Generic, Show, Ord)
+  | DLC_Zero_addr
```
jeapostrophe, line 200: It's not really a zero address... it is a zero token, which would be `DLC_Token_zero`

hs/src/Reach/AST/DLBase.hs
```diff
@@ -55,15 +55,19 @@ data DLType
   | T_Object (M.Map SLVar DLType)
   | T_Data (M.Map SLVar DLType)
   | T_Struct [(SLVar, DLType)]
-  | T_TokenBalances Integer
+  | T_Balances
```
jeapostrophe, line 58: Because of the way that `fluidVarType` is used, I don't think you actually need to define this, provided you special case it in the `While` part of `Linearize` with the actual type. This will simplify things greatly.

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -112,7 +112,9 @@ conName' :: T.Text
 conName' = "ALGO"
 
 conCons' :: DLConstant -> DLLiteral
-conCons' DLC_UInt_max = DLL_Int sb $ 2 ^ (64 :: Integer) - 1
+conCons' = \case
+  DLC_UInt_max  -> DLL_Int sb $ 2 ^ (64 :: Integer) - 1
+  DLC_Zero_addr -> DLL_Int sb $ 0
```
jeapostrophe, line 117: This is not really correct... Right now the code makes the assumption that "constants can be turned into literals", but this conversion would be of the wrong type, because `Token` is a different type than `Int`

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -57,7 +57,9 @@ conName' :: T.Text
 conName' = "ETH"
 
 conCons' :: DLConstant -> DLLiteral
-conCons' DLC_UInt_max = DLL_Int sb $ 2 ^ (256 :: Integer) - 1
+conCons' = \case
+  DLC_UInt_max  -> DLL_Int sb $ 2 ^ (256 :: Integer) - 1
+  DLC_Zero_addr -> impossible "ETH conCons': zero_addr"
```
jeapostrophe, line 62: See the comment about, but in terms of code generation it is possible... `address(0)`

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -478,6 +480,7 @@ solLit = \case
 solArg :: AppT DLArg
 solArg = \case
   DLA_Var v -> solVar v
+  DLA_Constant DLC_Zero_addr -> return "payable(address(0))"
```
jeapostrophe, line 483: Ah, you know that

hs/src/Reach/Linearize.hs
```diff
@@ -532,6 +573,19 @@ df_eb :: DKExportBlock -> DFApp DLExportBlock
...
+-- Initialize the fluid arrays
+df_init :: DKTail -> DFApp DKTail
+df_init k = do
+  dv <- fixFluidRefType_ =<< DLVar sb Nothing T_Balances <$> df_allocVar
+  nv <- DLVar sb Nothing tokenInfoElemTy <$> df_allocVar
+  let lv = DLV_Let DVC_Many dv
+  let false = DLA_Literal $ DLL_Bool False
+  let zero = DLA_Literal $ DLL_Int sb 0
+  let c1 = DKC_Let sb (DLV_Let DVC_Many nv) $ DLE_LArg sb $ DLLA_Tuple [DLA_Constant DLC_Zero_addr, zero, zero, false]
+  let c2 = DKC_Let sb lv $ DLE_BalanceInit nv
```
jeapostrophe, line 585: I don't understand what `DLE_BalanceInit` is for

hs/src/Reach/Simulator/Core.hs
```diff
@@ -363,7 +363,9 @@ interpPrim = \case
   (f, args) -> impossible $ "unhandled primop" <> show f <> " " <> show args
 
 conCons' :: DLConstant -> DLVal
-conCons' DLC_UInt_max = V_UInt $ 2 ^ (64 :: Integer) - 1
+conCons' = \case
+  DLC_UInt_max  -> V_UInt $ 2 ^ (64 :: Integer) - 1
+  DLC_Zero_addr -> V_Address 0
```
jeapostrophe, line 368: It's a token

jeapostrophe (changes requested):

hs/src/Reach/Linearize.hs
chrisnevers: I will remove it as I can just do the conversion to an array directly in this function

hs/src/Reach/Eval/Core.hs
```diff
@@ -4291,7 +4326,28 @@ compilePayAmt tt v = do
...
+                    when seenNet $ do
+                      locAt v_at $ expect_ $ Err_Transfer_DoubleNetworkToken tt
+                    a <- DLA_Var <$> (ctxt_lift_expr (DLVar at Nothing T_UInt) $ DLE_TupleRef at (DLA_Var dv) idx)
+                    return $ ((True, sks), DLPayAmt a tks)
+                  tupTy@(T_Tuple [T_UInt, T_Token]) -> do
+                    let ((seenNet, sks), DLPayAmt nts tks) = sa
+                    tup <- DLA_Var <$> (ctxt_lift_expr (DLVar at Nothing tupTy) $ DLE_TupleRef at (DLA_Var dv) idx)
+                    amt_a <- DLA_Var <$> (ctxt_lift_expr (DLVar at Nothing T_UInt) $ DLE_TupleRef at tup 0)
+                    token_a <- DLA_Var <$> (ctxt_lift_expr (DLVar at Nothing T_UInt) $ DLE_TupleRef at tup 1)
+                    when (token_a `elem` sks) $ do
```
chrisnevers, line 4345: @jeapostrophe Would verifying that all the tokens in the PayAmt are unique be left up to the programmer? For example, if there are two dynamic token references, would something like the following be required:

```Javascript
const tokA = // dynamic computation
const tokB = // dynamic computation
A.pay([ [1, tokA], [1, tokB] ]).payRequire(() => {
  require(tokA != tokB);
);
```

jeapostrophe: First, for now, you'd generate an `assert` and then the user is responsible for adding the appropriate `assume`/`require`. Meaning `Err_Transfer_DoubleToken` goes away and is replaced by a verification error

Second, ideally we'd be able to remove this constraint all together and support examples where the same token is given multiple times.

hs/src/Reach/Linearize.hs
```diff
@@ -389,88 +389,114 @@ expandFromFVMap at (DLAssignment updatem) = do
...
+        DT_Com (asn cnd $ DLE_PrimOp at IF_THEN_ELSE [DLA_Var is_found, DLA_Literal $ DLL_Bool True, DLA_Var toks_eq]) $
+        DT_Com (asn new_idx $ DLE_PrimOp at ADD [DLA_Var idx, DLA_Literal $ DLL_Int at 1]) $
+        DT_Com (asn failAcc $ DLE_LArg at $ DLLA_Tuple [DLA_Literal $ DLL_Bool False, DLA_Var new_idx]) $
+        DT_Com (asn succAcc $ DLE_LArg at $ DLLA_Tuple [DLA_Literal $ DLL_Bool True, DLA_Var idx]) $
+        DT_Com (asn bl_res $ DLE_PrimOp at IF_THEN_ELSE [DLA_Var cnd, DLA_Var succAcc, DLA_Var failAcc]) $
+        DT_Return at
+  let bl = DLBlock at [] ss $ DLA_Var bl_res
+  let s0 = asn init_acc_dv $ DLE_LArg at $ DLLA_Tuple [DLA_Literal $ DLL_Bool False, DLA_Literal $ DLL_Int at 0]
+  let s1 = DL_ArrayReduce at reduce_res bals (DLA_Var init_acc_dv) acc_dv elem_dv bl
+  -- XXX add check to die if `is_found` false?
```
jeapostrophe, line 443: Add an `assert` that it is `true`

hs/src/Reach/Linearize.hs
```diff
@@ -389,88 +389,114 @@ expandFromFVMap at (DLAssignment updatem) = do
...
     tup   <- mkVar at tokenInfoElemTy
-    bals' <- mkVar at T_Balances
+    bals' <- mkVar at arrTy
     tup'  <- mkVar at tokenInfoElemTy
-    tok   <- mkVar at T_Token
     bal   <- mkVar at $ T_UInt
     supply <- mkVar at $ T_UInt
     destroyed <- mkVar at $ T_Bool
     let c1 = asn tup $ DLE_ArrayRef at bals idx
     -- XXX Make DLE_TupleSet? This is kind of gross
```
jeapostrophe, line 480: Probably

hs/src/Reach/Optimize.hs
```diff
@@ -599,11 +599,12 @@ instance Optimize DLStmt where
...
     DL_LocalSwitch at ov csm ->
       optSwitch (DL_LocalDo at) DT_Com DL_LocalSwitch at ov csm
-    s@(DL_ArrayMap at ans x a i f) ->
-      maybeUnroll s x $
-        DL_ArrayMap at ans <$> opt x <*> pure a <*> pure i <*> opt f
-    s@(DL_ArrayReduce at ans x z b a f) -> maybeUnroll s x $ do
-      DL_ArrayReduce at ans <$> opt x <*> opt z <*> (pure b) <*> (pure a) <*> opt f
+    DL_ArrayMap at ans x a i f -> do
+      s' <- DL_ArrayMap at ans <$> opt x <*> pure a <*> pure i <*> opt f
+      maybeUnroll s' x $ return s'
```
jeapostrophe, line 604: Whoops! This is great

hs/t/n/nonunique_payamt.rsh
```diff
@@ -0,0 +1,31 @@
+'reach 0.1';
```
jeapostrophe, line 1: Missing `.txt`

chrisnevers:
@jeapostrophe  Some verification issues arose when I added the assertion that a dynamic token lookup must be successful. 
The issue is that the `tokenInfo` array can be a loop variable, so its token keys "could" change. A reach program can't express that the token keys will be static and I felt like enforcing the SMT to treat the `tokenInfo` array specially would be a large undertaking. 

So instead, I split the token references into their own array:

```Javascript
tokenInfo: Array([UInt, UInt, Bool], <size>);
tokens : Array(Token, <size>);
```

I explicitly prevent `tokens` from becoming a loop variable. Now, the token array is static once initialized. However, since it can't become a loop variable we still have the restriction that you can't publish tokens within a loop.

hs/src/Reach/Eval/Core.hs
```diff
@@ -2387,15 +2426,20 @@ evalPrim p sargs =
           DLE_TokenNew at tns
       tokdv <- doInternalLog_ Nothing tokdv_
       st <- readSt id
+      let existingToks = st_toks st
+      tokIsUniq <- tokIsUnique (map DLA_Var existingToks) $ DLA_Var tokdv
```
chrisnevers, line 2430: When creating a token, assume that it's unique from all the existing tokens. This is needed for verifying that the tokens in a pay amount are unique

jeapostrophe: This is a valid assumption too :)

hs/t/n/Err_Transfer_DoubleToken.rsh
```diff
@@ -1,20 +0,0 @@
-'reach 0.1';
```
jeapostrophe, line 1: We should never delete tests, but we can move them or rename them

hs/t/n/mint-fail1.txt
```diff
@@ -9,23 +9,133 @@ Verification failed:
...
+  const UInt.max = 1;
+
+  const v314 = protect<Object({"amt": UInt, "doEarlyTransfer": Bool, "metadata": Bytes(32), "name": Bytes(32), "supply": UInt, "symbol": Bytes(8), "url": Bytes(96)})>("Alice".interact.getParams());
   //    ^ could = {amt: 0, doEarlyTransfer: false, metadata: "Bytes!val!0", name: "Bytes!val!1", supply: 1, symbol: "Bytes!val!2", url: "Bytes!val!3" }
   //      from: ./mint-fail1.rsh:24:104:application
 
   // Theorem Formalization
 
-  const v362 = 0 == v257.supply;
+  const v307 = [0, 0, false ];
```
jeapostrophe, line 20: This is really confusing... is it plausible to insert in some labels on the variables in your variable making function?

jeapostrophe (changes requested):
Can you also check that the cost of Humble doesn't change?

chrisnevers:
Ready to go, duoswap is now 707 cost though

jeapostrophe (commented):
.

jeapostrophe (approved):

PR #560, opened 2022-02-03
informative node labels + color-coded multi-actor poi loc highlighting

chikeabuah:


examples/programmer-art/src/index.mjs
```diff
@@ -73,12 +73,18 @@ const renderObjects = async (nodeId) => {
...
   let actors = ``
+  let actorsNoCons = ``
   const actorEntries = Object.entries(actorSet)
   // NOTE: assumption: there is at least one non-consensus actor
   const firstActorId = actorEntries[0][0]
   for (const [k,v] of actorEntries) {
     actors = actors + `<option value="${k}">${v}</option>`
   }
+  for (const [k,v] of actorEntries) {
+    if (parseInt(k) != -1 ) {
```
hamirmahal, line 84: I think we usually prefer doing `===` and `!==` in JavaScript to avoid unexpected behavior.

We shouldn't need the leniency of `!=` in this case, anyway, because of `parseInt`.

```js
-1 === "-1"
false
-1 == "-1"
true
```

jeapostrophe (approved):

PR #563, opened 2022-02-07
CORE-1107 Add balancesOf to frontend stdlib

erer1243:
Add `balancesOf` to stdlib, a function that is more efficient for retrieving multiple token balances at once than repeated calls to `balanceOf`.' Also add a small example that uses it.


erer1243:
redoing on org owned branch

PR #565, opened 2022-02-07
test `ctc-address` `own-addr`

chikeabuah:


examples/simulator-own-addr/client.mjs
```diff
@@ -0,0 +1,47 @@
+const c = await import('@reach-sh/simulator-client');
```
jeapostrophe, line 1: I haven't noticed if the others are like this, but you should use the real import syntax

examples/simulator-own-addr/index.rsh
```diff
@@ -0,0 +1,33 @@
+'reach 0.1';
+'use strict';
+
+export const main =
+  Reach.App(
```
jeapostrophe, line 5: You should not commit new code with the ternary syntax

jeapostrophe: Or is this just a copy of something else? If so, you could do the thing where you copy and example into this directory.

jeapostrophe (changes requested):

examples/simulator-own-addr/client.mjs
chikeabuah: @jeapostrophe what's wrong with this syntax? i prefer this syntax because it also works in the node REPL

chikeabuah: every simulator example is written like this

chikeabuah: but okay, i'll change it if it's best practice, one moment

jeapostrophe (approved):

PR #570, opened 2022-02-09
test `map-any` 

chikeabuah:


hs/src/Reach/Simulator/Core.hs
```diff
@@ -260,12 +260,20 @@ data DLVal
   | V_Object (M.Map SLVar DLVal)
   | V_Data SLVar DLVal
   | V_Struct [(SLVar, DLVal)]
+  | V_Maybe DLMaybeVal
```
jeapostrophe, line 263: I don't get this and am inclined to disapprove. If it is there because of Haskell, then you want a `Maybe`, if it is there because of Reach, then you want a `V_Data`

hs/src/Reach/Simulator/Core.hs
```diff
@@ -486,8 +494,12 @@ instance Interp DLExpr where
...
       let linstate = e_linstate g
       acc <- vAddress <$> interp dlarg
-      let m = saferMaybe "DLE_MapRef1" $ M.lookup dlmvar linstate
-      return $ saferMaybe "DLE_MapRef2" $ M.lookup acc m
+      case M.lookup dlmvar linstate of
+        Nothing -> return $ V_Maybe V_Nothing
+        Just m -> do
+          case M.lookup acc m of
+            Nothing -> return $ V_Maybe V_Nothing
+            Just m' -> return $ V_Maybe $ V_Just m'
```
jeapostrophe, line 502: Ah, yes, these need to return a `V_Data`

jeapostrophe (changes requested):

hs/src/Reach/Simulator/Core.hs
chikeabuah: what should the `SLVar` in `V_Data SLVar DLVal` be here? there are no obvious string values.

chikeabuah: oh never mind

chikeabuah: okay!

hs/src/Reach/Simulator/Core.hs
chikeabuah: okay!

hs/src/Reach/Simulator/Core.hs
chikeabuah: thanks

jeapostrophe (approved):

PR #574, opened 2022-02-10
add command-line flag to print keyword/identifier completion info

willghatch:
So far this prints CompletionItemKind for keywords and identifiers in
the base_env.  This includes many identifiers that aren't in the
current keywordToItemKind.json file, but is still missing many others.

The other identifiers are mostly strewn inside pattern matches inside Eval/Core.hs and in stdlib.rsh.  I imagine we will want a mode that can take a Reach file and return this and other info about identifiers defined/used in the file, and that that would work well for getting the identifiers in the stdlib file.  I'm not sure what the best approach for the various identifiers mixed up in pattern matching would be, so I haven't tried to do that yet.  I think just having a table with the info is as good as anything else -- I don't think there's a great way to keep that part of the list in a way that's clearly and obviously always updated in sync.


jeapostrophe:
I don't think we've ever really nailed down what the best way to edit is, but ultimately you need to eventually rebase to stay up to date and that is going to confuse the conversation tracking a little bit, but as the code grows differently enough, a squash is typically optimal for dealing with the rebasing conflicts, so I think it is necessary.

jeapostrophe (approved):
This is cool! I'd like you and @hamirmahal to connect later about updating the files that the editor actually uses and think about the best way (please make a Jira card about this)

Ideas:
1. IDE calls this option and gets it live when it starts
2. This option is called during an IDE release to generate the file
3. We make a test that fails if the IDE's committed file doesn't match what this outputs, then it has to be updated manually (and we make that easy)

I think 3 is easy and perhaps the best, because it will force someone adding a new thing to the base environment to consider what completion should be made.

The next thing after this is to figure out a way to tell the IDE that `Array.length` exists.

PR #582, opened 2022-02-14
Fix seclinks with empty titles.

willghatch:
We have some section headings where (part of) the title of the section
is in backticks, like `this`.  When processing section headings like
that that define anchors for `seclink`, the backtick splits the text
into different list elements.  Previously we only used the first
element in a list to generate the title used for references, and thus
for `seclink`.  This resulted in links with empty text in the anchor.
The links could be discovered using the web inspector, but were
invisible and unclickable on the page.  Also, everywhere `seclink` is
used, some text output is expected gramatically at the use site.

This patch changes the processing to include backticked words and text
following them to be included in the title.  This makes previously
empty `seclink`s visible, legible, and clickable.

It also adds a couple of guard rails with `fail` if any empty titles
for xrefs make it through this improved processing in the future.

### DIFF
--- a/docs/dev/generator.js
+++ b/docs/dev/generator.js
@@ -98,6 +98,9 @@ const xrefPut = (s, t, v) => {
       fail(`Duplicated xref`, s, t, e, v);
     }
   }
+  if ( v.title == "" ){
+    fail(`Xref insertion with empty title`, s, t, v)
+  }
   const parts = t.split(".");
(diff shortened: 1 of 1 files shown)


willghatch:
Here are some screenshots for reference:

Before:
empty-seclink-before

After:
empty-seclink-after

jeapostrophe (approved):

jeapostrophe:
This is great!

PR #584, opened 2022-02-15
wip exception based implementation of core-579

erer1243:
WIP implementation of core-579, still with caveman prints and debugging code. Uploaded for comment on style.

An example that shows the json object that would be uploaded vs what would be printed to the user (everything immediately after the json) 
image

### DIFF
--- a/hs/src/Reach/Report.hs
+++ b/hs/src/Reach/Report.hs
@@ -6,16 +6,26 @@ import Control.Exception
 import Control.Monad.Reader
 import Data.Aeson
 import Data.Time
+import Data.Typeable (cast)
 import Network.HTTP.Client.Conduit (httpNoBody)
 import Network.HTTP.Client.TLS
 import Network.HTTP.Conduit
 import Network.HTTP.Simple (setRequestBodyJSON, setRequestMethod)
 import Reach.CommandLine
 import Reach.Version
+import Reach.AST.Base (CompileErrorException(..), encodeJSONString)
 import System.Environment
 
 --- TODO maybe have each part collect some information and report it back through a (Map String String)
 type Report = Either SomeException ()
+
+reportResult :: Report -> String
+reportResult = \case
+  Right () -> "Right ()"
+  report@(Left (SomeException inner)) ->
+    case (cast inner :: Maybe CompileErrorException) of
+      Just cee -> encodeJSONString cee
+      Nothing -> show report
 
 startReport :: Maybe String -> String -> IO (Report -> IO ())
 startReport mwho i = do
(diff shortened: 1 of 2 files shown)


examples/donation-balancesOf/index.rsh
```diff
@@ -2,6 +2,8 @@
 'use strict';
 
 export const main = Reach.App(() => {
+  declassify(); // make an error
```
jeapostrophe, line 5: You should test by running on one of the files in `hs/t/n`

hs/src/Reach/Report.hs
```diff
@@ -1,22 +1,44 @@
...
+  , cee_pretty :: String
+  }
+
+instance Show CompileErrorException where
+  show = cee_pretty
+
+instance Exception CompileErrorException
+
+reportResult :: Report -> String
+reportResult (Right ()) = "Right ()"
```
jeapostrophe, line 36: I like `\case`

hs/src/Reach/Report.hs
```diff
@@ -1,22 +1,44 @@
...
 
+-- Remove when finished
+import GHC.Exts
+import Data.List (intercalate)
+
 --- TODO maybe have each part collect some information and report it back through a (Map String String)
 type Report = Either SomeException ()
 
+data CompileErrorException = CompileErrorException
+  { cee_json :: String
```
jeapostrophe, line 26: How about making this a `CompilationError` or a JSON `Value` and then you can put the encoding stuff somewhere else

hs/src/Reach/Report.hs
```diff
@@ -1,22 +1,44 @@
...
+  , cee_pretty :: String
+  }
+
+instance Show CompileErrorException where
+  show = cee_pretty
```
jeapostrophe, line 31: This is actually wrong and we do need that unsafe thing, because it exists because the IDE wants to see errors as JSON.

hs/src/Reach/Report.hs
```diff
@@ -1,22 +1,44 @@
...
 import System.Environment
 
+-- Remove when finished
+import GHC.Exts
+import Data.List (intercalate)
+
 --- TODO maybe have each part collect some information and report it back through a (Map String String)
 type Report = Either SomeException ()
 
+data CompileErrorException = CompileErrorException
```
jeapostrophe, line 25: I think this type should be somewhere else, perhaps where `CompilationError` is defined.

jeapostrophe (changes requested):

erer1243:
An example using a test from `hs/t/n/` 
image

jeapostrophe (changes requested):
I believe this implementation is ready to merge (once you remove the testing stuff you added)

jeapostrophe:
Make sure you rebase, because William touched something close

erer1243:
Rebased & removed testing stuff. Good to go.

erer1243:
Actually, maybe I did something wrong with the rebase. Or misunderstand rebasing in general :P . Looks like what William did still caused a conflict, so I used the web interface to fix it and make that merge commit. We both changed how `CompilationError`s are converted to Strings. I just kept my changes to the relevant part of the file.

jeapostrophe:
You'll need to remove the merge commit so we can rebase this onto master. Also you should make sure his new code (that prints warnings) also uses your new function.

jeapostrophe:
When you remove the merge, you'll need to checkout to before it, reset the branch, rebase yourself, then force push. You may want to interactively rebase and squash the commits into one

erer1243:
Whoops my local master was not up to date. Wondered why the rebase was so easy the first time. Let's try this one?

jeapostrophe (approved):

PR #587, opened 2022-02-16
output compiler warnings as json like errors

willghatch:
As an example, if I use an integer as a time argument, it will print this:

```
warning: {"ce_errorCode":"RW0000","ce_offendingToken":null,"ce_errorMessage":"Using a bare value as a time argument is now deprecated. Please use relativeTime, absoluteTime, relativeSecs, or absoluteSecs.","ce_position":[58,13],"ce_suggestions":[]}
```
This matches the format it uses for errors.  @hamirmahal , is that the format you would like?  Also, I left it as printing to stderr.  Would you prefer it be on stdout?  I'm not sure what would be better for the IDE plugin.

### DIFF
--- a/hs/src/Reach/AST/Base.hs
+++ b/hs/src/Reach/AST/Base.hs
@@ -162,22 +162,26 @@ getErrorMessage mCtx src isWarning ce = do
       <> hardline
       <> docsUrl
 
+makeErrorJson :: (ErrorSuggestions a, ErrorMessageForJson a, Show a, HasErrorCode a) => SrcLoc -> a -> [Char]
+makeErrorJson src err =
+  (map w2c $
+     LB.unpack $
+       encode $
+         CompilationError
+           { ce_suggestions = snd $ errorSuggestions err
+           , ce_offendingToken = fst $ errorSuggestions err
+           , ce_errorMessage = errorMessageForJson err
+           , ce_position = srcloc_line_col src
+           , ce_errorCode = makeErrCode (errPrefix err) (errIndex err)
+           })
+
 expect_throw :: (HasErrorCode a, Show a, ErrorMessageForJson a, ErrorSuggestions a) => HasCallStack => Maybe ([SLCtxtFrame]) -> SrcLoc -> a -> b
 expect_throw mCtx src ce =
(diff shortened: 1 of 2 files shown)


hs/src/Reach/AST/Base.hs
```diff
@@ -162,22 +162,26 @@ getErrorMessage mCtx src isWarning ce = do
       <> hardline
       <> docsUrl
 
+makeErrorJson :: (ErrorSuggestions a, ErrorMessageForJson a, Show a, HasErrorCode a) => SrcLoc -> a -> [Char]
+makeErrorJson src err =
+  (map w2c $
```
jeapostrophe, line 167: You don't need this paren

hs/src/Reach/Warning.hs
```diff
@@ -80,13 +81,18 @@ instance Show Warning where
...
-            let style s = if hasColor then TC.style s else id
-            let color s = if hasColor then TC.color s else id
-            style TC.Bold (color TC.Yellow "WARNING") <> ": " <> show d
-  hPutStrLn stderr msg
+emitWarning at d =
+  case unsafeIsErrorFormatJson of
+    True ->
+      hPutStrLn stderr $ "warning: " ++ makeErrorJson (fromMaybe mempty at) d
+    False ->
+      do
```
jeapostrophe, line 89: I like this do on the previous line

jeapostrophe (changes requested):
@hamirmahal I would like Hamir to confirm (potentially by adding to this branch) that this format works for him before merging

willghatch:
Pushing so Hamir can see some extra suggestions added.  I don't love this one, because it essentially duplicates a lot of suggestion text between the JSON and the prose error message.  But I also don't love any fixes to that problem, frankly.

jeapostrophe:
I like it and don't find the duplication distasteful like you. Maybe you can define variables for the list of strings and use them in two places, but it's not so bad.

jeapostrophe:
@hamirmahal please comment

jeapostrophe (approved):
I'm actually going to merge this and let @hamirmahal do a separate branch

PR #590, opened 2022-02-16
link ReactJS docs for easier access for developers

hamirmahal:
https://user-images.githubusercontent.com/43425812/154350861-83860a60-4bb3-4ec1-9866-44a04b585312.mp4

### DIFF
--- a/docs/src/tut/rps/index.md
+++ b/docs/src/tut/rps/index.md
@@ -1593,7 +1593,7 @@ range: 39-41
 ```
 
 + On line 19, we initialize the component state to display @{seclink("tut-9-ConnectAccount")}.
-+ On lines 21 thru 31, we hook into React's `{!js} componentDidMount` lifecycle event, which is called when the component starts.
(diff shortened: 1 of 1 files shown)


willghatch:
One thing I've been thinking is that it would be nice if JS functions generally were hyperlinked in our docs.  Eg. maybe we could have an xref form like `@{ref("js", "someFunc", "url-to-react-or-MDN-or-something")}`, or `@{externalRef(...)}` some other way to enter xrefs into our database that point externally, but that can show up in our JS code blocks.  That might make it easier for people to see which JS bits are from Reach and which are external, and also help people reading the docs that aren't very familiar with Javascript.

willghatch:
Ultimately I absolutely agree with more hyperlinking.

jeapostrophe (approved):

jeapostrophe:
@willghatch I think that is an awesome idea and if that were implemented, we could potentially change this code block to "`{!js} React.componentDidMount`". Could you make a Jira card for that?

PR #595, opened 2022-02-18
CORE-1145 | fetch balances with BigInt precision

cbaunach:
## Summary of [CORE-1145](https://reachsh.atlassian.net/browse/CORE-1145)

This PR implements [setIntDecoding](https://algorand.github.io/js-algorand-sdk/classes/_internal_.LookupAccountByID.html#setIntDecoding) on two methods (`accountInformation` and `lookupAccountByID`) that were found to need it. Without this fix, balanceOf would return 18446744073709552000 when the balances specified on the account was 18446744073709551615; this is due to the fact that normally the two methods specified above returned the balances as numbers which do not have that high of precision. By implementing setIntDecoding I was able to have them return the balances as BigInts which do allow that level of precision; however, a side effect of this solution is now all values that were numbers are now returned as bigint type (though this does not appear to have had that drastic of a domino effect)

I could expand this PR to apply setIntDecoding to a larger segment of ALGO.ts but attempts showed that it would be too complex and I am pressed for time.

I created an example in the examples folder that can be run with `REACH_CONNECTOR_MODE=ALGO ./reach run`. This test only pertains to balanceOf but I can attempt to create additional tests for the other locations affected by this change. The other locations affected by this change I have tested by pulling out segments of the ALGO.ts file and testing in isolation (hence the change that was made to ApplicationID in the `getLocalState` function

I don't believe there needs to be any changes to the docs; this an internal fix that simply allows an existing function to work as expected in certain edgecases


cbaunach:
I have noticed some problems with this and will be addressing them on Friday

jeapostrophe:
Can you elaborate on the problem at all?

cbaunach:
> Can you elaborate on the problem at all?

The example needs to be more rigid. In it's current state, someone can access the account and change the balances of the two tokens being checked. I should be generating the tokens programmatically for a temporary account instead of relying on an existing account that might have it's balances change

jeapostrophe:
I made a test and example, but it fails. I think it is because the devnet algod/indexer version doesn't support this encoding

cbaunach:
> I made a test and example, but it fails. I think it is because the devnet algod/indexer version doesn't support this encoding

Hey Jay, I used npm link with your changes and tested your example in a separate location and it appears to work; do you have the error that it output when it failed?

jeapostrophe:
I was referring to this --- https://app.circleci.com/pipelines/github/reach-sh/reach-lang/5430/workflows/7f6c843b-495c-4698-bf4c-d9670ffae567/jobs/103278/tests

cbaunach:
> I was referring to this --- https://app.circleci.com/pipelines/github/reach-sh/reach-lang/5430/workflows/7f6c843b-495c-4698-bf4c-d9670ffae567/jobs/103278/tests

Ah okay, do you want me to try researching these failures and seeing if they can be fixed with adjustments to the examples or are you arguing that 1145 be abandoned due to these failures?

jeapostrophe:
I'm not saying it should be abandoned. I am just not dedicated to doing this and I have no idea what the status on them is... the programs should definitely work and whatever the issue with those tests is needs to be worked out somehow

cbaunach:
> I'm not saying it should be abandoned. I am just not dedicated to doing this and I have no idea what the status on them is... the programs should definitely work and whatever the issue with those tests is needs to be worked out somehow

Alright, I'll get to work on that; all three issues appear to relate to the ctcInfo, after the changes, being a BigNumber instead of a number so I'll see if there's any way to get that translated before it's handed back to the frontend

js/stdlib/ts/ALGO.ts
```diff
@@ -159,7 +159,7 @@ export type Backend = IBackend<AnyALGO_Ty> & {_Connectors: {ALGO: {
 type BackendViewsInfo = IBackendViewsInfo<AnyALGO_Ty>;
 type BackendViewInfo = IBackendViewInfo<AnyALGO_Ty>;
 
-export type ContractInfo = BigNumber;
```
jeapostrophe: We actually do want this and need to find the fix another way. If it stays `number`, then things will break once we get an application where JS numbers are imprecise

cbaunach: I agree, but without this it will break existing functionality; getInfo() will start returning a BigNumber on Algorand instead of a number and so many frontends that upgrade to this version will most likely run into difficulties because of that. For context, I discussed this with Dan Burton and he recommended against introducing breaking changes.

Taking all that into account, should I go back to returning a BigNumber (along with the required changes to the documentation if any) with the understanding that the release this is included in is accompanied by an announcement noting these changes to getInfo()?

jeapostrophe: The documentation has always said that it returns a UInt and that a UInt is either a BigNumber or a number, thus this change is not backwards incompatible. Furthermore, we KNOW this change is broken, because we KNOW that a number cannot hold the correct values.

cbaunach: Alright, I'll get it changed back and fix up those examples that were failing

cbaunach:
Due to this being a protected branch I cannot resolve the merge conflicts from my side

jeapostrophe:
I don't know what you mean by not being able to resolve conflicts. You can just rebase in the totally normal way and push to the branch

cbaunach:
I should have double-checked how rebasing is done first; going to close and delete this branch and start a new one

PR #598, opened 2022-02-19
Added formatWithDecimals to stdlib iface, test program, improved docs

erer1243:
Here's core-1146. I was a bit confused by the stdlib interfaces.ts hierarchy; I moved formatWithDecimals and its requisite functions to shared_impl.ts, then I added it to the Stdlib_User iface. Then, ETH_like and ALGO both import and re-export it. Let me know if there is a better place to put it.

Also added a small test program to show it in action.

### DIFF
--- a/js/stdlib/ts/ETH_like.ts
+++ b/js/stdlib/ts/ETH_like.ts
@@ -25,6 +25,7 @@ import {
   NotifySend,
   j2s,
   j2sf,
+  handleFormat,
 } from './shared_impl';
 import {
   bigNumberify,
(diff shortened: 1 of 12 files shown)


js/stdlib/ts/ETH_like.ts
```diff
@@ -74,7 +74,7 @@ export type {
 import type { // =>
   Stdlib_Backend
 } from './interfaces';
-import { setQueryLowerBound, getQueryLowerBound } from './shared_impl';
+import { setQueryLowerBound, getQueryLowerBound, formatWithDecimals } from './shared_impl';
```
jeapostrophe, line 77: Can you consider whether we should use this function to replace the one 1066 and unify the implementations of that too (basically copying `formatCurrency` from `ALGO.ts` with a different default)

erer1243:
I believe that's a good idea. Just committed a change with that done.

erer1243:
Just realized I accidentally committed my changes to the hie.yaml files as well. I can undo that if you'd like, or I can just leave it in since the master ones are incorrect.

jeapostrophe (approved):

jeapostrophe (approved):

PR #602, opened 2022-02-21
Keep all docker images declered on DEPS

vinmazzi:
## Summary
Declare all dependency docker images on the DEPS file.


examples/rps-7-rpc/client-py/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM       python:3.8-alpine
+FROM       {{PYTHON_IMAGE}}
```
jeapostrophe, line 1: I don't like that this is a `mo` rather than an ARG... mainly because I don't like having two

examples/rps-7-rpc/client-js/Dockerfile
```diff
@@ -1,6 +1,4 @@
-ARG NODE_VERSION
-ARG NODE_IMAGE=node:$NODE_VERSION
-FROM $NODE_IMAGE
+FROM {{NODE_IMAGE}}
```
jeapostrophe, line 1: As I said in the other, I prefer ARG over `mo`.

examples/rps-7-rpc/client-go/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM    golang:1.15-alpine
+FROM    {{GOLANG_IMAGE}}
```
jeapostrophe, line 1: ARG over `mo`

examples/rps-7-rpc/client-cs/Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine AS build
+FROM {{DOTNET_SDK_IMAGE}} AS build
```
jeapostrophe, line 1: ARG over `mo`

.circleci/config.pre.yml
```diff
@@ -72,6 +72,11 @@ commands:
               docker tag reachsh/$img:circleci-$CIRCLE_SHA1 reachsh/$img:$VERSION
             fi
           done
+  install_mo:
+    description: "Install mo"
+    steps:
+      - run: |
+          curl -sSL https://git.io/get-mo -o mo && chmod +x mo && sudo mv mo /usr/local/bin/
```
jeapostrophe, line 79: I believe this is here because the examples are using mo; I think that code path should either NOT use mo OR the `one.sh` directory should install it if it isn't there using the correct `MO_URL`

examples/one.sh
```diff
@@ -13,10 +13,21 @@ has_target() {
...
+  done
+}
+
 echo "$MODE $e"
 echo
 (
   cd "$e" || exit 1
+  dockerfiles=$(find . -iname Dockerfile)
+  if [ -n "${dockerfiles}" ]; then
+    mo_examples "${dockerfiles}"
```
jeapostrophe, line 29: I understand now why you did it this way... I think I still would prefer to not use mo, but you can push me a little bit if you think I'm wrong.

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #608, opened 2022-02-23
CORE-1162: `reach info`

mattaudesse:


examples/cli-version-compare/connector-required:algo+postgres#images-not-present/o.txt
```diff
@@ -0,0 +1,12 @@
+New images are available for:
+ * reachsh/devnet-algo   8e78dffa:  stable, 0, 0.1, 0.1.8, 34107655, 2022-01-31
```
jeapostrophe, line 2: One thing about these that I was noticing... this is telling you want is NEW but it doesn't show at all what you HAVE... I think that may be interesting... because you want to know just how old the thing you're on is

examples/cli-version-compare/run.sh
```diff
@@ -7,13 +7,23 @@ LDIR="$(for d in $DIRS; do echo "$d" | wc -m; done | sort -h | tail -n 1)"
 ROOT="$( ([ "$REACH_DOCKER" = 0 ] && pwd) || echo /app/src)"
 
 e () {
+  if [ -f "./$1/SKIP" ]; then
+    if [ -n "$CI" ]; then
+      echo "$1/SKIP: Skipping is only allowed during development."
+      exit 1
```
jeapostrophe, line 13: Very classy

jeapostrophe: Although I bet you'd prefer `./$1/ONLY` or something like that

hs/app/reach/Main.hs
```diff
@@ -1460,42 +1464,21 @@ upgrade :: Subcommand
 upgrade = command "upgrade" $ info f d
   where
     d = progDesc "Upgrade Reach"
-    f = pure . liftIO . exitWith $ ExitFailure 50
+    f = pure . liftIO $ putStrLn
+      "`reach upgrade` has been deprecated. Please use `reach info` and `reach update` instead."
```
jeapostrophe, line 1468: You should delete it entirely from the docs

jeapostrophe (changes requested):
I can't tell if you think you're done with this one, assuming no because still in Draft

jeapostrophe (commented):
Maaarvelous

examples/cli-version-compare/connector-required:algo+postgres#images-not-present/o.txt
mattaudesse: I've created CORE-1169 for this.

jeapostrophe (approved):

PR #609, opened 2022-02-23
Add source info to state errors

willghatch:
This is just a work in progress first draft, the first thing got got working end-to-end, not really ready for merge.  But I thought I would get some feedback about the Haskell code so far.  Is this more-or-less the direction you wanted this to go, @jeapostrophe ?  I added an extra field to PLProg for the map because I'm not really sure what the other things are yet.  None of them look like a good place to stuff this info, and all of them seemed to involve a similar footprint of boilerplate pattern matching (and reconstruction) editing.

Right now it prints something like `Error: Bob.payMe errored with Error: Expected state 2 at /home/wgh/s/mk/reach-lang/examples/api-call-fail/index.rsh:26:11:after expr stmt semicolon, got 1 at /home/wgh/s/mk/reach-lang/examples/api-call-fail/index.rsh:26:11:after expr stmt semicolon`, which obviously needs to be cleaned up, still needs the static language stack so it can disambiguate different states with the same source location.

### DIFF
--- a/hs/src/Reach/APICut.hs
+++ b/hs/src/Reach/APICut.hs
@@ -346,8 +346,8 @@ apc hc eWho = \case
   p -> return p
 
 apicut :: PLProg -> IO PLProg
-apicut (PLProg at plo dli dex epps cp) = do
+apicut (PLProg at plo dli dex ssm epps cp) = do
   let EPPs apis em = epps
   em' <- mapWithKeyM (apc plo) em
   let epps' = EPPs apis em'
-  return $ PLProg at plo dli dex epps' cp
+  return $ PLProg at plo dli dex ssm epps' cp
--- a/hs/src/Reach/Backend/JS.hs
+++ b/hs/src/Reach/Backend/JS.hs
@@ -1008,7 +1008,7 @@ reachBackendVersion :: Int
 reachBackendVersion = 10
 
 jsPIProg :: ConnectorResult -> PLProg -> App Doc
-jsPIProg cr (PLProg _ _ dli dexports (EPPs {..}) (CPProg _ vi _ devts _)) = do
+jsPIProg cr (PLProg _ _ dli dexports ssm (EPPs {..}) (CPProg _ vi _ devts _)) = do
   let DLInit {..} = dli
   let preamble =
         vsep
@@ -1022,6 +1022,7 @@ jsPIProg cr (PLProg _ _ dli dexports (EPPs {..}) (CPProg _ vi _ devts _)) = do
   partsp <- mapM (uncurry (jsPart dli)) $ M.toAscList epps_m
   cnpsp <- mapM (uncurry jsCnp) $ HM.toList cr
   let connMap = M.fromList [(name, "_" <> pretty name) | name <- HM.keys cr]
+  ssmDoc <- mapM (\x -> jsAssertInfo (fst x) (snd x) Nothing) ssm
   exportsp <- jsExports dexports
   viewsp <-
     local (\e -> e {ctxt_maps = dli_maps}) $
@@ -1038,7 +1039,7 @@ jsPIProg cr (PLProg _ _ dli dexports (EPPs {..}) (CPProg _ vi _ devts _)) = do
           mempty
           epps_apis
   eventsp <- jsEvents devts
-  return $ vsep $ [preamble, exportsp, eventsp, viewsp, mapsp] <> partsp <> cnpsp <> [jsObjectDef "_Connectors" connMap, jsObjectDef "_Participants" partMap, jsObjectDef "_APIs" apiMap]
+  return $ vsep $ [preamble, exportsp, eventsp, viewsp, mapsp] <> partsp <> cnpsp <> [jsObjectDef "_stateSourceMap" ssmDoc, jsObjectDef "_Connectors" connMap, jsObjectDef "_Participants" partMap, jsObjectDef "_APIs" apiMap]
 
 backend_js :: Backend
 backend_js outn crs pl = do
--- a/hs/src/Reach/EPP.hs
+++ b/hs/src/Reach/EPP.hs
@@ -163,6 +163,7 @@ data BEnv = BEnv
   , be_handlers :: IORef (M.Map Int (CApp CHandler))
   , be_flowr :: IORef FlowInput
   , be_more :: IORef Bool
+  , be_stateToSrcMap :: IORef StateSrcMap
   , be_loop :: Maybe (Int, Int)
   , be_output_vs :: IORef [DLVar]
   , be_toks :: [DLArg]
@@ -527,7 +528,7 @@ be_c = \case
           return (wrap k'c, wrap k'l)
     csm' <- mapM go csm
     return $ (,) (CT_Switch at ov <$> mapM fst csm') (ET_Switch at ov <$> mapM snd csm')
-  LLC_FromConsensus at1 _at2 s -> do
+  LLC_FromConsensus at1 _at2 fs s -> do
     this <- newSavePoint "fromConsensus"
     views <- asks be_views
     (more, s'l) <-
@@ -559,6 +560,8 @@ be_c = \case
     fg_saves this
     let cm = CT_From at1 this <$> mkfrom_info ce_readSave
     let lm = ET_FromConsensus at1 this <$> mkfrom_info ee_readSave <*> s'l
+    stateToSrcMap <- asks be_stateToSrcMap
+    liftIO $ modifyIORef stateToSrcMap $ M.insert this (at1, fs)
     return $ (,) cm lm
   LLC_While at asn _inv cond body k -> do
     let DLBlock cond_at cond_fs cond_l cond_a = cond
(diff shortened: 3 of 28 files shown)


hs/src/Reach/APICut.hs
```diff
@@ -338,8 +338,8 @@ apc hc eWho = \case
   p -> return p
 
 apicut :: PLProg -> IO PLProg
-apicut (PLProg at plo dli dex epps cp) = do
+apicut (PLProg at plo dli dex epps cp ssm) = do
```
jeapostrophe, line 341: I have a preference to put "extra" stuff before `epps`. Probably `PLProg` should become a record soon and these should be given nice names

hs/src/Reach/Backend/JS.hs
```diff
@@ -1022,6 +1022,7 @@ jsPIProg cr (PLProg _ _ dli dexports (EPPs {..}) (CPProg _ vi _ devts _)) = do
   partsp <- mapM (uncurry (jsPart dli)) $ M.toAscList epps_m
   cnpsp <- mapM (uncurry jsCnp) $ HM.toList cr
   let connMap = M.fromList [(name, "_" <> pretty name) | name <- HM.keys cr]
+  let ssmDoc = M.map (pretty . (\s -> "\"" ++ s ++ "\"") . show) ssm
```
jeapostrophe, line 1025: There's a function in this file for formatting "assertion info" that you should use

hs/src/Reach/EPP.hs
```diff
@@ -163,6 +163,7 @@ data BEnv = BEnv
   , be_handlers :: IORef (M.Map Int (CApp CHandler))
   , be_flowr :: IORef FlowInput
   , be_more :: IORef Bool
+  , be_stateToSrcMap :: IORef (M.Map Int SrcLoc)
```
jeapostrophe, line 166: Use the type you defined in PL

hs/src/Reach/EPP.hs
```diff
@@ -559,7 +560,10 @@ be_c = \case
     fg_saves this
     let cm = CT_From at1 this <$> mkfrom_info ce_readSave
     let lm = ET_FromConsensus at1 this <$> mkfrom_info ee_readSave <*> s'l
-    return $ (,) cm lm
+    do
```
jeapostrophe, line 563: No need for this do... you're already in a do

js/stdlib/ts/ALGO.ts
```diff
@@ -1309,7 +1309,9 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
         debug('getState');
         return await getState_(getC, (vibna:BigNumber) => {
           if ( vibne.eq(vibna) ) { return vtys; }
-          throw Error(`Expected state ${vibne}, got ${vibna}`);
+          const expectedLoc = bin._stateSourceMap[vibne.toNumber().valueOf()];
+          const actualLoc = bin._stateSourceMap[vibne.toNumber().valueOf()];
+          throw Error(`Expected state ${vibne} at ${expectedLoc}, got ${vibna} at ${actualLoc}`);
```
jeapostrophe, line 1314: There's a function in shared_impl (or maybe backend... look at `assert`) for printing these things.

I prefer: "Expected the DApp to be at the commit() on line XYZ, but it is at the commit() on line ABC."

If you add the stack traces (which would require doing something to the `DLS_FromConsensus` statement) then you'd put them after... like "\nStack trace of expected: ...\nStack trace of actual: ..."

jeapostrophe (changes requested):

willghatch:
Using the functions you recommend for formatting, the message ends up like this:

```
Error: Bob.payMe errored with Error: Expected the Dapp to be at commit() at : Module
    at ./index.rsh:26:11:after expr stmt semicolon, Dapp was actually at commit() at : Module
    at ./index.rsh:16:11:after expr stmt semicolon
```

I've now threaded a `[SLCtxtFrame]` through to the place where the message is created.  But now I see that my test program (the `api-call` program with the order rearranged in the frontend) has an empty stack for each `commit()`.  Also, I'm not certain that's enough to uniquely identify states, and I'm a little confused.  In the output `mjs` for the api-call program, there are 5 states where states 4 and 5 both correspond to the last commit.  Why are there 2 states there, and is this something that can happen elsewhere?

hs/src/Reach/EPP.hs
willghatch: If one do is good, two must be better.

js/stdlib/ts/ETH_like.ts
```diff
@@ -543,7 +543,11 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account>
         const [ vibna, vsbs ] = await ethersC["_reachCurrentState"]();
         debug(`getState`, { vibne, vibna, vsbs });
         if ( ! vibne.eq(vibna) ) {
-          throw Error(`expected state ${vibne}, got ${vibna}`);
+          const formatLoc = (l:BigNumber) =>
+            formatAssertInfo(bin._stateSourceMap[l.toNumber()]);
+          const e = formatLoc(vibne);
+          const a = formatLoc(vibna);
+          throw Error(`Expected the Dapp to be at commit() at ${e}, Dapp was actually at commit() at ${a}`);
```
jeapostrophe, line 550: You should make a function in `shared_backend` and use it in both of these

js/stdlib/ts/ALGO.ts
```diff
@@ -1309,9 +1310,11 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
...
         return await getState_(getC, (vibna:BigNumber) => {
           if ( vibne.eq(vibna) ) { return vtys; }
-          const expectedLoc = bin._stateSourceMap[vibne.toNumber().valueOf()];
-          const actualLoc = bin._stateSourceMap[vibne.toNumber().valueOf()];
-          throw Error(`Expected state ${vibne} at ${expectedLoc}, got ${vibna} at ${actualLoc}`);
+          const formatLoc = (l:BigNumber) =>
+            formatAssertInfo(bin._stateSourceMap[l.toNumber()]);
+          const e = formatLoc(vibne);
+          const a = formatLoc(vibna);
+          throw Error(`Expected the Dapp to be at commit() at ${e}, Dapp was actually at commit() at ${a}`);
```
jeapostrophe, line 1317: DApp

jeapostrophe: I think the print out with the stack in between is really ugly and you should try to make it so the stacks come after (using the pieces of formatAssertInfo)

jeapostrophe (changes requested):

jeapostrophe:
> I've now threaded a `[SLCtxtFrame]` through to the place where the message is created.

Nice!

> But now I see that my test program (the `api-call` program with the order rearranged in the frontend) has an empty stack for each `commit()`.

That's fine. Make sure it's not ugly when there's no stack

> Also, I'm not certain that's enough to uniquely identify states, and I'm a little confused. In the output `mjs` for the api-call program, there are 5 states where states 4 and 5 both correspond to the last commit. Why are there 2 states there, and is this something that can happen elsewhere?

During linearization, we duplicate the continuation, because DL has no stack so we turn something like

```
if ( c ) { t } else { f }
k
```

into

```
if ( c ) { t; k } else { f; k }
```

This will be confusing if we're in a commit in the first copy but expect to be in the second one. Maybe we can help that situation by leaving the numbers and adding this new information as an extra piece of metadata. Like:

```
Error: Bob.payMe errored with Error: Expected the DApp to be in state 6; but it was actually in state 7.

State 6 corresponds to the commit() at : Module
    at ./index.rsh:26:11:after expr stmt semicolon
State 7 corresponds to the commit() at : Module
    at ./index.rsh:16:11:after expr stmt semicolon
```

We could even explicitly check when the two srclocs are the same and add

```
(This means that the commit() is the continuation of an effect.)
```

js/stdlib/ts/shared_backend.ts
```diff
@@ -72,6 +72,17 @@ export const formatAssertInfo = (ai:any = null) => {
   return msg;
 };
 
+export const APIStateMismatchError = (bin: any, es: BigNumber, as: BigNumber) : Error => {
```
willghatch, line 75: Would you rather this be in `shared_impl.ts` so that I can type `bin` as `IBackend` instead of `any`?

jeapostrophe: Oh, yea that's what I meant, sorry I'm dumb.

jeapostrophe (commented):
.

willghatch:
Ok, I moved the error constructor into `shared_impl` so it can be fully typed, slightly fussed with naming, and squashed it.

willghatch:
Also, does a comment "." mean something, or was it just an accident?

jeapostrophe:
> Also, does a comment "." mean something, or was it just an accident?

It means "Github requires that I put in a comment, but I have nothing to say, so I want to reference old school Matthew"

jeapostrophe:
This is good to go, but there's a conflict so it can't be merged :/

willghatch:
Merge conflict fixed.

jeapostrophe (approved):

PR #610, opened 2022-02-23
Initial Commit

vinmazzi:
## Summary

Add a script to execute solc without the need of having it installed to the local system.

### DIFF
--- a/scripts/solc.sh
+++ b/scripts/solc.sh
@@ -1,0 +1,9 @@
+. ../DEPS
+docker run --rm \
+    -v "$PWD:/mnt" \
+    -u "$(id -ru):$(id -rg)" \
+    -w /mnt \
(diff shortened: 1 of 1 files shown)


jeapostrophe (changes requested):
I think this crashed sh-lint and looks like it has some obvious problems, like no `#!` and not quoting `$SOLC_IMAGE` and not being robust to the location. Also, by using the `SOLC_IMAGE`, it means that the image might not actually be present... it should use one of the Reach images

vinmazzi:
> 
If the image is not present locally docker will download it.
It would also be a lighter than downloading our haskell docker image.

jeapostrophe (approved):

jeapostrophe:
Cool!

PR #613, opened 2022-02-23
check port instead of just wait 60

DanBurton:


jeapostrophe (approved):

jeapostrophe:
I just pressed approve, but I don't know if the test with the `done` error is accurate or the force push overwrote it

PR #616, opened 2022-02-24
test `api-full`

chikeabuah:


examples/simulator-remote/client.mjs
```diff
@@ -0,0 +1,47 @@
+import * as c from '@reach-sh/simulator-client';
```
jeapostrophe, line 1: This push depends on the last one

jeapostrophe (changes requested):

PR #629, opened 2022-02-26
replace js annotation typeclass method

chikeabuah:
this didn't seem to work for my particular issue, but i wrote the code, here it is.

### DIFF
--- a/hs/src/Reach/JSUtil.hs
+++ b/hs/src/Reach/JSUtil.hs
@@ -177,7 +177,10 @@ a2sp = JSSemi
 
 class HasJSAnnot a where
   jsa :: a -> JSAnnot
-  rjsa :: a -> JSAnnot -> a
+  rjsa :: JSAnnot -> a -> a
+
+rjsaJSL :: HasJSAnnot a => JSAnnot -> JSCommaList a -> JSCommaList a
+rjsaJSL a' y = (rjsaJSL a' y)
 
 instance HasJSAnnot JSAssignOp where
   jsa = \case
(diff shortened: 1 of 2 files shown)


hs/src/Reach/JSUtil.hs
```diff
@@ -176,6 +177,7 @@ a2sp = JSSemi
 
 class HasJSAnnot a where
   jsa :: a -> JSAnnot
+  rjsa :: a -> JSAnnot -> a
```
jeapostrophe, line 180: I'd give this the type `JSAnnot -> a -> a` so you can use `\case`

hs/src/Reach/JSUtil.hs
```diff
@@ -191,6 +193,19 @@ instance HasJSAnnot JSAssignOp where
     JSBwAndAssign a -> a
     JSBwXorAssign a -> a
     JSBwOrAssign a -> a
+  rjsa b a' = case b of
```
jeapostrophe, line 196: `\case` is so much more pretty

hs/src/Reach/JSUtil.hs
```diff
@@ -267,19 +317,63 @@ instance HasJSAnnot JSExpression where
...
     JSYieldFromExpression a _ _ -> a
+  rjsa b a' = case b of
+    JSIdentifier _a x -> JSIdentifier a' x
+    JSDecimal _a x -> JSDecimal a' x
+    JSLiteral _a x -> JSLiteral a' x
+    JSHexInteger _a x -> JSHexInteger a' x
+    JSOctal _a x -> JSOctal a' x
+    JSStringLiteral _a x -> JSStringLiteral a' x
+    JSRegEx _a x -> JSRegEx a' x
+    JSArrayLiteral _a x y -> JSArrayLiteral a' x y
```
jeapostrophe, line 328: You're going to have to recur inside everything; also `y` is ALSO an annotation (it's the right `]`)

jeapostrophe (changes requested):
It's not working because you're not recurring

PR #637, opened 2022-03-01
Core 1167 get default account error msg

erer1243:
Updated error message to say `Providers created by environment ... Calling setProviderByEnv or setProviderByName removes this capability. Try removing calls to those functions.`


erer1243:
accidentally pushed on core-1147 branch whoops

PR #641, opened 2022-03-01
CORE-1145 | fetch balances with BigInt precision

cbaunach:
This is primarily composed of adjustment Jay has made to ALGO.ts to allow for BigInt precision on the results of balanceOf and other functions using setIntEncoding. 

Whereas before a token with the maximum supply would return 18446744073709552000 (since it could only get up to number precision) it will now return 18446744073709551615 (since it now has bigint precision). 

Jay also created an example to test and prove this functionality of the balanceOf function.

### DIFF
--- a/examples/api-raw/index.mjs
+++ b/examples/api-raw/index.mjs
@@ -53,7 +53,7 @@ const run = async ({fn, isRaw}) => {
       exp = [ x, ...args ];
       const fnM = `${isView ? 'V' : 'P'}_${fn}`;
 
-      const ctcInfo = await ctcA.getInfo();
+      const ctcInfo = stdlib.bigNumberToNumber(await ctcA.getInfo());
       const ctc = accB.contract(backend, ctcInfo);
       const ABI = ctc.getABI(true);
       console.log(ABI);
--- a/examples/api-raw/index.mjs
+++ b/examples/api-raw/index.mjs
@@ -53,7 +53,7 @@ const run = async ({fn, isRaw}) => {
       exp = [ x, ...args ];
       const fnM = `${isView ? 'V' : 'P'}_${fn}`;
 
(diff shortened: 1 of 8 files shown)


examples/ctc-address/index.mjs
```diff
@@ -14,7 +14,7 @@ const stdlib = loadStdlib(process.env);
 
   const common = (ctcMe) => ({
     showCtcInfo: async (ctc) => {
-      const info = await ctcMe.getInfo();
+      const info = stdlib.bigNumberToNumber(await ctcMe.getInfo());
       console.log(`Contract Info From Reach      : ${ctc}`);
       console.log(`Contract Info From ctc.getInfo: ${info}`);
       stdlib.assert(ctc == info);
```
jeapostrophe, line 20: I believe the real problem is that `==` is invalid and we should do something like turning them into JSON-ified strings and comparing those

jeapostrophe: Or, we should add `stdlib.contract_eq`

examples/api-raw/index.mjs
```diff
@@ -53,7 +53,7 @@ const run = async ({fn, isRaw}) => {
       exp = [ x, ...args ];
       const fnM = `${isView ? 'V' : 'P'}_${fn}`;
 
-      const ctcInfo = await ctcA.getInfo();
+      const ctcInfo = stdlib.bigNumberToNumber(await ctcA.getInfo());
```
jeapostrophe, line 56: You should adapt JUST the `ALGO` case

jeapostrophe (changes requested):

jeapostrophe (approved):

jeapostrophe:
Great work @cbaunach !

PR #642, opened 2022-03-02
Better Labels and SMT Errors

chrisnevers:
* **Freshen**: Keep track of labels better.
* **SMT**: Associate unbound variables from DLRecv with `SMTModel O_Publish`

### DIFF
--- a/hs/src/Reach/Freshen.hs
+++ b/hs/src/Reach/Freshen.hs
@@ -78,11 +78,13 @@ instance Freshen DLVar where
     -- only subsequent uses of the DLVar will have a label associated with it.
     -- If we use the DLVar from the assignment everywhere, we will lose labels, which is needed for nice SMT errors.
     -- If we can ever recover the label information, replace the key.
-    case (mv', v) of
-      (Just (DLVar at Nothing t i), DLVar _ lab@(Just {}) _ _) ->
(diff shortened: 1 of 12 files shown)


hs/src/Reach/Freshen.hs
```diff
@@ -73,7 +73,16 @@ instance Freshen DLVar where
...
     rho <- liftIO $ readIORef fRho
-    return $ fromMaybe v $ M.lookup v rho
+    let mv' = M.lookup v rho
+    -- Often, an expression will be assigned to a DLVar with no label, and
+    -- only subsequent uses of the DLVar will have a label associated with it.
+    -- If we use the DLVar from the assignment everywhere, we will lose labels, which is needed for nice SMT errors.
+    -- If we can ever recover the label information, replace the key.
+    case (mv', v) of
+      (Just (DLVar at Nothing t i), DLVar _ lab@(Just {}) _ _) ->
+        liftIO $ modifyIORef fRho $ M.insert v $ DLVar at lab t i
```
jeapostrophe, line 83: I believe that this will update the `v` binding but not actually use it for THIS return

chrisnevers: ah right, thanks for catching that

jeapostrophe (approved):

PR #646, opened 2022-03-02
adhere more closely to arc-6 updates

DanBurton:
Added the ability to do stuff like this:

```js
loadStdlib({
  REACH_CONNECTOR_MODE: 'ALGO',
  ALGO_GENESIS_ID: 'testnet-v1.0',
  ALGO_GENESIS_HASH: 'asdf',
  ALGO_ACCOUNT: 'deadbeef',
});
```

The spirit of [arc-6](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0006.md) is for these to guide the wallet in prompting the user to accept a connection to that specific network (if specified) with that specific account (if specified). While our wallet doesn't do that, it at least conforms to the part of the spec that says if the wallet doesn't support that specific thing then it should throw an Error.

The ability to ask for a specific account is usually not what people will want to do, so the docs say it "should usually be left unspecified."

`ALGO_GENESIS_ID` correlates to what the `network` option used to be, but they changed arc-6 to name it genesisID, and they added genesisHash, so I updated our stuff accordingly. Apparently genesisID does not uniquely identify the network but genesisHash does.

I also rearranged `process.env`-related stuff in a way that I think is better. No more `updateProcessEnv` on the fly. `walletFallback` and `setWalletFallback` now rely only on their arguments, but the created wallet exposes `_env` (undocumented) so that `makeProviderByWallet` can extract useful info about it.

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -128,7 +128,13 @@ const reach = await loadStdlib(process.env);
 
 Or construct a custom object that has all of the environment keys and fields you need.
 
-As a special case, you may instead pass in the string `'ETH'` or the string `'ALGO'`
+On Algorand, this object may include the keys `ALGO_GENESIS_ID` or `ALGO_GENESIS_HASH` (or both),
+which allows you to request that the user's [arc-6 compliant wallet](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0006.md) connect to a specific network.
+When left unspecified, it allows the user to select one of their wallet's supported networks.
+This object may also include the key `ALGO_ACCOUNT`,
+which allows you to request the use of a specific account from the user's arc-6 compliant wallet by address. This should usually be left unspecified, which allows the user to instead select their preferred account.
+
+As a special case, you may instead pass in the string `'ETH'`, `'ALGO'`, or `'CFX'`,
 to select the desired connector directly.
 
 ---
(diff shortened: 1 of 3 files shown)


docs/src/frontend/index.md
```diff
@@ -128,7 +128,13 @@ const reach = await loadStdlib(process.env);
 
 Or construct a custom object that has all of the environment keys and fields you need.
 
-As a special case, you may instead pass in the string `'ETH'` or the string `'ALGO'`
+On Algorand, this object may include the keys `ALGO_GENESIS_ID` or `ALGO_GENESIS_HASH` (or both),
+which allows you to request that the user's [arc-6 compliant wallet](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0006.md) connect to a specific network.
+When left unspecified, it allows the user to select one of their wallet's supported networks.
+This object may also include the key `ALGO_ACCOUNT`,
+which allows you to request the use of a specific account from the user's arc-6 compliant wallet by address. This should usually be left unspecified, which allows the user to instead select their preferred account.
```
jeapostrophe, line 135: Stylize as `ARC-6`

docs/src/frontend/index.md
```diff
@@ -128,7 +128,13 @@ const reach = await loadStdlib(process.env);
 
 Or construct a custom object that has all of the environment keys and fields you need.
 
-As a special case, you may instead pass in the string `'ETH'` or the string `'ALGO'`
+On Algorand, this object may include the keys `ALGO_GENESIS_ID` or `ALGO_GENESIS_HASH` (or both),
+which allows you to request that the user's [arc-6 compliant wallet](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0006.md) connect to a specific network.
```
jeapostrophe, line 132: Here too

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #647, opened 2022-03-02
make goal-devnet script respect environment variable that goal uses

willghatch:
The `goal` program apparently checks the $TMPDIR environment variable,
defaulting to `/tmp` if it's unset.  This change makes our script do
the same instead of always mounting `/tmp` without checking.

This also switches the script to use bash instead of `/bin/sh`,
because I think the `:-` variable default operation is a bashism.

### DIFF
--- a/scripts/goal-devnet
+++ b/scripts/goal-devnet
@@ -1,9 +1,10 @@
 #!/bin/sh
+tmp="${TMPDIR:-/tmp}"
 exec docker run \
   --rm \
   -u "$(id -ru):$(id -rg)" \
-  --volume "$PWD:/app" \
-  --volume "/tmp:/tmp" \
+  --volume "${PWD}:/app" \
+  --volume "${tmp}:${tmp}" \
   --workdir /app \
   --entrypoint /bin/goal \
   reachsh/devnet-algo \


willghatch:
I said `goal` respects the environment variable.  Now that I think about it, I think `reachc` is the one that uses it, not `goal`.  `reachc` just passes the path to `goal`.

scripts/goal-devnet
```diff
@@ -1,9 +1,10 @@
-#!/bin/sh
+#!/usr/bin/env bash
+tmp="${TMPDIR:-/tmp}"
 exec docker run \
   --rm \
   -u "$(id -ru):$(id -rg)" \
   --volume "$PWD:/app" \
-  --volume "/tmp:/tmp" \
+  --volume "$tmp:$tmp" \
```
jeapostrophe, line 7: sh-link doesn't care? I'd do `${tmp}` because I'm paranoid

scripts/goal-devnet
```diff
@@ -1,9 +1,10 @@
-#!/bin/sh
+#!/usr/bin/env bash
+tmp="${TMPDIR:-/tmp}"
 exec docker run \
   --rm \
   -u "$(id -ru):$(id -rg)" \
   --volume "$PWD:/app" \
```
jeapostrophe, line 6: Same here... surprised we got away with this

jeapostrophe (changes requested):

willghatch:
I pushed a version that uses `${...}` everywhere to be paranoid. :)

But now that I look more, `$(...)` is definitely a bashism, so this script is working with `/bin/sh` by virtue of the fact that it's almost always actually linked to bash.

jeapostrophe:
> I pushed a version that uses `${...}` everywhere to be paranoid. :)
> 
> But now that I look more, `$(...)` is definitely a bashism, so this script is working with `/bin/sh` by virtue of the fact that it's almost always actually linked to bash.

No, `$()` is not bash. We use `shellcheck` to check these things... https://circleci.com/gh/reach-sh/reach-lang/105831?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

willghatch:
OK, you're right.  I don't know what shell I used then that didn't allow `$()`.  But for what it's worth, I recommend that we change scripts that have `#!/bin/bash` to use `#!/usr/bin/env bash` instead, bash really isn't always under `/bin`.

But for the task at hand, should I push a version that uses `/bin/sh` and let shellcheck decide whether `${VAR:-DEFAULT}` is valid Posix shell?

jeapostrophe:
`/bin/sh`

willghatch:
I guess I'll take that as a yes.

willghatch:
(My comment about /bin/bash is not in reference to this script, but to other scripts.  A quick grep says we have 13 scripts that use /bin/bash.)

jeapostrophe:
I'm willing to change all the `/bin/bash` to `/usr/bin/env bash` but I don't really understand why it is better, but I trust you

willghatch:
I'll probably make another PR for that sometime.  `/bin/bash` is the most common location for bash, but it's not 100% standardized.  I've used some systems where it's not there.  Generally I think the only paths that are really safe for `#!` are `/bin/sh` and `/usr/bin/env`, everything else can move.  The ones that most obviously come to mind are NixOS and Guix, which... without going into a special environment more-or-less enforce the view that only those two paths are reliable.  But I've seen other systems put bash at places like `/usr/bin/bash` with no symlink to `/bin/bash`.  I think FreeBSD (or some BSD that I tried at some point) puts it somewhere like `/usr/local/bin/bash`, but I might be misremembering.  Some of those might be systems we don't support, but switching to `/usr/bin/env bash` never hurts.

willghatch:
Is there something blocking this?  It would be convenient.

jeapostrophe (approved):

PR #648, opened 2022-03-03
make code that prints keywords more intuitive

hamirmahal:


willghatch:
👍

DanBurton (approved):
It's potato potato to me.  ¯\\\_(ツ)_/¯

willghatch:
That phrase just doesn't work as well when written as when spoken.  Agreed, it's trivial, but I like the change anyway. :)

jeapostrophe (approved):

PR #650, opened 2022-03-04
CORE-1172|1173: `reach compile` optimizations

mattaudesse:
My observation while testing was that this PR shaves anywhere from 0.6 - 2.0+ seconds off `reach compile` times - it does so by forking logs to another UNIX process in a re-used Docker container. The mean seems to be around 1.1 seconds saved.

`reach down` has also been modified to clear out the new background `reachc` containers.

Running `docker container stats`, even with many concurrent instances of `reachsh/reach` launched from different project directories and with different `REACH_VERSION` targets, indicates the containers use almost no system resources at all.

### DIFF
--- a/hs/app/reach/Main.hs
+++ b/hs/app/reach/Main.hs
@@ -1058,17 +1058,29 @@ compile = command "compile" $ info f d
             $reachc_dev
           fi
         else
-          docker run \
-            --rm \
-            --volume "$$PWD:/app" \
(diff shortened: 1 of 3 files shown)


jeapostrophe (commented):
I don't get what's going on here but it doesn't look insane

mattaudesse:
> I don't get what's going on here but it doesn't look insane

I won't not take that as a good sign :-)

PR #651, opened 2022-03-05
CORE-1153 Print srclocs w/ remote object and step arg length warnings

erer1243:
Adds SrcLoc printing to warnings about using remote objects on ALGO and about having steps that take too large of arguments. Let me know if you'd like it phrased differently.

Example of warning phrasing here:
image


erer1243:
Also let me know if there is a better way to make an example that causes these errors, ha ha. Don't worry I didn't type out all those var names...

jeapostrophe (changes requested):
You need to add the `index.txt` file

jeapostrophe:
Otherwise good

PR #659, opened 2022-03-07
CORE-1154 Examples of posthumous view and event behavior

erer1243:
posthumous-events example:
image

posthumous-views example:
image

### DIFF
--- a/examples/posthumous-events/index.mjs
+++ b/examples/posthumous-events/index.mjs
@@ -5,14 +5,17 @@ const stdlib = loadStdlib(process.env);
 const startingBalance = stdlib.parseCurrency(100);
 const [ accAlice, accObserver ] = await stdlib.newTestAccounts(2, startingBalance);
 const ctcAlice = accAlice.contract(backend);
-
-// After this Promise finishes, the contract has exited
-await backend.Alice(ctcAlice, {});
-
-// 'Posthumous' code, looking at events emitted while the contract was alive
 const ctcObserver = accObserver.contract(backend, ctcAlice.getInfo());
 
+// After this Promise finishes, the contract has exited
+await Promise.all([
+  ctcAlice.p.Alice({}),
+  ctcObserver.p.Observer({}),
+]);
+
+// 'Posthumous' code, looking at events emitted while the contract was alive
+const ctcEvents = accObserver.contract(backend, ctcAlice.getInfo());
 for (var n = 0; n < 10; n++) {
-  const ev = await ctcObserver.events.Numbers.number.next();
+  const ev = await ctcEvents.events.Numbers.number.next();
   console.log(`Number event #${ev.what}`);
 }
--- a/examples/posthumous-events/index.mjs
+++ b/examples/posthumous-events/index.mjs
@@ -3,15 +3,16 @@ import * as backend from './build/index.main.mjs';
 const stdlib = loadStdlib(process.env);
 
 const startingBalance = stdlib.parseCurrency(100);
-const [ accAlice ] = await stdlib.newTestAccounts(1, startingBalance);
-const ctc = accAlice.contract(backend);
+const [ accAlice, accObserver ] = await stdlib.newTestAccounts(2, startingBalance);
+const ctcAlice = accAlice.contract(backend);
 
 // After this Promise finishes, the contract has exited
-await backend.Alice(ctc, {});
+await backend.Alice(ctcAlice, {});
 
 // 'Posthumous' code, looking at events emitted while the contract was alive
-const numMonPromise = 
-  ctc.events.Numbers.number.monitor(ev => console.log(`Number event #${ev.what}`));
+const ctcObserver = accObserver.contract(backend, ctcAlice.getInfo());
 
-await Promise.race([numMonPromise, new Promise(r => setTimeout(r, 2000))]);
-process.exit(0);
+for (var n = 0; n < 10; n++) {
+  const ev = await ctcObserver.events.Numbers.number.next();
+  console.log(`Number event #${ev.what}`);
+}
--- a/examples/posthumous-events/index.mjs
+++ b/examples/posthumous-events/index.mjs
@@ -1,0 +1,17 @@
+import {loadStdlib} from '@reach-sh/stdlib';
+import * as backend from './build/index.main.mjs';
+const stdlib = loadStdlib(process.env);
+
+const startingBalance = stdlib.parseCurrency(100);
+const [ accAlice ] = await stdlib.newTestAccounts(1, startingBalance);
+const ctc = accAlice.contract(backend);
+
+// After this Promise finishes, the contract has exited
+await backend.Alice(ctc, {});
+
+// 'Posthumous' code, looking at events emitted while the contract was alive
+const numMonPromise = 
+  ctc.events.Numbers.number.monitor(ev => console.log(`Number event #${ev.what}`));
+
+await Promise.race([numMonPromise, new Promise(r => setTimeout(r, 2000))]);
+process.exit(0);
--- a/examples/posthumous-views/index.mjs
+++ b/examples/posthumous-views/index.mjs
@@ -1,0 +1,22 @@
+import {loadStdlib} from '@reach-sh/stdlib';
+import * as backend from './build/index.main.mjs';
+const stdlib = loadStdlib(process.env);
+
+const startingBalance = stdlib.parseCurrency(100);
+const [ accAlice ] = await stdlib.newTestAccounts(1, startingBalance);
+const ctc = accAlice.contract(backend);
+
+// After this Promise finishes, the contract has exited
+await backend.Alice(ctc, {
+  observeNumber: async () => {
+    const maybeN = await ctc.views.Number.number();
+    console.log(`Alice observes the current number as ${maybeN[1]}`);
+  },
+});
+
+// 'Posthumous' code, looking at view set while the contract was alive
+
+const maybeN = await ctc.views.Number.number();
+if (maybeN[0] !== 'None') 
+  throw Error("view set after ctc exited");
+console.log("After the ctc exited, the current number is None")
--- a/examples/posthumous-views/index.mjs
+++ b/examples/posthumous-views/index.mjs
@@ -3,19 +3,23 @@ import * as backend from './build/index.main.mjs';
 const stdlib = loadStdlib(process.env);
 
 const startingBalance = stdlib.parseCurrency(100);
-const [ accAlice ] = await stdlib.newTestAccounts(1, startingBalance);
-const ctc = accAlice.contract(backend);
+const [accAlice, accBob] = await stdlib.newTestAccounts(2, startingBalance);
+const ctcAlice = accAlice.contract(backend);
+const ctcBob = accBob.contract(backend, ctcAlice.getInfo());
 
 // After this Promise finishes, the contract has exited
-await backend.Alice(ctc, {
-  observeNumber: async () => {
-    const maybeN = await ctc.views.Number.number();
-    console.log(`Alice observes the current number as ${maybeN[1]}`);
-  },
-});
+await Promise.all([
+  backend.Alice(ctcAlice, {
+    observeNumber: async () => {
+      const maybeN = await ctcAlice.views.Number.number();
+      console.log(`Alices observes the current number as ${maybeN[1]}`);
+    },
+  }),
+  backend.Bob(ctcBob, {}),
+]);
 
 // 'Posthumous' code, looking at view set while the contract was alive
-const maybeN = await ctc.views.Number.number();
+const maybeN = await ctcAlice.views.Number.number();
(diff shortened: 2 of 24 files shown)


examples/posthumous-events/index.mjs
```diff
@@ -0,0 +1,17 @@
...
+const ctc = accAlice.contract(backend);
+
+// After this Promise finishes, the contract has exited
+await backend.Alice(ctc, {});
+
+// 'Posthumous' code, looking at events emitted while the contract was alive
+const numMonPromise = 
+  ctc.events.Numbers.number.monitor(ev => console.log(`Number event #${ev.what}`));
+
+await Promise.race([numMonPromise, new Promise(r => setTimeout(r, 2000))]);
```
jeapostrophe, line 16: No timeout

examples/posthumous-events/index.mjs
```diff
@@ -0,0 +1,17 @@
...
+const startingBalance = stdlib.parseCurrency(100);
+const [ accAlice ] = await stdlib.newTestAccounts(1, startingBalance);
+const ctc = accAlice.contract(backend);
+
+// After this Promise finishes, the contract has exited
+await backend.Alice(ctc, {});
+
+// 'Posthumous' code, looking at events emitted while the contract was alive
+const numMonPromise = 
+  ctc.events.Numbers.number.monitor(ev => console.log(`Number event #${ev.what}`));
```
jeapostrophe, line 14: This is might not be a good test because `ctc` has the data from the initial run. You should make a NEW contract object with a new account based on `ctc.getInfo()` after Alice is over

jeapostrophe (changes requested):

examples/posthumous-views/index.mjs
```diff
@@ -0,0 +1,22 @@
...
+  observeNumber: async () => {
+    const maybeN = await ctc.views.Number.number();
+    console.log(`Alice observes the current number as ${maybeN[1]}`);
+  },
+});
+
+// 'Posthumous' code, looking at view set while the contract was alive
+
+const maybeN = await ctc.views.Number.number();
+if (maybeN[0] !== 'None')
```
jeapostrophe, line 20: Get thee hence not using `{}`s on `if`!

js/stdlib/ts/ALGO.ts
```diff
@@ -2189,18 +2217,21 @@ const verifyContract_ = async (label:string, info: ContractInfo, bin: Backend, e
...
   const chkeq_bn_ = chkeq_x<BigNumber>((a:BigNumber, b:BigNumber) => a.eq(b));
   const chkeq_bn = (a:unknown, b:unknown, msg:string) => chkeq_bn_(bigNumberify(a), bigNumberify(b), msg);
   type Program = string|undefined;
   const chkeq_bs = chkeq_x<Program>((a:Program, b:Program) => j2sf(a) === j2sf(b));
 
-  const appInfoM = await getApplicationInfoM(ApplicationID);
-  if ( 'exn' in appInfoM ) {
+  const appInfoM_live = await getApplicationInfoM(ApplicationID);
+  const appInfoM = 'val' in appInfoM_live ? appInfoM_live
+                                          : await getDeletedApplicationInfoM(ApplicationID);
```
jeapostrophe, line 2227: You should put all of this logic just in `getApplicationInfoM`... it should just do this

js/stdlib/ts/ALGO.ts
```diff
@@ -1199,6 +1199,34 @@ const getApplicationInfoM = async (idn:BigNumber): Promise<OrExn<AppInfo>> => {
   return 'exn' in res ? res : { val: res.val.application };
 };
 
+const getDeletedApplicationInfoM = async (idn: BigNumber): Promise<OrExn<AppInfo>> => {
+  const dhead = 'getDeletedApplicationInfoM'
+  const indexer = await getIndexer();
+  const query = indexer.searchForTransactions()
+                       .txType('appl')
+                       .applicationID(bigNumberToNumber(idn))
+                       .limit(1);
```
jeapostrophe, line 1208: First, I don't believe that there is a guarantee you will get the first one.

Second, you should cast this to the correct `ApiCall` type so that the type checker works

js/stdlib/ts/ALGO.ts
```diff
@@ -1199,6 +1199,34 @@ const getApplicationInfoM = async (idn:BigNumber): Promise<OrExn<AppInfo>> => {
...
+  const queryRes = await doQueryM_(dhead, query);
+
+  if ('val' in queryRes) {
+    const txn = queryRes.val.transactions[0];
+    const appTxn = txn['application-transaction'];
+    debug(dhead, {txn});
+    return { val: {
+      'id': bigNumberToBigInt(idn),
```
jeapostrophe, line 1216: You should read this from the txn

js/stdlib/ts/ALGO.ts
```diff
@@ -1199,6 +1199,34 @@ const getApplicationInfoM = async (idn:BigNumber): Promise<OrExn<AppInfo>> => {
...
+const getDeletedApplicationInfoM = async (idn: BigNumber): Promise<OrExn<AppInfo>> => {
+  const dhead = 'getDeletedApplicationInfoM'
+  const indexer = await getIndexer();
+  const query = indexer.searchForTransactions()
+                       .txType('appl')
+                       .applicationID(bigNumberToNumber(idn))
+                       .limit(1);
+  const queryRes = await doQueryM_(dhead, query);
+
+  if ('val' in queryRes) {
```
jeapostrophe, line 1211: This isn't precise enough... you need to ensure that the transaction you found is actually the creation one. You can tell because the application id will be 0 and the created application id will be the one you searched for

js/stdlib/ts/ALGO.ts
jeapostrophe: And, you should specifically look for the error that says "It doesn't exist" or "It was deleted" or whatever it is

jeapostrophe: I thought, by the way, that you said that when you query the indexer, you can get it to tell you about deleted applications, but that it didn't have enough information. I think the flow is: Ask node, it says no, ask indexer, it says it exists but was deleted, then you go ask the indexer for the transactions so you can find the created one, and then synthesize the response

js/stdlib/ts/ALGO.ts
```diff
@@ -2189,18 +2217,21 @@ const verifyContract_ = async (label:string, info: ContractInfo, bin: Backend, e
...
   type Program = string|undefined;
   const chkeq_bs = chkeq_x<Program>((a:Program, b:Program) => j2sf(a) === j2sf(b));
 
-  const appInfoM = await getApplicationInfoM(ApplicationID);
-  if ( 'exn' in appInfoM ) {
+  const appInfoM_live = await getApplicationInfoM(ApplicationID);
+  const appInfoM = 'val' in appInfoM_live ? appInfoM_live
+                                          : await getDeletedApplicationInfoM(ApplicationID);
+
+  if ('exn' in appInfoM)
```
jeapostrophe, line 2229: No `{}`!

jeapostrophe (changes requested):
You should also test on Ethereum

js/stdlib/ts/ALGO.ts
erer1243: The id will be set to 0 if we find the right txn, as you mentioned, so I think it doesn't make sense to do that.

js/stdlib/ts/ALGO.ts
erer1243: Following the transaction querying logic in the indexer, you come to a line of the relevant query gen (see link) that shows this query will always get the first txn. However, I agree that since this is relying on an implementation detail of the indexer it's a good idea to add a sanity check 
 https://github.com/algorand/indexer/blob/82b9fe3e8e0cd5599e219a5adff79d30119099ea/idb/postgres/postgres.go#L825-L828

js/stdlib/ts/ALGO.ts
```diff
@@ -1188,38 +1192,64 @@ const getApplicationInfoM = async (idn:BigNumber): Promise<OrExn<AppInfo>> => {
...
-  debug(dhead, 'indexer', res);
-  return 'exn' in res ? res : { val: res.val.application };
+  const query = indexer.lookupApplications(id)
+                       .includeAll(true) as unknown as ApiCall<IndexerAppInfoRes>;
+  const queryRes = await doQueryM_(dhead, query);
+
+  if ('val' in queryRes) {
+    // If application was deleted, synthesize AppInfo from transaction data
+    return queryRes.val.application.deleted ? getDeletedApplicationInfoM(id)
+                                            : { val: queryRes.val.application };
```
jeapostrophe, line 1206: I really hate this indentation :vomiting_face:

js/stdlib/ts/ALGO.ts
```diff
@@ -1188,38 +1192,64 @@ const getApplicationInfoM = async (idn:BigNumber): Promise<OrExn<AppInfo>> => {
...
+
+    if (appTxn === undefined
+        || txn['created-application-index'] === undefined
+        || appTxn['application-id'] !== BigInt(0)
+        || appTxn['approval-program'] === undefined
+        || appTxn['clear-state-program'] === undefined
+        || appTxn['local-state-schema'] === undefined
+        || appTxn['global-state-schema'] === undefined)
+    {
+      return { exn: 'sanity check failed' };
```
jeapostrophe, line 1238: I don't like this error... I think it should say something like "I thought it was deleted, but when I went to go get the deployment transaction it wasn't right"

jeapostrophe (changes requested):
Your CI tests are still failing, btw

js/stdlib/ts/ALGO.ts
```diff
@@ -1188,38 +1192,64 @@ const getApplicationInfoM = async (idn:BigNumber): Promise<OrExn<AppInfo>> => {
...
+    if (queryRes.val.transactions.length === 0) {
+      return { exn: 'application does not exist' };
+    }
+
     const txn = queryRes.val.transactions[0];
     const appTxn = txn['application-transaction'];
     debug(dhead, {txn});
+
+    if (appTxn === undefined
+        || txn['created-application-index'] === undefined
```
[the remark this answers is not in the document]
erer1243: good catch

js/stdlib/ts/ALGO.ts
```diff
@@ -1188,38 +1192,64 @@ const getApplicationInfoM = async (idn:BigNumber): Promise<OrExn<AppInfo>> => {
...
       'deleted': true,
       'params': {
         'creator': txn['sender'],
+        'approval-program': appTxn['approval-program'],
+        'clear-state-program': appTxn['clear-state-program'],
+        'local-state-schema': appTxn['local-state-schema'],
+        'global-state-schema': appTxn['global-state-schema'],
         'global-state': [],
-        ...appTxn
+        'extra-program-pages': BigInt(0),
```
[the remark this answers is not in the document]
erer1243: I wasn't sure what to do about that. The field is not present in the transaction data received from the indexer.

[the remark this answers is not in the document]
jeapostrophe: Yea it is --- https://developer.algorand.org/docs/rest-apis/indexer/#transactionapplication --- `extra-program-pages`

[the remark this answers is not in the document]
erer1243: It's listed there but when the program is run it is not given in either the deployment transaction nor the initial application lookup results.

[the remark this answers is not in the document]
erer1243: TransactionApplication query result for the deployment transaction lookup:
```
{
    accounts: [],
    'application-args': [],
    'application-id': 0n,
    'approval-program': <omitted>,
    'clear-state-program': 'Bg==',
    'foreign-apps': [],
    'foreign-assets': [],
    'global-state-schema': { 'num-byte-slice': 2n, 'num-uint': 0n },
    'local-state-schema': { 'num-byte-slice': 0n, 'num-uint': 0n },
    'on-completion': 'noop'
  }
}
```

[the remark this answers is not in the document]
erer1243: Oh, that makes sense. I'll throw it in the type

jeapostrophe (commented):
These changes are great... just pass CI

jeapostrophe:
Alright, I investigated and I made a few tweaks. The bad advice I gave before was that we could make ALL of the view checks on Bob, but actually, we should only do the LAST one as Bob... the ones before aren't right because Alice's client might get multiple "turns" before Bob's client gets to run

jeapostrophe:
I believe that what I pushed will have less errors than what you did and you can use its strategy for the other errors

jeapostrophe (approved):

erer1243:
rebased onto master bc there was a conflict.

erer1243:
and cancelled ci

jeapostrophe (approved):

PR #660, opened 2022-03-08
js/browser-tests: try snapping screenshots

DanBurton:
Good news! By adding screenshots, we can see that Circle is, in fact, running through the whole testing process successfully. I'm not sure what's up with the video, but I'm satisfied by the screenshots.

https://app.circleci.com/pipelines/github/reach-sh/reach-lang/5684/workflows/a6168dda-4783-45e8-9547-a4ea0c0b51fd/jobs/106891/artifacts

### DIFF
--- a/.circleci/config.pre.yml
+++ b/.circleci/config.pre.yml
@@ -310,6 +310,8 @@ jobs:
         (cd js/browser-tests && make run)
     - store_artifacts:
         path: js/browser-tests/cypress/videos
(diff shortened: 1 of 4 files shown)


jeapostrophe (approved):
Cool; I really want to see what a failure looks like so we can be sure it gets caught. I think it would be useful to make a branch where you deliberately break something about we can see this fail

PR #662, opened 2022-03-08
CORE-676 suggestions

mattaudesse:


vsce/client/src/util/checkForUpdates.ts
```diff
@@ -114,7 +114,7 @@ export default async (
 				} catch (error) {
 					OUTPUT_CHANNEL.appendLine(
 						'Checking for updates failed. ' +
-						new Date().toLocaleTimeString()
+						`${new Date().toLocaleTimeString()}: ${error}`
```
mattaudesse, line 117: This caught the error but didn't share it before.

vsce/client/src/util/spawnVersionCompareJson.ts
```diff
@@ -72,7 +72,7 @@ export default (
 						process.stderr.readableLength
 					)
 				);
-				reject(null);
+				return reject(null);
```
mattaudesse, line 75: `return` is important here so we don't fall through to the `resolve` below.

hamirmahal: That makes sense.

PR #665, opened 2022-03-09
CORE-480: Pass Private Values Through Fork

chrisnevers:
Changes:
* Added `_local` field to PUBLISH_EXPR of a `fork`/`parallelReduce` `.case`. When there are multiple cases, a `Data` instance will be used to pass values of different types.
* Added a `Pretty` instance for `JSExpression` and `JSStatement`
* Handle secret values properly in `Object`s. Projections were always treated as public before.


docs/src/rsh/errors/index.md
```diff
@@ -2566,6 +2566,12 @@ const x = forall(UInt);
 assert(x >= 0);
 ```
 
+## {#RX0001} RX0001
+
+This error should never be encountered.
+If you receive this error, please report it to the Reach team along with the program that produced it.
```
jeapostrophe, line 2572: Then why isn't it an `impossible`?

hs/src/Reach/AST/DLBase.hs
```diff
@@ -23,6 +23,7 @@ import Reach.Counter
 import Reach.Pretty
 import Reach.Texty
 import Reach.Util
+import Data.Bifunctor
```
jeapostrophe, line 26: You and your bifunctors!

hs/src/Reach/AST/DLBase.hs
```diff
@@ -53,7 +54,7 @@ data DLType
   | T_Token
   | T_Array DLType Integer
   | T_Tuple [DLType]
-  | T_Object (M.Map SLVar DLType)
+  | T_Object (M.Map SLVar (SecurityLevel, DLType))
```
jeapostrophe, line 57: No no no, we can't do that. Security levels cannot exist at run-time

hs/src/Reach/Eval/Core.hs
```diff
@@ -4907,12 +4959,20 @@ doFork ks (ForkRec {..}) = locAt slf_at $ do
...
         let cfc_data_def = map data_go $ indexed res_ty_ms
+        let cfc_local_data_def = map local_data_go $ indexed res_ty_ms
+        let local_data_tys = map lookupLocalTy res_ty_ms
         let ra = jsa res_e
         let res_msg = JSMemberDot res_e ra (jid "msg")
         let res_when = JSMemberDot res_e ra (jid "when")
+        let res_local = JSMemberDot res_e ra (jid "_local")
         let run_ss = zipWith (defcon . jid) beforeNames beforeClosures
-        let only_body = run_ss <> cr_ss <> [defcon msg_e res_msg, defcon when_e res_when]
+        let def_local = if all (== T_Null) local_data_tys then
```
jeapostrophe, line 4971: You know I hate `if then else`

hs/src/Reach/Eval/Core.hs
```diff
@@ -5034,6 +5102,15 @@ getReturnAnnot = \case
           JSReturn a _ _:_ -> Just a
           _ -> Nothing
 
+deconstructFun :: JSExpression -> App ([JSExpression], JSBlock)
+deconstructFun = \case
+  a@(JSArrowExpression _ _ cb) -> return $ (getFormals a, jsArrowBodyToRetBlock cb)
```
jeapostrophe, line 5107: You should make this robust to having parens and to using `function` and thus it should probably be a helper function... there might even already be something like in JSUtil

hs/src/Reach/JSOrphans.hs
```diff
@@ -18,8 +19,190 @@ deriving instance Read Token
...
+    JSBinOpOf _ -> "of "
+    JSBinOpOr _ -> "||"
+    JSBinOpPlus _ -> "+"
+    JSBinOpRsh _ -> ">>"
+    JSBinOpStrictEq _ -> "==="
+    JSBinOpStrictNeq _ -> "!=="
+    JSBinOpTimes _ -> "*"
+    JSBinOpUrsh _ -> ">>>"
+
+
```
jeapostrophe, line 69: !

hs/src/Reach/JSUtil.hs
```diff
@@ -173,6 +174,15 @@ mkCommaTrailingList xs = JSCTLComma (toJSCL xs) JSNoAnnot
 a2sp :: JSAnnot -> JSSemi
 a2sp = JSSemi
 
+getFormals :: JSExpression -> [JSExpression]
+getFormals = \case
```
jeapostrophe, line 178: There's gotta be some other code that does this... so you should make it use this function

jeapostrophe (changes requested):

docs/src/rsh/errors/index.md
chrisnevers: I piggy-backed off how DLE_Impossible uses `expect_thrown` in the connectors, but I'll special case this to use `impossible`

hs/src/Reach/AST/DLBase.hs
chrisnevers: okay, I'll try to work it out another way

jeapostrophe (approved):

PR #667, opened 2022-03-09
fix error message to suggest init instead of deploy

willghatch:


willghatch:
Earlier today I wrote a program and forgot to write `init`, and the compiler suggested the deprecated `deploy`.

jeapostrophe (approved):

PR #671, opened 2022-03-10
use minbalance from network, not our calc

jeapostrophe:
The indexer doesn't actually support this


jeapostrophe:
Algorand doesn't want to update the indexer

PR #673, opened 2022-03-11
CORE-740: protest superuser invocation

mattaudesse:
I've left a `REACH_ALLOW_SU` escape-hatch for those savvy enough to use it just in case.

`id -u` seems to work well at catching all of `root`/`su`/`sudo`/and `doas` based on my testing.

### DIFF
--- a/reach
+++ b/reach
@@ -6,6 +6,32 @@ if [ ! "$REACH_ALLOW_SU" = '1' ] && [ "$(id -u)" -eq 0 ]; then
   echo "Please try again while logged into a normal, non-root account, and without using \
 \`sudo\`, \`su\`, or \`doas\` to grant elevated permissions."
   echo
+  case "$(uname -s)" in
+    Linux)
+      if ! (groups "$(logname)" 2>&1 | grep -q docker 2>/dev/null); then
+        echo "Adding your account to the \`docker\` group should make this possible:"
+        echo "https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user"
+        echo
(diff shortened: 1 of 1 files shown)


jeapostrophe (approved):
Maybe say something like, "if you think you need to do this, then maybe you really need to add Yourself to the Docker group"?

reach
```diff
@@ -1,5 +1,14 @@
 #!/bin/sh
 
+if [ ! "$REACH_ALLOW_SU" = '1' ] && [ "$(id -u)" -eq 0 ]; then
+  echo "Reach isn't intended to be run with superuser privileges."
+  echo
+  echo "Please try again while logged into a normal, non-root account, and without using \
+\`sudo\`, \`su\`, or \`doas\` to grant elevated permissions."
+  echo
```
hamirmahal, line 8:
```suggestion
  echo
  echo "If you think you need to do this, then maybe you\
really need to add yourself to the Docker group."

I believe this is [Jay's suggestion](https://github.com/reach-sh/reach-lang/pull/673#pullrequestreview-906688362).
  echo
```

hamirmahal, line 8:
```suggestion
  echo
  echo "If you think you need to do this, then maybe you\
really need to add yourself to the Docker group."
  echo
```

I believe this is [Jay's suggestion](https://github.com/reach-sh/reach-lang/pull/673#pullrequestreview-906688362).

mattaudesse: It's a little more complicated than that, unfortunately. I'm about to push an update if you're curious.

jeapostrophe (approved):

PR #674, opened 2022-03-11
CORE-741: make `reach config` harder to break

mattaudesse:


jeapostrophe (commented):
This is an improvement but if I understand correctly, I'm suggesting something bolder: you just source the config if you see it and you don't see the environment variable defined. I think what you have here will bug "power users" that define the variable and didn't source their config

docs/src/tool/index.md
```diff
@@ -27,44 +27,77 @@ You can copy this file to other repositories or move it to a directory in your `
 If you're using Windows, consult [the guide to using Reach on Windows](##guide-windows).
 :::
 
-# {#ref-usage} Select a version
+# {#ref-usage} Usage
```
jeapostrophe, line 30: I don't like the three levels of headings. I think it looks better on the page to just have one heading level the entire time, even though it is less ontologically pure

jeapostrophe (changes requested):

docs/src/tool/index.md
```diff
@@ -108,7 +139,7 @@ The output name of a library is the same as if it exported a `{!rsh} Reach.App`
 
 A `{!cmd} reach compile` usage example is available in [Overview](##over-compile).
 
-# {#ref-usage-init} @{ref("cmd", "reach init")} `reach init`
+# {#ref-usage-init} @{ref("cmd", "reach init")} reach init
```
jeapostrophe, line 142: Why remove the "`"?

jeapostrophe (changes requested):

docs/src/tool/index.md
mattaudesse: I feel like we should either remove the ticks here or improve our CSS because the current formatting isn't great:

image

Headings should look like headings IMHO. Right now they're lost in the crowd. I also think subheadings make more sense.

mattaudesse: I realize you were probably too busy to respond on Slack, so I've taken a stab in 5f355dad and created DCT-161 to track this.

jeapostrophe (approved):

PR #679, opened 2022-03-11
print-keyword-info: include stdenv exports

willghatch:
Note that this still doesn't quite bring the generated map to parity with the old JSON file.  The old file includes `Array.map`, while the stdlib has only `Array_map`, and other similar Array methods.

### DIFF
--- a/hs/src/Reach/EditorInfo.hs
+++ b/hs/src/Reach/EditorInfo.hs
@@ -1,4 +1,4 @@
-module Reach.EditorInfo (printKeywordInfo) where
+module Reach.EditorInfo (printBaseKeywordInfo) where
 
 import qualified Data.Aeson as A
 import qualified Data.Aeson.Encode.Pretty as A
@@ -18,23 +18,15 @@ customConfig = A.Config {
   confTrailingNewline = True
 }
 
-printKeywordInfo :: IO ()
-printKeywordInfo =
+printBaseKeywordInfo :: (M.Map String SLVal) -> IO ()
+printBaseKeywordInfo stdlibExports =
   B.putStr $
-    A.encodePretty'
-      customConfig $ infoMap
+    A.encodePretty' customConfig $ A.toJSON $
+       M.map
+         (\v -> M.singleton ("CompletionItemKind" :: String) $ show v)
+         $ M.mapMaybe completionKind
+         $ M.union (M.fromList C.base_env_slvals) stdlibExports
 
-infoMap :: A.Value
-infoMap =
-  A.toJSON $
-    M.map
-      (\v -> M.singleton ("CompletionItemKind" :: String) $ show v)
-      completionTypeMap
-
-completionTypeMap :: M.Map String CompletionItemKind
-completionTypeMap =
-  M.mapMaybe completionKind
-  $ M.fromList C.base_env_slvals
 
 data CompletionItemKind
   = CK_Text
@@ -70,6 +62,22 @@ instance Show CompletionItemKind where
 completionKind :: SLVal -> Maybe CompletionItemKind
 completionKind v =
   case v of
+    SLV_Null _ _ -> Just CK_Constant
+    SLV_Bool _ _ -> Just CK_Constant
+    SLV_Int _ _ -> Just CK_Constant
+    SLV_Bytes _ _ -> Just CK_Constant
+    SLV_Array _ _ _ -> Just CK_Constant
+    SLV_Tuple _ _ -> Just CK_Constant
+    SLV_Object _ _ _ -> Just CK_Constant
+    SLV_Struct _ _ -> Just CK_Constant
+    SLV_Clo _ _ _ -> Just CK_Function
+    SLV_Data _ _ _ _ -> Just CK_Constructor -- ?
+    SLV_DLC _ -> Just CK_Constant
+    SLV_Connector _ -> Nothing
+    SLV_RaceParticipant _ _ -> Just CK_Constant -- ?
+    SLV_Participant _ _ _ _ -> Just CK_Constant -- ?
+    SLV_Map _ -> Just CK_Variable -- ?
+    SLV_Deprecated _ _ -> Nothing -- ?
     SLV_Type _ -> Just CK_TypeParameter
     SLV_Kwd _ -> Just CK_Keyword
     SLV_DLVar _ -> Just CK_Variable
@@ -180,4 +188,4 @@ completionKind v =
         SLForm_apiCall_partial _ -> Nothing
         SLForm_wait -> Just CK_Function
         SLForm_setApiDetails -> Nothing
-    _ -> Nothing
+    --_ -> Nothing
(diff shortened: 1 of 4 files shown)


jeapostrophe (approved):
You should go into each symbol and see if evalAsObj lets you see inside and then show those things too

hs/src/Reach/EditorInfo.hs
```diff
@@ -70,6 +62,22 @@ instance Show CompletionItemKind where
...
+    SLV_Null _ _ -> Just CK_Constant
+    SLV_Bool _ _ -> Just CK_Constant
+    SLV_Int _ _ -> Just CK_Constant
+    SLV_Bytes _ _ -> Just CK_Constant
+    SLV_Array _ _ _ -> Just CK_Constant
+    SLV_Tuple _ _ -> Just CK_Constant
+    SLV_Object _ _ _ -> Just CK_Constant
+    SLV_Struct _ _ -> Just CK_Constant
+    SLV_Clo _ _ _ -> Just CK_Function
+    SLV_Data _ _ _ _ -> Just CK_Constructor -- ?
```
jeapostrophe, line 74: Yea, this is a value

hs/src/Reach/EditorInfo.hs
```diff
@@ -70,6 +62,22 @@ instance Show CompletionItemKind where
...
+    SLV_DLC _ -> Just CK_Constant
+    SLV_Connector _ -> Nothing
+    SLV_RaceParticipant _ _ -> Just CK_Constant -- ?
+    SLV_Participant _ _ _ _ -> Just CK_Constant -- ?
```
jeapostrophe, line 78: These are both values too

hs/src/Reach/EditorInfo.hs
```diff
@@ -70,6 +62,22 @@ instance Show CompletionItemKind where
...
+    SLV_Int _ _ -> Just CK_Constant
+    SLV_Bytes _ _ -> Just CK_Constant
+    SLV_Array _ _ _ -> Just CK_Constant
+    SLV_Tuple _ _ -> Just CK_Constant
+    SLV_Object _ _ _ -> Just CK_Constant
+    SLV_Struct _ _ -> Just CK_Constant
+    SLV_Clo _ _ _ -> Just CK_Function
+    SLV_Data _ _ _ _ -> Just CK_Constructor -- ?
+    SLV_DLC _ -> Just CK_Constant
+    SLV_Connector _ -> Nothing
```
jeapostrophe, line 76: We actually do want to share these. They are values

hs/src/Reach/EditorInfo.hs
```diff
@@ -70,6 +62,22 @@ instance Show CompletionItemKind where
...
+    SLV_RaceParticipant _ _ -> Just CK_Constant -- ?
+    SLV_Participant _ _ _ _ -> Just CK_Constant -- ?
+    SLV_Map _ -> Just CK_Variable -- ?
```
jeapostrophe, line 79: Yup

hs/src/Reach/EditorInfo.hs
```diff
@@ -70,6 +62,22 @@ instance Show CompletionItemKind where
...
+    SLV_Deprecated _ _ -> Nothing -- ?
```
jeapostrophe, line 80: You could go inside and show what the thing inside is or leave it this way and they won't show

hs/src/Reach/EditorInfo.hs
```diff
@@ -180,4 +188,4 @@ completionKind v =
         SLForm_apiCall_partial _ -> Nothing
         SLForm_wait -> Just CK_Function
         SLForm_setApiDetails -> Nothing
-    _ -> Nothing
+    --_ -> Nothing
```
jeapostrophe, line 191: Delete, don't comment

hs/src/Reach/Parser.hs
```diff
@@ -342,12 +342,16 @@ map_order dm = order
...
 gatherDeps_top src_p e_install e_dreachp = do
   let e_at = srcloc_top
   e_bm <- liftIO $ newIORef (mempty, mempty)
   flip runReaderT (Env {..}) $ do
-    _src_abs_p <- gatherDeps_file GatherTop src_p
+    unless (src_p == ReachStdLib) $ do
+      _src_abs_p <- case src_p of
+        ReachSourceFile f -> gatherDeps_file GatherTop f
+        ReachStdLib -> impossible "gatherDeps"
+      return ()
```
jeapostrophe, line 354: Use void and stop the bind and return

jeapostrophe: And just case and return () rather than having the unless

jeapostrophe (changes requested):

hs/src/Reach/EditorInfo.hs
willghatch: Oops, I put those comments while I was working on it and then forgot about them.

hs/src/Reach/EditorInfo.hs
jeapostrophe: !

jeapostrophe (changes requested):
All else good

willghatch:
Oh, looking at this again I haven't yet done the “look in and see if evalAsObject works” bit.

jeapostrophe:
It's okay if that is a different pr. There is the thing about connectors

willghatch:
Ok, I pushed the commit that adds the connector bit.

I've been working on the eval part.  I don't see anything called `evalAsObj`, so I tried it with `evalAsEnv`, which I think is what you meant.  After writing some code to use it, I realized that it has a fall-through case that just errors, so I transformed it to return a Maybe and return Nothing on fall-through.  But it looks like it also has indirect `impossible` calls in some of the other branches that I'm tripping.  So before I push on it farther and muck up `evalAsEnv` with extra code to return `Nothing` instead of erroring, I thought I would ask if that's the direction you were hoping for.

jeapostrophe (approved):

jeapostrophe:
You should include an updated file when you do commits like this, too

jeapostrophe:
1. `evalAsEnv` is what I meant
2. Yes, it is fine to make `evalAsEnvM` and just remove the one error case and turn it into a `Nothing` where a new `evalAsEnv` throws the same old error.
3. From what I can tell, there's only one `impossible` case and that's on a `Participant`. You should just refactor the code to do the check first and don't include `"interact"` in the map that gets returned

willghatch:
>     3. From what I can tell, there's only one `impossible` case and that's on a `Participant`. You should just refactor the code to do the check first and don't include `"interact"` in the map that gets returned

I'm hitting errors on calls to `aisd` (also, what does that stand for?) which I worked around, and now `lookupStdlib`.  This could be because I'm using a bad environment.  I made an `Env` for `runReaderT` with `makeEnv mempty`, and I can imagine these lookup calls would work if I put the `base_env` in it, but I'm not certain how to do that.

jeapostrophe:
> > You should include an updated file when you do commits like this, too
> 
> I'm not sure what you mean.

I mean to re-produce the `keywordToItemKind.json` file; you should make a test that it matches the output of this to remember

PR #682, opened 2022-03-12
Item kind with eval

willghatch:
Since the other PR (#679) has been merged but still had some trailing questions, I thought I would just make a draft PR.


hs/src/Reach/Eval/Core.hs
```diff
@@ -1137,28 +1151,31 @@ evalAsEnv obj = case obj of
...
-    at <- withAt id
-    apis <- ae_apis <$> aisd
-    whos <- S.fromList . M.keys <$> (ae_ios <$> aisd)
-    let ps = S.difference whos apis
-    case S.toList ps of
-      [] -> expect_ $ Err_No_Participants
-      [h] -> evalAsEnv $ SLV_Participant at h Nothing Nothing
-      _ -> evalAsEnv $ SLV_RaceParticipant sb ps
+    aisdM <- aisd_
+    case aisdM of
```
jeapostrophe, line 1166: I tend to use `aisd_ >>= \case` for this pattern

jeapostrophe (changes requested):

willghatch:
I've been stuck with a problem for while now that I haven't gotten past.

When it is doing `evalObjEnv` on `Array_empty`, it does `lookstdlib` with an empty environment and fails to look up `Foldable_all1`.  If I have it skip lookups where the key matches the string `Array_empty`, it runs to completion and I get identifiers included like `Set.new` and a few others.  But it does not include any `Participant.___` or `Array.___` identifiers.

I'm not sure why `lookstdlib` is getting an empty environment, especially since some others of these work successfully.  I'm also not sure why I'm getting an environment for `Set` but not `Participant` or `parallelReduce`, etc.

willghatch:
Oh, of course it's not looking things up for the base_env because I left that part out of that loop...  But I have yet to figure out the empty environment thing.

willghatch:
Here is a version that runs to completion by skipping the keys that it crashes on.  If it tries to do `evalObjEnv` on any of those five values it crashes because it has an empty environment.  So that's what I'm stuck on.

jeapostrophe:
I'm experimenting with this

PR #686, opened 2022-03-15
CORE-1261: Support `xor` on Digests

chrisnevers:
Add `polyXor` that works on UInts, Digests and Bools

### DIFF
--- a/docs/src/rsh/compute/index.md
+++ b/docs/src/rsh/compute/index.md
@@ -585,9 +585,10 @@ A @{defn("binary expression")} is written `{!rsh} EXPR_lhs BINOP EXPR_rhs`, wher
 Numeric operations, like `{!rsh} +` and `{!rsh} >`, only operate on numbers.
 Since all numbers in Reach are integers, operations like `{!rsh} /` truncate their result.
 Boolean operations, like `{!rsh} &&`, only operate on booleans.
+`{!rsh} polyXor` operates on `{!rsh} UInt`s, `{!rsh} Bool`s, and `{!rsh} Digest`s.
 It is invalid to use binary operations on the wrong types of values.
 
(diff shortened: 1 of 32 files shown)


docs/src/rsh/compute/index.md
```diff
@@ -585,9 +585,10 @@ A @{defn("binary expression")} is written `{!rsh} EXPR_lhs BINOP EXPR_rhs`, wher
 Numeric operations, like `{!rsh} +` and `{!rsh} >`, only operate on numbers.
 Since all numbers in Reach are integers, operations like `{!rsh} /` truncate their result.
 Boolean operations, like `{!rsh} &&`, only operate on booleans.
+`{!rsh} polyXor` operates on `{!rsh} UInt`s, `{!rsh} Bool`s, and `{!rsh} Digest`s.
```
jeapostrophe, line 588: It could work on BYTES if we wanted

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1016,6 +1016,7 @@ cprim = \case
   BAND -> call "&"
   BIOR -> call "|"
   BXOR -> call "^"
+  DIGEST_XOR -> call "b^"
```
jeapostrophe, line 1019: You have to adjust the cost analysis for this --- https://developer.algorand.org/docs/get-details/dapps/avm/teal/opcodes/#b_12

jeapostrophe (changes requested):

hs/src/Reach/Connector/ALGO.hs
chrisnevers: Is it not already handled by this? https://github.com/reach-sh/reach-lang/blob/946d738daaf83e25002475550c473bf3ed516dbe/hs/src/Reach/Connector/ALGO.hs#L530

jeapostrophe (approved):

PR #691, opened 2022-03-16
make variables verbose for future readability

hamirmahal:


hamirmahal (commented):
If it helps, the line count is the same, now.

PR #693, opened 2022-03-16
WIP - working on testing that docs are present

willghatch:
I've written some code to check that everything output by the `--print-keyword-info` flag is documented.  However, many things in that list are currently failing:
• Some things that we need to add reference anchors (`@{ref(...)}`) for, eg. `Token.burn` -- these are really documented, and just need this little push because I guess I missed them in my previous anchor adding.
• Unused Javascript keywords.
• Things that are documented with a different name, eg. `Foldable_any1` and `Array_empty.average`.

I think it would be reasonable to add a section mentioning that there are JS keywords that we don't use but which are still reserved as keywords in Reach.

What do you think I should do with respect to `Foldable_any1` and `Array_empty.average`?  @jeapostrophe  didn't mind that these show up in editor completion, but it doesn't make sense to add more to the documentation, especially for `Array_empty` nonsense.  Unless we want to document this stuff, they need some way to either be filtered from the documentation check or filtered from the completion.  I think that the things we don't want to document should be filtered from the completion provided with `--print-keyword-info` and not just from the documentation check.


hs/test/Reach/Test_Docs.hs
```diff
@@ -14,10 +16,18 @@ import Reach.PackageImport (PkgError (..))
...
-docName :: String
-docName = "../docs/src/rsh/errors/index.md"
+rshDoc :: String -> String
+rshDoc path = "../docs/src/rsh/" <> path <> "/index.md"
+
+errDoc :: String
+errDoc = rshDoc "errors"
+
+allRshDocs :: [String]
+-- TODO - get this dynamically
```
jeapostrophe, line 29: Yea

jeapostrophe (changes requested):

willghatch:
My suggestion is that we filter everything from base_env and stdlib that has an underscore from the `--print-keyword-info` output.  These are all documented with their `.` versions.

docs/src/rsh/compute/index.md
```diff
@@ -588,15 +591,17 @@ Boolean operations, like `{!rsh} &&`, only operate on booleans.
 `{!rsh} polyXor` operates on `{!rsh} UInt`s, `{!rsh} Bool`s, `{!rsh} Digest`s, and `{!rsh} Bytes` of the same length.
 It is invalid to use binary operations on the wrong types of values.
 
-@{ref("rsh", "and")}@{ref("rsh", "or")}@{ref("rsh", "add")}@{ref("rsh", "sub")}@{ref("rsh", "mul")}@{ref("rsh", "div")}@{ref("rsh", "mod")}@{ref("rsh", "lt")}@{ref("rsh", "le")}@{ref("rsh", "ge")}@{ref("rsh", "gt")}@{ref("rsh", "lsh")}@{ref("rsh", "rsh")}@{ref("rsh", "band")}@{ref("rsh", "bior")}@{ref("rsh", "band")}@{ref("rsh", "polyXor")}@{ref("rsh", "polyEq")}@{ref("rsh", "polyNeq")}
+@{ref("rsh", "and")}@{ref("rsh", "or")}@{ref("rsh", "xor")}@{ref("rsh", "add")}@{ref("rsh", "sub")}@{ref("rsh", "mul")}@{ref("rsh", "div")}@{ref("rsh", "mod")}@{ref("rsh", "eq")}@{ref("rsh", "lt")}@{ref("rsh", "le")}@{ref("rsh", "ge")}@{ref("rsh", "gt")}@{ref("rsh", "lsh")}@{ref("rsh", "rsh")}@{ref("rsh", "band")}@{ref("rsh", "bior")}@{ref("rsh", "bxor")}@{ref("rsh", "polyXor")}@{ref("rsh", "polyEq")}@{ref("rsh", "polyNeq")}
 ```reach
 and(a, b)     // &&
 or(a, b)      // ||
+xor(a, b)
```
jeapostrophe, line 598: ^

willghatch:
I've pushed a new version that filters out identifiers with `_`, and added anchors and docs for a few things that were undocumented.  (Also I fixed the TODO to actually get the doc directories dynamically.)

It is still failing for 2 identifiers that I wanted to ask about:  `intervalEmpty` and `intervalEmpty.length`.  `intervalEmpty` is not documented and doesn't seem very useful.  `<interval>.length` is always 4 from what I can tell, because it is getting the length of the struct that implements the interval.  I don't think we want to document `intervalEmpty.length`, and `intervalEmpty` is pretty useless, too, I think.  How would you feel about renaming `intervalEmpty` to `interval_empty` to make it fail the filter?

jeapostrophe (approved):

jeapostrophe:
> How would you feel about renaming `intervalEmpty` to `interval_empty` to make it fail the filter?

Go for it

docs/src/rsh/compute/index.md
willghatch: Oh, yeah, `xor` doesn't desugar to a simple operator.  `^` is `bxor`.  Probably I should move `xor` to a different place.

willghatch:
I've clarified the `xor` docs, and renamed `intervalEmpty` to `interval_empty`.  I think this is ready.

docs/src/rsh/compute/index.md
willghatch: Oh, I'm uncertain.  Maybe I should fix my doc fix.

docs/src/rsh/compute/index.md
```diff
@@ -588,15 +591,17 @@ Boolean operations, like `{!rsh} &&`, only operate on booleans.
 `{!rsh} polyXor` operates on `{!rsh} UInt`s, `{!rsh} Bool`s, `{!rsh} Digest`s, and `{!rsh} Bytes` of the same length.
 It is invalid to use binary operations on the wrong types of values.
 
-@{ref("rsh", "and")}@{ref("rsh", "or")}@{ref("rsh", "add")}@{ref("rsh", "sub")}@{ref("rsh", "mul")}@{ref("rsh", "div")}@{ref("rsh", "mod")}@{ref("rsh", "lt")}@{ref("rsh", "le")}@{ref("rsh", "ge")}@{ref("rsh", "gt")}@{ref("rsh", "lsh")}@{ref("rsh", "rsh")}@{ref("rsh", "band")}@{ref("rsh", "bior")}@{ref("rsh", "band")}@{ref("rsh", "polyXor")}@{ref("rsh", "polyEq")}@{ref("rsh", "polyNeq")}
+@{ref("rsh", "and")}@{ref("rsh", "or")}@{ref("rsh", "xor")}@{ref("rsh", "add")}@{ref("rsh", "sub")}@{ref("rsh", "mul")}@{ref("rsh", "div")}@{ref("rsh", "mod")}@{ref("rsh", "eq")}@{ref("rsh", "lt")}@{ref("rsh", "le")}@{ref("rsh", "ge")}@{ref("rsh", "gt")}@{ref("rsh", "lsh")}@{ref("rsh", "rsh")}@{ref("rsh", "band")}@{ref("rsh", "bior")}@{ref("rsh", "bxor")}@{ref("rsh", "polyXor")}@{ref("rsh", "polyEq")}@{ref("rsh", "polyNeq")}
 ```reach
 and(a, b)     // &&
 or(a, b)      // ||
+xor(a, b)     // No equivalent infix operator.
```
jeapostrophe, line 598: @chrisnevers Can you confirm this? I thought you made `xor` turn into `polyXor` too? If not, then can you plan to make that change after this PR is merged and update this line in the docs too?

PR #700, opened 2022-03-18
add `RW` 4,5 prob/correct code, `RW` 6 prob code

hamirmahal:
image

### DIFF
--- a/docs/src/rsh/errors/index.md
+++ b/docs/src/rsh/errors/index.md
@@ -2697,7 +2697,7 @@ You can fix this issue by making sure at least one `{!rsh} Participant` performs
 'reach 0.1';
 
 export const main = Reach.App(() => {
-	const aisha = Participant('Aisha', {});
+  const aisha = Participant('Aisha', {});
 });
 ```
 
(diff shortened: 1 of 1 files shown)


jeapostrophe (dismissed):
Can you remove the extra new lines, use two spaces for indenting, and put a comment on the line that is bad and changed

DanBurton (dismissed):
On my phone so pardon the lack of responding to code inline.

This is a step in the right direction and can be merged. But I have suggestions for improvement which can be done now or as a follow on PR if you prefer.

Comment in general: don't show entire reach programs for the examples. Only show a relevant code snippet. Don't double space the lines. Compact examples are easier for the reader to digest and easier to eyeball the diff between the problematic and correct code samples.

* For `RW0000`, deprecation warnings are too broad of an error class; I think this example will be confusing for any deprecated identifier besides `deploy` and it's better to not have a specific example on this one. (Arguably the compiler should be updated to emit more specific warning codes for each specific deprecated identifier so that we can give specific examples of how to address each case.)
* For `RW0004`, lowercase main, and make Aisha's interact interface empty (`{}`) for simplicity.
* For `RW0005`, only show the line defining the View, and add a suggested fix.
* For `RW0006`, this example is much too cluttered. Just show one or two lines of reach code that actually reference network seconds.

DanBurton (approved):

PR #704, opened 2022-03-21
CORE-1305 Remove "appl" filter on txn search

erer1243:
as Dan mentioned; apparently Randlabs has trouble with it


jeapostrophe (changes requested):
We need to filter on the client-side though --- https://github.com/reach-sh/reach-lang/blob/master/js/stdlib/ts/ALGO.ts#L713

erer1243:
I don't think we need to here. We're only getting one txn, ordered by time, so we must get the first txn mentioning the app. That has to be it's launch txn, right?

jeapostrophe:
Better safe than sorry

jeapostrophe (approved):

PR #705, opened 2022-03-21
CORE-915 Dead man's switch example

erer1243:
wip of dead man's switch example. I'm sure there are better ways to implement some parts of it :^). Figuring there is a better way to refer to a single member of a participant class, and a better way to 'die' / opt out of a contract than just sleeping.

### DIFF
--- a/examples/dead-mans-switch-nft/index.rsh
+++ b/examples/dead-mans-switch-nft/index.rsh
@@ -1,0 +1,74 @@
+'reach 0.1';
+'use strict';
+
+export const main = Reach.App(() => {
+  const Creator = Participant('Creator', {
+    transferTo: Fun([], Address),
+  });
+  const Owner = ParticipantClass('Owner', {
+    notifyReceive: Fun([], Null),
+    notifyRenewNeeded: Fun([], Null),
+    notifyRenewSuccess: Fun([], Null),
+    notifyRenewFail: Fun([], Null),
+  });
+  const vNFT = View('NFT', { owner: Address });
+
+  init();
+
+  Creator.publish();
+
+  // const ownerOnly = (currentOwner, fn) =>
+  //   Owner.only(() => { if (this == currentOwner) { fn(interact); } });
+
+  var owner = Creator;
+  invariant(true);
+  while (true) {
+    vNFT.owner.set(owner);
+
+    if (owner == Creator) {
+      // Creator has the NFT
+      // Either the contract is brand new or the dead man's switch has triggered
+      commit();
+
+      // Creator gives the NFT to someone else to arm the switch
+      Creator.only(() => {
+        const recipient = declassify(interact.transferTo());
+      });
+      Creator.publish(recipient);
+
+      Owner.only(() => { if (this == recipient) { interact.notifyReceive(); } })
+
+      owner = recipient;
+      continue;
+    } else {
+      // Someone other than the owner has the NFT
+      commit();
+
+      // The owner is 'safe' for 3 seconds
+      wait(relativeSecs(3));
+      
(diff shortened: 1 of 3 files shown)


examples/dead-mans-switch-nft/index.rsh
```diff
@@ -0,0 +1,74 @@
+'reach 0.1';
+'use strict';
+
+export const main = Reach.App(() => {
+  const Creator = Participant('Creator', {
+    transferTo: Fun([], Address),
+  });
+  const Owner = ParticipantClass('Owner', {
```
jeapostrophe, line 8: Don't use participant classes, ever :)

examples/dead-mans-switch-nft/index.rsh
```diff
@@ -0,0 +1,74 @@
...
+
+  // const ownerOnly = (currentOwner, fn) =>
+  //   Owner.only(() => { if (this == currentOwner) { fn(interact); } });
+
+  var owner = Creator;
+  invariant(true);
+  while (true) {
+    vNFT.owner.set(owner);
+
+    if (owner == Creator) {
```
jeapostrophe, line 28: This doesn't mean what you think it means... the way you have it the creator can never be a "normal" owner. You should not have any `if` like this in the program

examples/dead-mans-switch-nft/index.rsh
```diff
@@ -0,0 +1,74 @@
...
+  const vNFT = View('NFT', { owner: Address });
+
+  init();
+
+  Creator.publish();
+
+  // const ownerOnly = (currentOwner, fn) =>
+  //   Owner.only(() => { if (this == currentOwner) { fn(interact); } });
+
+  var owner = Creator;
```
jeapostrophe, line 23: First, I think you should use `parallelReduce`

Second, the idea is that there's always an "owner" and an "heir". If the owner doesn't ping, then the heir can claim it and become the new owner (setting up the next heir).

jeapostrophe (changes requested):

examples/dead-mans-switch-nft/index.rsh
```diff
@@ -0,0 +1,48 @@
...
+    .api(O.ping,
+      () => check(this == owner),
+      () => 0,
+      (k) => {
+        check(this == owner);
+        k(null);
+        return [owner, heir];
+      }
+    )
+    .timeout(relativeTime(3), () => {
```
jeapostrophe, line 37: That's too little... it should be a parameter of the Creator

jeapostrophe (changes requested):

examples/dead-mans-switch-nft/index.rsh
erer1243: Got it

examples/dead-mans-switch-nft/index.rsh
```diff
@@ -0,0 +1,48 @@
...
+      const [[nextHeir], k] =
+        call(O.setNextHeir)
+        .assume((nextHeir) => check(this != nextHeir));
```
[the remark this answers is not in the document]
erer1243: Whoops, thanks.

examples/dead-mans-switch-nft/index.rsh
```diff
@@ -34,11 +36,11 @@ export const main = Reach.App(() => {
         return [owner, heir];
```
jeapostrophe, line 36: You could add `owner != heir` into the invariant

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #729, opened 2022-03-27
handle errors in order of ascending compiler codes

hamirmahal:


jeapostrophe (approved):

jeapostrophe:
I think you should consider making a dictionary keyed on the code where the values are functions that you run on the code.

hamirmahal:
I was thinking about doing that, especially as we start getting more of these.

PR #739, opened 2022-03-30
View cache

willghatch:
Is this more or less what you were thinking, Jay?

The one thing that I'm sure is wrong is the cache invalidation -- I'm not really sure where things talk to the network and thus see new block numbers.  I don't think there is a single key place where that happens -- eg. there is `getNetworkTime` that does a health check request to get the time, there is `recv`, and maybe there are more?  So I just put it some place convenient where I had been logging some debug info for testing as a “good enough until I get feedback” location.

Should the cache also be keyed by the contract ID or something?  I imagine we could have one frontend running on multiple instances of the same contract at once, though I'm not sure whether that's currently possible.

### DIFF
--- a/js/stdlib/ts/ALGO.ts
+++ b/js/stdlib/ts/ALGO.ts
@@ -321,6 +321,7 @@ const reqLock = new Lock();
 let currentReqNum: number | undefined = undefined;
 let currentReqLabel: string | undefined = undefined;
 let lastReqSentAt: number | undefined = undefined; // ms
+const appStateMinRefreshMillis = 1000;
 
 async function httpEventHandler(e: RHC.Event): Promise<void> {
   const en = e.eventName;
@@ -1398,7 +1399,19 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
           }
         };
 
-        const getAppState = async (): Promise<AppStateKVs|undefined> => {
+        let lastAppState : AppStateKVs|undefined = undefined;
+        let lastAppStateTime : number = 0;
+        const getAppState = async () => {
+          const now = Date.now();
+          const minMillis = isIsolatedNetwork() ? 0 : appStateMinRefreshMillis;
+          if ( lastAppState && now - lastAppStateTime < minMillis){
+            return lastAppState;
+          }
+          lastAppState = await getAppStateFresh();
+          lastAppStateTime = now;
+          return lastAppState;
+        };
+        const getAppStateFresh = async (): Promise<AppStateKVs|undefined> => {
           const lab = `getAppState`;
           const appInfoM = await getApplicationInfoM(ApplicationID);
           if ( 'exn' in appInfoM || appInfoM.val.deleted ) {


js/stdlib/ts/ALGO.ts
```diff
@@ -1285,6 +1285,25 @@ const getDeletedApplicationInfoM = async (id: number): Promise<OrExn<AppInfo>> =
   }
 }
 
+let viewCache = new Map<string, Map<string, any>>();
```
jeapostrophe, line 1288: Use `const` for this because the container is constant but the contents are mutable

js/stdlib/ts/ALGO.ts
```diff
@@ -1978,9 +1997,19 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
           return await viewMapRef(mapi, a);
         },
       };
+
```
jeapostrophe, line 2000: You can actually define the cache right here and it will be specialized for THIS call to `contract`. Scope, man. It's what's for dinner.

js/stdlib/ts/ALGO.ts
```diff
@@ -1992,13 +2021,17 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
             });
             const vres = await decode(vi, vvs, args);
             debug({vres});
-            return isSafe ? ['Some', vres] : vres;
+            const safeResult = ['Some', vres];
+            viewCacheSet(v, k, {lastCheckedTime: now, lastValue: safeResult});
```
jeapostrophe, line 2025: However, what you have is wrong... you should make a test case that shows it wrong, but let me tell you why. You are keyed on `${v}.${k}`, but the value you're caching is `${v}.${k}` applied to `args`... so you'll do something like if the view is `f = (x) => x + y` where `y` is in SVS, then `f(3)` returns `5` and then `f(4)` would return `5` rather than `6`.

js/stdlib/ts/ALGO.ts
```diff
@@ -1992,13 +2021,17 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
             });
             const vres = await decode(vi, vvs, args);
```
jeapostrophe, line 2022: The thing we REALLY want to cache is the call to `getState_`

jeapostrophe: And since we want to cache it, I think you should move the cache right next to the definition of `getState_`

js/stdlib/ts/ALGO.ts
```diff
@@ -2248,6 +2281,8 @@ export const getNetworkTime = async (): Promise<BigNumber> => {
   return bigNumberify(hc['round']);
 };
 const getTimeSecs = async (now_bn: BigNumber): Promise<BigNumber> => {
+  // TODO - this is not the right place to do invalidation, but I'm not sure where is...
+  viewCacheInvalidate(now_bn);
```
jeapostrophe, line 2285: Yea, this is not a good place to invalidate... we know that a block lasts ~5 seconds (except on our devnet) so we could just check the time of the cache and invalidate if it is more than 5 seconds in the past

jeapostrophe (changes requested):

js/stdlib/ts/ALGO.ts
willghatch: Ah, yes, I was only looking at constant function views when I was writing this...

js/stdlib/ts/ALGO.ts
willghatch: Yes, I would have done that, but from what I could tell, the places I needed to invalidate from were outside of that scope.  If that's not the case, I'll move it back in.

js/stdlib/ts/ALGO.ts
willghatch: Is there a way to check whether we are on devnet?  You mentioned caching that expires with elapsed time OR when a new block is ready, but we could also just use a smaller timeout for devnet.

jeapostrophe: We know we're on the devnet from `isIsolatedNetwork`

willghatch:
How much do we want to cache for ETH_like?  Etherium has a ~12-14 second block time according to a quick web search, but ETH_like is for many EVM compatible chains, and I have no idea what block time is like on other EVM chains.

js/stdlib/ts/ALGO.ts
```diff
@@ -1451,13 +1451,31 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
       };
     };
 
+
+    const stateCache = new Map<GetC, { time: number, val: [AppStateKVs, GlobalState] }>();
```
jeapostrophe, line 1455: This is really weird now... now you're making it so it is indexed on the `GetC` value which is a lambda for the whole contract representation. I was just thinking that there'd be one box AS PART OF `getC` (another field) that was the cache, or rather that internal to `getC`... `getAppState` and `getGlobalState` were themselves caching

jeapostrophe (changes requested):

jeapostrophe:
Re: ETH-like ---- Right now you're only caching for 1sec anyways. That's fine for now.

willghatch:
So for ETH_like, are the view functions generated by the solidity compiler?  My best approach right now would have to be caching the result of the call based on the argument list, but that doesn't seem great.  I would like to insert caching wherever the view functions are generated, but that doesn't sound very feasible if they are generated by `solc`.

jeapostrophe:
> So for ETH_like, are the view functions generated by the solidity compiler?

View functions are actual interactions with `geth`, so there's no way to peek into their guts.

> My best approach right now would have to be caching the result of the call based on the argument list, but that doesn't seem great.

I'm willing to say that this kind of caching can't be done effectively on ETH

> I would like to insert caching wherever the view functions are generated, but that doesn't sound very feasible if they are generated by `solc`.

Correct

jeapostrophe (approved):

PR #742, opened 2022-03-30
CORE-1313: Expose `thisConsensusTime` and `thisConsensusSecs`

chrisnevers:


docs/src/rsh/compute/index.md
```diff
@@ -1641,20 +1641,32 @@ lastConsensusTime()
 The @{defn("lastConsensusTime")} primitive returns the network time of the last publication of the DApp.
 This may not be available if there was no such previous publication, such as at the beginning of an application before the first publication.
 
-:::note
-Why is there no `thisConsensusTime`?
-Some networks do not support observing the time of a consensus operation until after it has finalized.
-This aides scalability, because it increases the number of times when an operation could be finalized.
```
DanBurton, line 1647: We might want to retain this info in the form of cautionary words on the docs for `thisConsensusTime`.

A more extreme opinion might be that we name it `unsafeGetCurrentConsensusTime()` or something silly like that.

examples/thisConsensusTime/index.rsh
```diff
@@ -0,0 +1,23 @@
...
+export const main = Reach.App(() => {
+  const A = Participant('Alice', {});
+  init();
+  A.publish();
+  commit();
+
+  A.publish();
+
+  var [ lastTime, lastSecs, i ] = [ thisConsensusTime(), thisConsensusSecs(), 0 ];
+  invariant(lastTime > lastConsensusTime() &&
```
jeapostrophe, line 12: The way you did it, is it possible to prove something about the connection between `this` and `last`?

jeapostrophe (changes requested):
Why did you not need to change the JS backend?

hs/src/Reach/Linearize.hs
```diff
@@ -603,14 +603,15 @@ df_step :: DKTail -> DFApp LLStep
 df_step = \case
   DK_Stop at -> return $ LLS_Stop at
   DK_ToConsensus at send recv mtime -> do
-    lt <- fmap snd <$> fluidRefm FV_thisConsensusTime
+    lt <- fmap snd <$> fluidRefm FV_lastConsensusTime
```
DanBurton, line 606: Was this just wrong before? I'm not sure I grasp how both the previous code and the current code for this line could both be okay, since these concepts are not interchangeable.

jeapostrophe: Oh ya, good catch. It was CORRECT before, because the way we learn what time it was is looking at what the old "this" time WAS

mattaudesse (commented):
I took a look but don't have anything to add beyond what Jay and Dan already said :pizza:

chrisnevers:
@jeapostrophe the JS backend didn't need to change because `thisConsensusTime` is already tracked via `ET_ToConsensus { et_tc_from_timev :: DLVar }`

examples/thisConsensusTime/index.rsh
jeapostrophe: I think that's all I mean... that you could put an invariant `thisTime > lastTime`

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1108,11 +1108,11 @@ smt_e at_dv mdv de = do
...
     DLE_TokenDestroy at _ -> unbound at
-    DLE_TimeOrder at op ps -> do
+    DLE_TimeOrder at op p -> do
       let go f n se = do
             n' <- smt_v at n
             smtAssert $ f n' se
-      forM_ ps $ \case
+      case p of
         (Nothing, n) -> go smtEq n $ smt_lt at $ DLL_Int at 0
         (Just o, n) -> do
```
jeapostrophe, line 1117: These both name the other thing `n`

jeapostrophe:
@willghatch @keeganthomp You can now use `this` in the randomness

PR #749, opened 2022-03-31
CORE-663 daily devnet ping

erer1243:
wip
todo: infer reach version (how?), infer reach userId (should this be the result of whoami at container build time?) 

because docker refuses to accept any files outside of the build context, 3 copies of the script are seemingly needed (or, that's the simplest way)


scripts/devnet-algo/daily-ping.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+START_TIME_UNIX="$(date +%s)"
+START_TIME_UTC="$(date --utc --iso-8601=seconds)Z"
+
+ping_devnet() {
+  TIME_NOW="$(date +%s)"
+  ELAPSED="$(( "$TIME_NOW" - "$START_TIME_UNIX" )).0"
+  REPORT="{
+    \"userId\": \"ALGO-devnet\",
```
jeapostrophe, line 10: If this is the userId, then it will be useless

jeapostrophe (changes requested):
You should NOT make 3 copies. You should make one script and put as part of the `Makefile`, copying it and (potentially customizing it).

You should look at how `reach` exposes the id and how `reachc` gets (potentially asking @mattaudesse ) and get it the same way. Similarly, the version should be available through an environment variable too

jeapostrophe (commented):
These appears ready for now and I think we should just do it and if Matt thinks there's a better way to do it integrated into the HS, he should make a new Jira task for that

mattaudesse (approved):
👍🏼

jeapostrophe (approved):

PR #753, opened 2022-04-01
docs: add hiddenNote and alongside

willghatch:
This adds two features:

The `hiddenNote` directive makes a button that toggles a note (like the `note` directive) to appear below it.
The label field [in brackets] determines the text of the button.

Example:

```
:::hiddenNote[Label of button here]
This text is hidden until the user clicks the button, and is styled like `note`.
Frankly, I would recommend that you use `note` instead unless the digression is particularly large and tangential.
But I suppose that's largely a matter of taste.
:::
```

Next this also adds the `alongside` directive (with it's subordinate helper, `alongsideColumn`.
The `alongside` directive lets you put blocks of content side-by-side when there is enough horizontal screen space.

Example:

```
::::alongside
:::alongsideColumn
Here is some text that will be laid out on the left.
:::
:::alongsideColumn
Here is some text that will be laid out on the right.
:::
::::
```

You can have an arbitrary number of `alongsideColumn`s, but they will eventually wrap to multiple rows even on a wide screen.
On a thin screen (eg. a portrait-mode phone browser or desktop browser tiled to the left or right half of a screen), these will probably ultimately end up vertically anyway.

### DIFF
--- a/docs/dev/generator.js
+++ b/docs/dev/generator.js
@@ -423,11 +423,44 @@ const processMd = async ({baseConfig, relDir, in_folder, iPath, oPath}) => {
     return `<a class="ui-tooltip" title="${hoverText}"><span style="cursor">${text}</span></a>`;
   };
 
-  const directive_note = (node) => {
-    const data = node.data;
-    data.hName = "div";
-    data.hProperties = { class: "note" };
+  const makeDirectiveWithClass = (classString) => {
+    return (node) => {
+      const data = node.data;
+      data.hName = "div";
+      const attrs = node.attributes || {};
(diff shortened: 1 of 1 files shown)


willghatch:
@Craig-Elre Will this suit your needs?  Frankly instead of `hiddenNote`, I prefer a regular note that hyperlinks to a larger explanation if it's not short, but I'm a weirdo with strange preferences.

jeapostrophe:
I mostly agree with @willghatch ; the main thing that is good about something hidden is to reveal a "solution"

jeapostrophe (approved):

PR #754, opened 2022-04-02
Update rpc_server.ts

xarmian:
## Summary

The rpc server transfer function does not pass through the fourth argument (token: non-network token id) to the stdlib transfer function. This change adds the token argument for the non-token token id, which is needed to transfer non-network tokens.

No design changes

Untested

No doc changes should be needed


jeapostrophe (approved):

jeapostrophe:
Thank you!

PR #758, opened 2022-04-04
Add solidity reserved words

willghatch:
I'm not sure these are technically reserved words in the Solidity language, but they are keys in the resulting contract objects.
Without adding these to the list, they can be used as view names.
When used as view names, they conflict with other keys in the contract object and break.


willghatch:
I think these keys are probably added by the solidity compiler.  If that's the case, maybe we should add a test that compiles a simple program with no views or other features that would be spliced in to the contract object keys, then looks at the keys on the resulting contract object and checks that all keys that don't start with an underscore are included in this list.

jeapostrophe (approved):

PR #760, opened 2022-04-05
CORE-24 Warn about unconstrained maps

erer1243:
todo: Write docs for new RW0007 - what should it say? Why is an unconstrained map a bad thing?

### DIFF
--- a/examples/unconstrained-map-warning/index.mjs
+++ b/examples/unconstrained-map-warning/index.mjs
@@ -1,0 +1,1 @@
+// index.rsh is not used in this example
--- a/hs/src/Reach/Verify/SMT.hs
+++ b/hs/src/Reach/Verify/SMT.hs
@@ -38,6 +38,7 @@ import Reach.Util
 import Reach.Verify.SMTAst
 import Reach.Verify.SMTParser
 import Reach.Verify.Shared
+import Reach.Warning
 import SimpleSMT (Logger (Logger), Result (..), SExpr (..), Solver (..))
 import qualified SimpleSMT as SMT
 import qualified System.Console.Pretty as TC
@@ -188,6 +189,7 @@ data SMTCtxt = SMTCtxt
   , ctxt_inv_mode :: BlockMode
   , ctxt_pay_amt :: Maybe (SExpr, SExpr)
   , ctxt_smt_trace :: IORef [SMTLet]
+  , ctxt_constrained_maps :: IORef (S.Set DLMVar)
   }
 
 ctxt_mode :: App VerifyMode
@@ -1327,7 +1329,10 @@ smt_n = \case
     um <- asks ctxt_untrustworthyMaps
     when um $ smtMapRefresh at
     smt_s s
-  LLC_While at asn inv cond body k ->
+  LLC_While at asn inv cond body k -> do
+    constrainedMapsRef <- asks ctxt_constrained_maps
+    liftIO $ modifyIORef' constrainedMapsRef insertConstrainedMaps
+
     mapM_ ctxtNewScope [before_m, loop_m, after_m]
     where
       with_inv = local (\e -> e {ctxt_while_invariant = Just inv})
@@ -1344,6 +1349,20 @@ smt_n = \case
         smt_invblock (B_Assume True) inv
         smt_invblock (B_Assume False) cond
         smt_n k
+
+      insertConstrainedMaps :: S.Set DLMVar -> S.Set DLMVar
+      insertConstrainedMaps = insertConstrainedMaps' invTail
+      DLBlock _ _ invTail _ = inv
+
+      insertConstrainedMaps' :: DLTail -> S.Set DLMVar -> S.Set DLMVar
+      insertConstrainedMaps' dltail maps = case dltail of
+        DT_Return _ -> maps
+        DT_Com stmt dltail' ->
+          let tailMaps = insertConstrainedMaps' dltail' maps in
+            case stmt of
+              DL_MapReduce _ _ _ dlmv _ _ _ _ -> S.insert dlmv tailMaps
+              _ -> tailMaps
+
   LLC_Continue _at asn -> smt_while_jump True asn
   LLC_ViewIs _ _ _ ma k -> do
     maybe mempty smt_eb ma
@@ -1610,6 +1629,7 @@ _verify_smt mc ctxt_vst smt lp = do
   let ctxt_untrustworthyMaps = llo_untrustworthyMaps
   let ctxt_pay_amt = Nothing
   ctxt_smt_trace <- newIORef mempty
+  ctxt_constrained_maps <- newIORef mempty
   flip runReaderT (SMTCtxt {..}) $ do
     let defineMap (mpv, SMTMapInfo {..}) = do
           mi <- liftIO $ incCounter sm_c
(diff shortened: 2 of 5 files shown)


examples/unconstrained-map-warning/index.mjs
```diff
@@ -0,0 +1 @@
+// index.rsh is not used in this example
```
jeapostrophe, line 1: Typo

jeapostrophe: But also, this should just be in hs/t/y if there's no runtime component

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1344,6 +1349,20 @@ smt_n = \case
         smt_invblock (B_Assume True) inv
         smt_invblock (B_Assume False) cond
         smt_n k
+
```
jeapostrophe, line 1352: Heathen! Extra confusing newline!

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1344,6 +1349,20 @@ smt_n = \case
...
+      insertConstrainedMaps :: S.Set DLMVar -> S.Set DLMVar
+      insertConstrainedMaps = insertConstrainedMaps' invTail
+      DLBlock _ _ invTail _ = inv
+
```
jeapostrophe, line 1356: Ditto

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1647,6 +1667,13 @@ _verify_smt mc ctxt_vst smt lp = do
     let ms = [VM_Honest, VM_Dishonest RoleContract] -- <> (map (VM_Dishonest . RolePart) $ M.keys pies_m)
     mapM_ smt_s_top ms
```
jeapostrophe, line 1669: Yuck!

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1327,7 +1329,10 @@ smt_n = \case
     um <- asks ctxt_untrustworthyMaps
     when um $ smtMapRefresh at
     smt_s s
-  LLC_While at asn inv cond body k ->
+  LLC_While at asn inv cond body k -> do
+    constrainedMapsRef <- asks ctxt_constrained_maps
+    liftIO $ modifyIORef' constrainedMapsRef insertConstrainedMaps
+
```
jeapostrophe, line 1335: Yuck!

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1647,6 +1667,13 @@ _verify_smt mc ctxt_vst smt lp = do
     let ms = [VM_Honest, VM_Dishonest RoleContract] -- <> (map (VM_Dishonest . RolePart) $ M.keys pies_m)
     mapM_ smt_s_top ms
 
+    let allMaps = S.fromList $ M.keys dli_maps
+    constrainedMaps <- liftIO $ readIORef ctxt_constrained_maps
+    let unconstrainedMaps = S.difference allMaps constrainedMaps
+    liftIO $ flip mapM_ unconstrainedMaps $ \m -> do
```
jeapostrophe, line 1673: They should actually be mentioned in EACH `while`, not in ANY `while`

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1344,6 +1349,20 @@ smt_n = \case
...
+
+      insertConstrainedMaps' :: DLTail -> S.Set DLMVar -> S.Set DLMVar
+      insertConstrainedMaps' dltail maps = case dltail of
+        DT_Return _ -> maps
+        DT_Com stmt dltail' ->
+          let tailMaps = insertConstrainedMaps' dltail' maps in
+            case stmt of
+              DL_MapReduce _ _ _ dlmv _ _ _ _ -> S.insert dlmv tailMaps
+              _ -> tailMaps
+
```
jeapostrophe, line 1365: :shiver:

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1344,6 +1349,20 @@ smt_n = \case
...
+      insertConstrainedMaps :: S.Set DLMVar -> S.Set DLMVar
+      insertConstrainedMaps = insertConstrainedMaps' invTail
+      DLBlock _ _ invTail _ = inv
+
+      insertConstrainedMaps' :: DLTail -> S.Set DLMVar -> S.Set DLMVar
+      insertConstrainedMaps' dltail maps = case dltail of
+        DT_Return _ -> maps
+        DT_Com stmt dltail' ->
+          let tailMaps = insertConstrainedMaps' dltail' maps in
```
jeapostrophe, line 1361: We don't use `let ... in`

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1344,6 +1349,20 @@ smt_n = \case
...
+            case stmt of
+              DL_MapReduce _ _ _ dlmv _ _ _ _ -> S.insert dlmv tailMaps
```
jeapostrophe, line 1363: I think it would be better to stick something in the ctxt and then in the code that runs on `DL_MapReduce`, record that it was looked at it so there's no another walking of the code

jeapostrophe (changes requested):
An unconstrained map is dangerous because it means that there are no constraints on what it contains so there is probably some invariant about it that the programmer has in their head that they haven't written down

examples/unconstrained-map-warning/index.mjs
erer1243: I also figured it might as well be a test at this point.

hs/src/Reach/Verify/SMT.hs
erer1243: ah but i'm always so confused by how congealed the code is!! I can't tell quickly where ideas separate

hs/src/Reach/Verify/SMT.hs
erer1243: I did remember that rule, but I wonder why we don't use it? Eg here there seems to be no better way to avoid code repetition.

jeapostrophe: It makes the indentation weird, so we use `do` or `where` instead

hs/src/Reach/Verify/SMT.hs
erer1243: gotcha

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1330,10 +1333,12 @@ smt_n = \case
...
   LLC_While at asn inv cond body k -> do
-    constrainedMapsRef <- asks ctxt_constrained_maps
-    liftIO $ modifyIORef' constrainedMapsRef insertConstrainedMaps
-
+    mclRef <- asks ctxt_maps_constrained_local
+    liftIO $ writeIORef mclRef S.empty
     mapM_ ctxtNewScope [before_m, loop_m, after_m]
+    mcl <- liftIO $ readIORef mclRef
+    mcaRef <- asks ctxt_maps_constrained_always
+    liftIO $ modifyIORef' mcaRef $ M.mapWithKey $ \m _ -> S.member m mcl
```
jeapostrophe, line 1341: I don't believe that this implementation works. You should make some more tests

jeapostrophe (changes requested):

hs/src/Reach/Verify/SMT.hs
erer1243: You're right, it doesn't. It warns about maps that cannot be constrained in a given invariant because they were defined later. I'm not sure how to glean which maps have been defined so far at a given MapReduce

jeapostrophe: Start with more tests that demonstrate the problem. Then I think we'll need to add some sort of counter for invariants (loops) and remember that in the definition of a map, then record which whiles check each map, and then warn when one is missing a while after the one it started on

PR #763, opened 2022-04-05
CORE-928: Overloaded APIs and Remote Functions

chrisnevers:


hs/src/Reach/AST/DK.hs
```diff
@@ -104,10 +104,10 @@ type DKExports = M.Map SLVar DKExportBlock
 type DKViews = DLViews
 
 data DKProg
-  = DKProg SrcLoc DLOpts SLParts DLInit DKExports DKViews DLAPIs DLEvents DKTail
+  = DKProg SrcLoc DLOpts SLParts DLInit DKExports DKViews DLAPIs Aliases DLEvents DKTail
```
jeapostrophe, line 107: Can you make a Jira card to turn all of the `*Prog` datas to records so we can name these things?

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2967,8 +2980,13 @@ doWrapData tys mk = do
 
 cmeth :: Int -> CMeth -> App ()
 cmeth sigi = \case
-  CApi who sig _ which tys doWrap -> do
-    block_' (bunpack who) $ do
+  CAlias alias (CApi {..}) -> do
```
jeapostrophe, line 2983: Interesting... you're making the real thing available also AND the alias. That makes sense

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1367,7 +1368,8 @@ apiDef who ApiInfo {..} = do
   ret_ty' <- solType_ ai_ret_ty
   let ret_ty'' = ret_ty' <+> withArgLoc ai_ret_ty
   let ret = "external payable returns" <+> parens ret_ty''
-  return $ solFunction (pretty who_s) argDefns ret body
+  let mk w = solFunction (pretty w) argDefns ret body
+  return $ vsep $ mk who_s : maybe [] ((: []) . mk . bunpack) ai_alias
```
jeapostrophe, line 1372: It looks to me like on ETH you are not giving access to the Reach name AND the alias like you did on ALGO. You should decide if this is what we want and add that to the example

examples/api-overload/client.rsh
```diff
@@ -0,0 +1,38 @@
+'reach 0.1';
```
jeapostrophe, line 1: I would put these in the same `rsh` file and just have two exports

jeapostrophe (changes requested):

hs/src/Reach/Connector/ETH_Solidity.hs
chrisnevers: It's generating both the reach name and alias; it's calling `mk` twice, one for the og name, one for the alias, if it exists.

chrisnevers:
I put `nub` into UnsafeUtils. I didn't want to use a `Set` when verifying that the aliases are unique because it would require adding `Ord` to pretty much everything in the source language and JS ASTs

hs/src/Reach/Connector/ETH_Solidity.hs
jeapostrophe: Ah, gross, this is duplicating the code then? Do we know the code is small?

docs/src/rsh/appinit/index.md
```diff
@@ -120,8 +120,9 @@ API({ add2: Fun([UInt, UInt], UInt), add1: Fun([UInt], UInt) }, { add1: "add", a
 
 APIs are functions that can be called by other contracts, as well as off-chain.
 
-An API is defined with `{!rsh} API(apiName, apiInterface, apiAlias)` or `{!rsh} API(apiInterface, apiAlias)`, where `{!rsh} apiName` is a string that labels the API, `{!rsh} apiInterface` is an object where each field indicates the type of a function provided by the contract as an API, and `{!rsh} apiAlias` is an object that maps function names from the `{!rsh} apiInterface` to an alias.
-This @{defn("function alias")} allows overloaded methods to be created. Many functions may map to the same alias as long as each function domain is unique.
+An API is defined with `{!rsh} API(apiName, apiInterface, ?apiAlias)` or `{!rsh} API(apiInterface, apiAlias)`, where `{!rsh} apiName` is a string that labels the API, `{!rsh} apiInterface` is an object where each field indicates the type of a function provided by the contract as an API, and `{!rsh} apiAlias` is an optional object that maps function names from the `{!rsh} apiInterface` to an alias.
```
jeapostrophe, line 123: The second `apiAlias` is not optional?

hs/src/Reach/Connector/ETH_Solidity.hs
chrisnevers: It's pretty small:
```
function mul2(uint256 _a0, uint256 _a1)  external payable returns (uint256 ) {
    ApiRng memory _r;
    T8 memory _t;
    _t.msg = T7(T6(_a0, _a1));
    _reach_m2(_t, _r);
    return _r.mul2;
    }
  function mul(uint256 _a0, uint256 _a1)  external payable returns (uint256 ) {
    ApiRng memory _r;
    T8 memory _t;
    _t.msg = T7(T6(_a0, _a1));
    _reach_m2(_t, _r);
    return _r.mul2;
    }
```

but I could make it like:
```
function mul(uint256 _a0, uint256 _a1)  external payable returns (uint256 ) {
    return this.mul2(_a0, _a1);
    }
```

jeapostrophe (approved):

hs/src/Reach/Connector/ETH_Solidity.hs
jeapostrophe: I think that would be better

jeapostrophe: Sorry :(

jeapostrophe (changes requested):
!

jeapostrophe (approved):

PR #765, opened 2022-04-05
add walletFallback opt for existing wc client

xarmian:
This change allows a WalletConnectClient object, instantiated outside of the Reach stdlib, to be passed to the walletFallback() function as follows:

```
import WalletConnect from '@walletconnect/client';
import { ALGO_WalletConnect } from '@reach-sh/stdlib';
let wcClient = new WalletConnect();
stdlib.setWalletFallback(stdlib.walletFallback({ providerEnv: 'TestNet', WalletConnect: ALGO_WalletConnect, connector: wcClient }));
```

## Summary

The intention is to be able to pass an existing walletConnect client session (i.e. if a walletconnect session was recreated from browser session storage, or passed to a backend nodejs function) in place of a newly instantiated client session. I may have missed it, but I could not find another way to make this work.

It's born from a desire to be able to take an existing walletConnect session object from local browser storage, pass it to a backend nodejs function, re-create the client session, and then use that session within the reach standard library.

Testing was performed on a transpiled version of the ALGO.ts (ALGO.js) file with a nodejs backend with Google Cloud Functions. However the direct modification of the TypeScript file is untested.

Documentation for WalletConnect should be updated with an example like the one above.

### DIFF
--- a/js/stdlib/ts/ALGO.ts
+++ b/js/stdlib/ts/ALGO.ts
@@ -939,9 +939,10 @@ const walletFallback_MyAlgoWallet = (MyAlgoConnect: unknown, opts: object) => ()
   };
   return doWalletFallback_signOnly(opts, getAddr, signTxns);
 };
-const walletFallback_WalletConnect = (WalletConnect:any, opts:object) => (): ARC11_Wallet => {
+const walletFallback_WalletConnect = (WalletConnect:any, opts:any) => (): ARC11_Wallet => {
   debug(`using WalletConnect wallet fallback`);
   const wc = new WalletConnect();
+  if (opts.connector) wc.wc = opts.connector;
   return doWalletFallback_signOnly(opts, (() => wc.getAddr()), ((ts) => wc.signTxns(ts)));
 };
 export const walletFallback = (opts:any) => {


jeapostrophe:
Does it work for you re: loading for storage? I wonder if I can include a test about that on my side

xarmian:
It does (surprisingly?) work, however I'm not convinced it's the best way to solve my particular problem -- in my case I'm using Flutter/Dart for the frontend and Google Cloud Functions (with NodeJS) for a centralized backend, and passing the session from the Browser to the Cloud Function. I'm then using the standard library to interact with the Reach contract's APIs.

To reproduce this scenario, I've created a simple test example here (https://github.com/xarmian/wc-example) that consists of a NodeJS server which listens on localhost:3001 (server.js) and a shell script that uses curl to pass a previously established WalletConnect session (which I pulled from the browser's local storage) in an HTTP POST to the nodejs server.

If you start the server with `node server.mjs` and then run `./test.sh` it should reproduce the scenario.

If you use an unmodified version of the Reach stdlib, you'll get a QR code output to the console of the server. But if you modify line 1025 of node_modules/@reach-sh/stdlib/dist/cjs/ALGO.js (this line # is from the 0.1.9-rc.7 distribution package) and add the following line:

   var wc = new WalletConnect();
    **if (opts.connector) wc.wc = opts.connector;**
    return doWalletFallback_signOnly(opts, (function () { return wc.getAddr(); }), (function (ts) { return wc.signTxns(ts); })); };

It should give you the following output from the server:

```Listening at http://localhost:3001
POST /
WalletConnect Account is: QUCT2PQNESPFJZR56MY7UCKG6NTJKCT6MUBYDGSJG4ET3BVMK6JDVCOSSM
AWC ctor
AWC ensureWC
AWC session exists
AWC getAddr [ 'QUCT2PQNESPFJZR56MY7UCKG6NTJKCT6MUBYDGSJG4ET3BVMK6JDVCOSSM' ]
Connected account from Reach is 0x85053d3e0d249e54e63df331fa0946f366950a7e6503819a4937093d86ac5792
Reach account formatted is: QUCT2PQNESPFJZR56MY7UCKG6NTJKCT6MUBYDGSJG4ET3BVMK6JDVCOSSM
```

This all being said, the change I'm proposing could be used simply to allow a developer to create a WalletConnect session outside of Reach in order to utilize a custom clientMeta parameter (to provide an app name, url, and image).

One correction to my original comment is that, in my example, I'm importing WalletConnectClient and using `const connector = new WalletConnectClient.default({bridge: 'https://bridge.walletconnect.org', session: wcSession});` because `new WalletConnect(...);` fails with `WalletConnect is not a constructor`. I haven't had a chance to dive into that further.

jeapostrophe:
I've finally implemented this. You may enjoy checking out this example program: https://github.com/reach-sh/reach-lang/tree/master/examples/algo-wallet-demo

PR #766, opened 2022-04-06
CORE-896: Support Arbitrary Map Keys on ETH

chrisnevers:


docs/src/rsh/consensus/index.md
```diff
@@ -523,19 +523,23 @@ This operation may not be used with `{!rsh} REMOTE_FUN.bill`.
...
-Such dereferences return a value of type `{!rsh} Maybe(TYPE_EXPR)`, because the mapping may not contain a value for `{!rsh} ADDR_EXPR`.
+This returns a value which may be used to dereference particular mappings via `{!rsh} map[EXPR]`.
+Such dereferences return a value of type `{!rsh} Maybe(VAL_TYPE_EXPR)`, because the mapping may not contain a value for `{!rsh} EXPR`.
 
-A mapping may be modified by writing `{!rsh} map[ADDR_EXPR] = VALUE_EXPR` to install `{!rsh} VALUE_EXPR` (of type `{!rsh} TYPE_EXPR`) at `{!rsh} ADDR_EXPR`, or by writing `{!rsh} delete map[ADDR_EXPR]` to remove the mapping entry.
+A mapping may be modified by writing `{!rsh} map[EXPR] = VALUE_EXPR` to install `{!rsh} VALUE_EXPR` (of type `{!rsh} VAL_TYPE_EXPR`) at `{!rsh} EXPR`, or by writing `{!rsh} delete map[EXPR]` to remove the mapping entry.
 Such modifications may only occur in a consensus step.
 
+:::note
+Algorand only supports `{!rsh} KEY_TYPE_EXPR`s of type `{!rsh} Address`.
```
jeapostrophe, line 540: I'd put this in the discussion of the networks

examples/algo-try-csp/index.rsh
```diff
@@ -10,7 +10,7 @@ const makeMain = (untrustworthyMaps) => Reach.App(() => {
   init();
 
   A.publish();
-  const m = new Map(UInt);
+  const m = new Map(Address, UInt);
```
jeapostrophe, line 13: Revert all these

hs/src/Reach/Verify/SMT.hs
```diff
@@ -140,6 +140,7 @@ instance Monoid a => Monoid (App a) where
 data SMTMapInfo = SMTMapInfo
```
jeapostrophe, line 140: SMT is going to confuse arrays and maps keyed on numbers. Write some tests that generate errors and you'll see this and figure out a solution

chrisnevers:
@jeapostrophe  What do you mean by N+2 relations?

hs/src/Reach/Verify/SMT.hs
chrisnevers: I think this actually won't happen because when we generate the SMT arrays, we use SMT's `Int` type for `Array`s & and theres no Reach type that converts to an SMT `Int`, so Map would never use it

jeapostrophe: `UInt` (and `UInt256`) are `Int`. They are just type aliases (called sort aliases) for `Int` and when we look at the model, they are expressed as `Int`, not `UInt`

jeapostrophe:
> @jeapostrophe What do you mean by N+2 relations?

An old Map is a 2-relation of (Address, Value). Your new Maps are 2-relations of (Key, Value). Some people want 3-relations of (Address, Address, Value) for instance, which would be `(Tuple(Address, Address), Value)`. In Solidity, most people do 3-relations by nested mappings.

jeapostrophe (approved):
Approved assuming CI passes

PR #767, opened 2022-04-06
WIP - make api test for more than 15 arguments

willghatch:
I'm a little confused.

I made this example that uses 20 arguments.  Somewhere between 8 and 16 arguments the Solidity compiler complains about stack depth.  At 16+, I expected Algorand to fail, since this ticket is about making ALGO work with 16+ args.  But... it seems to run just fine with 16 or 20.  So I'm wondering if I'm still misunderstanding something fundamental about what I'm trying to do.


willghatch:
@jeapostrophe

examples/api-arg-many/index.rsh
```diff
@@ -0,0 +1,46 @@
...
+                  UInt, UInt, UInt, UInt,
+                  // I wanted to test more than 15, but when 16 worked I
+                  // decided to keep going higher.
+                  UInt, UInt, UInt, UInt,
+                 ],
+                 UInt)
+  })
+  setOptions({
+    connectors: [
+      // With 16 arguments, the Solidity compiler errors saying that the stack is too deep.
```
jeapostrophe, line 20: You should fix that

jeapostrophe (commented):
The program actually ran? or it just compiled okay?

https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Connector/ALGO.hs#L3019 --- This line should either fail to compile or fail to execute

https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L4341-L4345

willghatch:
It actually ran.  It ran with `REACH_CONNECTOR=ALGO`, and in `test.sh` with `CI ALGO api-arg-many`.  I think both of those are running with our devnet, and I'm not sure how the devnet is different from the real thing, but I wouldn't think it would affect this.

willghatch:
I did look around the Algorand source a bit, but while it looks like it should fail there, I've never read their source.  I don't know if that code is ever even executed, or whether `maxAppArgs` corresponds to what I would think it does.  So... I thought I would ask you about it.  But it runs and sums things correctly and passes the test.

jeapostrophe:
Can you post the generated `.appApproval.teal` in the PR or in a gist?

willghatch:
Here you go:

```
#pragma version 6
// resource: R_Txn
int 0
store 0 // GV_txnCounter :: UInt
txn ApplicationID
bz alloc
byte base64()
app_global_get
dup
int 0
extract_uint64
store 1 // GV_currentStep :: UInt
int 8
extract_uint64
store 2 // GV_currentTime :: UInt
txna ApplicationArgs 0
btoi
preamble:
dup
bz publish
api:
int 0
store 4 // GV_argTime :: UInt
int 1
store 6 // GV_wasMeth :: Bool
int 4196147840
==
assert

  // { l0_A_sumMany
  l0_A_sumMany:
  // API: A_sumMany(uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64,uint64)uint64
  //  ui: 4196147840
  txna ApplicationArgs 1
  txna ApplicationArgs 2
  concat
  txna ApplicationArgs 3
  concat
  txna ApplicationArgs 4
  concat
  txna ApplicationArgs 5
  concat
  txna ApplicationArgs 6
  concat
  txna ApplicationArgs 7
  concat
  txna ApplicationArgs 8
  concat
  txna ApplicationArgs 9
  concat
  txna ApplicationArgs 10
  concat
  txna ApplicationArgs 11
  concat
  txna ApplicationArgs 12
  concat
  txna ApplicationArgs 13
  concat
  txna ApplicationArgs 14
  concat
  txna ApplicationArgs 15
  concat
  txna ApplicationArgs 16
  concat
  txna ApplicationArgs 17
  concat
  txna ApplicationArgs 18
  concat
  txna ApplicationArgs 19
  concat
  txna ApplicationArgs 20
  concat
  b publish1
  // l0_A_sumMany }

publish:
txna ApplicationArgs 2
btoi
store 4 // GV_argTime :: UInt
txna ApplicationArgs 3
txna ApplicationArgs 1
btoi
dup
int 1
<
bnz l1_publish_lt_1
int 1
==
assert

  // { publish1
  publish1:
  // check step
  int 1
  load 1 // GV_currentStep :: UInt
  ==
  assert
  // check time
  load 4 // GV_argTime :: UInt
  dup
  int 0
  ==
  swap
  load 2 // GV_currentTime :: UInt
  ==
  ||
  assert
  dup
  store 5 // GV_argMsg :: Null
  store 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  byte base64(LSE92g==)
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  concat
  log // up to 164 bytes
  api_A_sumMany:
  // ^ CheckPay
  // at ./index.rsh:31:3:dot
  byte base64(AAAAAAAAAKc=)
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 0
  extract_uint64
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 8
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 16
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 24
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 32
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 40
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 48
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 56
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 64
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 72
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 80
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 88
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 96
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 104
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 112
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 120
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 128
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 136
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 144
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 152
  extract_uint64
  +
  itob
  concat
  log // up to 16 bytes
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 0
  extract_uint64
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 8
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 16
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 24
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 32
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 40
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 48
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 56
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 64
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 72
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 80
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 88
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 96
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 104
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 112
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 120
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 128
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 136
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 144
  extract_uint64
  +
  load 255 // .api27.dom/125 :: Tuple(UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt)
  int 152
  extract_uint64
  +
  itob
  store 7 // GV_apiRet :: Null
  b updateStateHalt
  // publish1 }

l1_publish_lt_1:
pop

  // { publish0
  publish0:
  // check step
  int 0
  load 1 // GV_currentStep :: UInt
  ==
  assert
  // check time
  load 4 // GV_argTime :: UInt
  dup
  int 0
  ==
  swap
  load 2 // GV_currentTime :: UInt
  ==
  ||
  assert
  byte base64(Xw2r+g==)
  log // up to 4 bytes

    // { checkTxn
    int 100000
    callsub l2_LF_checkTxn_net
    // checkTxn }

  // ^ CheckPay
  // at ./index.rsh:27:12:dot
  // ^ CheckPay
  // at ./index.rsh:27:12:dot
  int 1
  store 1 // GV_currentStep :: UInt
  global Round
  store 2 // GV_currentTime :: UInt
  b updateStateNoOp
  // publish0 }

updateStateHalt:
txn OnCompletion
int DeleteApplication
==
assert
// checked on completion

  // { makeTxn
  itxn_begin
  int 0
  itxn_field Fee
  int 0
  itxn_field Amount
  int pay
  itxn_field TypeEnum
  global CreatorAddress
  itxn_field CloseRemainderTo
  global CurrentApplicationAddress
  itxn_field Receiver
  itxn_submit
  // makeTxn }

b updateState
updateStateNoOp:
txn OnCompletion
int NoOp
==
assert
// checked on completion
updateState:
byte base64()
load 1 // GV_currentStep :: UInt
itob
load 2 // GV_currentTime :: UInt
itob
concat
app_global_put
load 6 // GV_wasMeth :: Bool
bz checkSize
apiReturn_noCheck:
byte base64(FR98dQ==)
load 7 // GV_apiRet :: Null
concat
log // up to 12 bytes
checkSize:
load 0 // GV_txnCounter :: UInt
dup
int 1
+
global GroupSize
==
assert
txn GroupIndex
==
assert
done:
int 1
return
apiReturn_check:
txn OnCompletion
int NoOp
==
assert
// checked on completion
b apiReturn_noCheck
alloc:
int 0
store 1 // GV_currentStep :: UInt
int 0
store 2 // GV_currentTime :: UInt
b updateStateNoOp
l2_LF_checkTxn_net:
// resource: R_Txn
load 0 // GV_txnCounter :: UInt
dup
dup2
int 1
+
store 0 // GV_txnCounter :: UInt
gtxns Receiver
global CurrentApplicationAddress
==
assert
gtxns TypeEnum
int pay
==
assert
gtxns Amount
==
assert
retsub
```

jeapostrophe:
Yea, this is really weird. Here's another place that gets checked...

https://github.com/algorand/go-algorand/blob/1504f5f39ec2893d2b0ac9b60d1b87f6e7d4e2fc/data/transactions/transaction.go#L407

Can you run your program, with REACH_DEBUG=Y before starting the devnet, then run `reach-algo-dbg` and we can see the txn? We should assemble an error to tell Algorand about.

Next, you can make progress by doing something like https://github.com/reach-sh/reach-lang/tree/master/examples/remote-rsh but where the server has 17 arguments. I'm pretty sure you'll see an error there

examples/api-arg-many/index.rsh
willghatch: I've been looking at this. I don't know if we can get around this in Solidity.  A relevant issue I came across is [1], which says that it's a problem with accessing more than 16 things on the stack that could be solved with this draft EIP [2].  That tells me that we can use workarounds discussed in those pages if we generate EVM code directly, but probably not with Solidity.  That said, I don't know that much about Solidity.  Can we play a trick in Solidity like turning one item on the stack into a tuple or something?  If so then maybe we can just do that.

[1] https://github.com/ethereum/solidity/issues/7624
[2] https://eips.ethereum.org/EIPS/eip-663

jeapostrophe: We already do stuff like that. That's what memvars and the depthOf type class are about. We just need to use them more effectively for arguments.

jeapostrophe: (It may be that the way to do this is for Reach/ETH to do the same thing as Reach/ALGO where if you make more arguments we tuple the last one, but I don't know what the rest of the ETH world does)

willghatch:
Ok, I've got `reach-algo-dbg` working now, but I don't really know what I'm looking at.  It looks like I run `reach-algo-dbg` once to get a list of transactions, then I can pass the name of one in to see the transaction json.

So, (1) what should I be looking for to find the transaction, and (2) is there a way to just see them all and search them, because I don't want to look through 200 transactions one-by-one.  Or in other words, what is the expected way to use `reach-algo-dbg`?

willghatch:
I'm having trouble figuring out why my new `remote-arg-many` test isn't working when it has only 8 args.  I get this error:

```
/stdlib/dist/cjs/shared_impl.js:306
                            fail(new Error("".concat(bl, " errored with ").concat(err)));
                                 ^

Error: A.poke errored with Error: API call failed: {
  "type": "signAndPost",
  "e": {
    "status": 400,
    "response": {
      "message": "TransactionPool.Remember: transaction AMLVXFXO6JWO2DJD323S5WS7WQDYGM3AFJOESEUCC4VHKWKTJ5IA: logic eval error: logic eval error: assert failed pc=55. Details: pc=55, opcodes=pushint 2807893464\n==\nassert\n. Details: pc=247, opcodes=extract 56 8\nitxn_field ApplicationArgs\nitxn_submit\n"
    }
  },
  "es": "Error: Network request error. Received status 400: TransactionPool.Remember: transaction AMLVXFXO6JWO2DJD323S5WS7WQDYGM3AFJOESEUCC4VHKWKTJ5IA: logic eval error: logic eval error: assert failed pc=55. Details: pc=55, opcodes=pushint 2807893464\n==\nassert\n. Details: pc=247, opcodes=extract 56 8\nitxn_field ApplicationArgs\nitxn_submit\n"
}
    at /stdlib/dist/cjs/shared_impl.js:306:34
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
ERROR: 1
```

Am I doing something wrong with remote or API calls?

examples/remote-arg-many/index.rsh
```diff
@@ -0,0 +1,88 @@
...
+
+  Deployer.only(() => {
+    const serverCtcInfo = declassify(interact.serverCtcInfo)
+  })
+
+  Deployer.publish(serverCtcInfo)
+  Deployer.interact.ready()
+
+  const sumServer = remote(serverCtcInfo, {
+    sumMany: manyArgFunSig,
```
jeapostrophe, line 75: You named the function `A_sumMany` on the other side

jeapostrophe (changes requested):

willghatch:
Well, I had high hopes when you pointed out the naming issue.  But even fixing that, I can't seem to get it to work.  I can call the server API directly from the frontend, or if I comment out the remote call and return a constant I can call the `poke` api fine from the client.  The remote call is the one thing that isn't working.  But I don't know why.

jeapostrophe:
What error do you get?

willghatch:
Sorry, should have posted that before:

```
Error: poke errored with Error: API call failed: {
  "type": "signAndPost",
  "e": {
    "status": 400,
    "response": {
      "message": "TransactionPool.Remember: transaction PBBUCPDQSFZ64XMTKB6XE33P4ZCMP2IVO7HMZFE5EBFLFVWVUMBA: logic eval error: logic eval error: assert failed pc=296. Details: pc=296, opcodes=pushint 5\n==\nassert\n. Details: pc=247, opcodes=extract 56 8\nitxn_field ApplicationArgs\nitxn_submit\n"
    }
  },
  "es": "Error: Network request error. Received status 400: TransactionPool.Remember: transaction PBBUCPDQSFZ64XMTKB6XE33P4ZCMP2IVO7HMZFE5EBFLFVWVUMBA: logic eval error: logic eval error: assert failed pc=296. Details: pc=296, opcodes=pushint 5\n==\nassert\n. Details: pc=247, opcodes=extract 56 8\nitxn_field ApplicationArgs\nitxn_submit\n"
}
    at /stdlib/dist/cjs/shared_impl.js:306:34
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
ERROR: 1
```

willghatch:
But it looks... very similar to the previous error to me.

willghatch:
How are APIs addressed?  Looking at the teal code for the client contract, I see a comment about a call to `sumMany`, but I don't see how that's actually addressed in the contract.  I was expecting there to be a string somewhere to access the particular API of the remote contract, but I don't see that anywhere.

jeapostrophe:
> How are APIs addressed? Looking at the teal code for the client contract, I see a comment about a call to `sumMany`, but I don't see how that's actually addressed in the contract. I was expecting there to be a string somewhere to access the particular API of the remote contract, but I don't see that anywhere.

What do you mean by "addressed"? When you call a function `name(dom)rng`, you SHA512/256 that string, take the first 4 bytes of that, and that's the first argument to the call.

willghatch:
I've pushed some more commits, but this is still not ready to merge.

* These commits make ALGO remote API calls work with 16+ args.
* I made an untested change to another place that needs to handle 16+ args, but it wasn't exercised by the code I wrote.  (It was actually the code I wrote first in ALGO, despite it being in the last commit.)  I still need to figure out how to exercise that code and write a test for it.
* I haven't yet addressed the issues with ETH for 16+ args in APIs.

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2028,10 +2028,24 @@ ce = \case
...
+              makeTxn1 "ApplicationArgs"
+        -- If there are more than 15 arguments, args 15+ must be packed as a tuple
+        -- into argument at index 15.  Note that the API id is arg 0, so we have
+        -- a total of up to 16 arguments.
+        case 16 < (length as') of
+          False -> do
+            forM_ as' processArg
+          True -> do
+            let as'15 = take 15 as'
+            let as'More = drop 15 as'
```
jeapostrophe, line 2046: `splitAt`

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2991,7 +3005,11 @@ cmeth sigi = \case
       comment $ LT.pack $ " ui: " <> show sigi
       let f :: DLType -> Integer -> (DLType, App ())
           f t i = (t, code "txna" ["ApplicationArgs", texty i])
-      cconcatbs_ (const $ return ()) $ zipWith f tys [1 ..]
+      -- If there are more than 15 args, args 15+ are packed as a tuple in arg 15.
+      let effectiveTys = case 15 < (length tys) of
+            False -> tys
+            True -> (take 14 tys) <> [T_Tuple $ drop 14 tys]
```
jeapostrophe, line 3011: `splitAt`

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3015,9 +3033,30 @@ cmeth sigi = \case
...
-  foldl' (flip go) m (zip vs [(1::Integer) ..])
+  let goSingle (v, i) = sallocVarLet v False (code "txna" ["ApplicationArgs", texty i] >> cfrombs (varLetType v))
+  let goSingles singles k =
+        foldl' (flip goSingle) k (zip singles [(1 :: Integer) ..])
+  -- When there are more than 15 args, the 15th arg is a tuple of args 15+
+  case 15 < (length vs) of
+    False -> do
+      goSingles vs m
+    True -> do
+      -- TODO - I wrote this thinking it was the code path I needed to edit, but it's not exercised by my test.  I should write a test that exercises it and be sure it works.
```
jeapostrophe, line 3044: This code is used by View functions, so you'd need to make a View function with 15+ args and call it

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2991,7 +3005,11 @@ cmeth sigi = \case
       comment $ LT.pack $ " ui: " <> show sigi
       let f :: DLType -> Integer -> (DLType, App ())
           f t i = (t, code "txna" ["ApplicationArgs", texty i])
-      cconcatbs_ (const $ return ()) $ zipWith f tys [1 ..]
+      -- If there are more than 15 args, args 15+ are packed as a tuple in arg 15.
+      let effectiveTys = case 15 < (length tys) of
+            False -> tys
+            True -> (take 14 tys) <> [T_Tuple $ drop 14 tys]
+      cconcatbs_ (const $ return ()) $ zipWith f effectiveTys [1 ..]
```
jeapostrophe, line 3012: This is very suspicious that it works, because the code that we're jumping to really expects to have all e.g. 20 arguments. I think it is just a coincidence that it works because the format on the single tuple is the same as the 20 all concatenated together. Thus, this code is not type-correct but happens to have the same representation. At the very least, you should comment on that

jeapostrophe (changes requested):
You're going to have to change the way the ABI signature gets constructed. You should test this by making something like `api-raw` for this program with many arguments. You'll see that the `algosdk` library will not generate the call you want. You may need to double check what ARC-4 says... whether it is supposed to change the signature string or not when you do the tupling.

hs/src/Reach/Connector/ALGO.hs
willghatch: I think I don't understand views.  I thought that a view was basically computed client-side after observing the contract state, which doesn't require a publication to request anything of the contract or cause any computation in the contract.  That was what I understood when working on the stdlib code for views.  But this looks like views are more like APIs, that you can publish to the contract to compute something in the consensus state.  So... how do views work?

hs/src/Reach/Connector/ALGO.hs
jeapostrophe: Comment then

hs/src/Reach/Connector/ALGO.hs
jeapostrophe: When you write a View, it is part of the published API and it can be called on-chain.

ALSO, Reach knows how to decode the state and run it locally as JS.

This is because on ALGO, there are no free/read-only functions like there are on EVM, so we had to implement them ourselves. Once ALGO supports them, we'll delete the existing client-side view implementation code and just call the node itself.

willghatch:
See the comment in the last commit.

Anyway, I got stuck on a dumb detail writing the test so this took me forever, but the bit about changing how the ABI signature is constructed is now done.  But... I don't love it.

Anyway, I'll get working on views with many arguments and look for Solidity programs that use many arguments like you recommended.

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3015,9 +3040,29 @@ cmeth sigi = \case
...
+    True -> do
+      -- TODO - I wrote this thinking it was the code path I needed to edit, but it's not exercised by my test.  I should write a test that exercises it and be sure it works.
+      let (vs14, vsMore) = splitAt 14 vs
+      let tupleTy = T_Tuple $ map varLetType vsMore
+      let tupleDupExtract i = do
+            op "dup"
+            cTupleRef (SrcLoc Nothing Nothing Nothing) tupleTy i
+      let goTuple (v, i) = sallocVarLet v False (tupleDupExtract i)
+      let m2 = do
+            -- TODO - instead of this dup/pop dance, I think I can just access arg 15 many times
```
jeapostrophe, line 3059: `txna AppArg i` costs 3 bytes. `dup`/`pop` are 1. They have cost 1.

(A) What you have is --- 3 + 1n + 1

(B) Referencing it over and over is 3n

(C) If you alloc a slot with `salloc_` then you can read it once (3), store it (2), then reference it (2)... the cost is 3 + 2 + 2n

So:

A: 4+n size; 2+n cost
B: 3n size; n cost
C: 5+2n; 2+n cost

So, I think what you did is optimal.

willghatch:
I now have a raw test using a view of many arguments, and I fixed the view generation code.  I've been looking for some Solidity code that uses many args.  In OpenZeppelin the highest number of args I found was 8, and I didn't see any evidence of ever wanting more arguments than used, or tricks to get around any limitations.  So I don't think I've gained any useful info about what we should do about ETH-like and many arguments.

willghatch:
To be clear: everything is ready here as far as ALGO is concerned, but ETH still gives a solidity error.  Should I try that struct construction order swap you mentioned to see if that fixes things up to 16 args, then try tupling arguments after that?

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2028,10 +2028,23 @@ ce = \case
...
+              cconcatbs $ map (\a -> (argTypeOf a, ca a)) tas
+              makeTxn1 "ApplicationArgs"
+        -- If there are more than 15 arguments, args 15+ must be packed as a tuple
+        -- into argument at index 15.  Note that the API id is arg 0, so we have
+        -- a total of up to 16 arguments.
+        case 16 < (length as') of
+          False -> do
+            forM_ as' processArg
+          True -> do
+            let (as'15, as'More) = splitAt 15 as'
```
jeapostrophe, line 2045: Why does this split at 15 but the others split at 14? Because it includes the method number? I believe the conflict you have with ALGO.hs is because I changed this to not have the number

jeapostrophe:
> To be clear: everything is ready here as far as ALGO is concerned, but ETH still gives a solidity error. Should I try that struct construction order swap you mentioned to see if that fixes things up to 16 args, then try tupling arguments after that?

Step 1. Confirm the error is in the place I think it is and not something else. You can get the error, and post a gist of the sol w/ the line of the warning
Step 2. Try constructing the argument differently.
Step 3. If it still doesn't work, then emit a warning like we do in ALGO that says "You can't do that on ETHlike"

hs/src/Reach/Connector/ALGO.hs
willghatch: Yes, because of the method number.  I guess maybe I should rebase this then to solve the conflicts.  But maybe I'll wait until I finish figuring out the solidity stuff.

willghatch:
A note to myself for when I get back to this -- the “APIs are almost ready” commit I just pushed still doesn't change anything about views, which still fail in solidity when they have 16+ args, so I still need to finish that as well before this is ready.

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -800,8 +803,17 @@ solCom = \case
...
-    dom'mem <- mapM (solType_withArgLoc . argTypeOf) as
+    let apiLongArgSplit = length as > apiArgPackCutoff
+    let (asBare, asToPack) = if apiLongArgSplit
+          then splitAt apiArgPackCutoff as else (as, [])
+    asBare' <- mapM solArg asBare
+    asToPack' <- mapM solArg asToPack
+    let asPackedTy = T_Tuple $ map argTypeOf as
+    asPacked <- if apiLongArgSplit
+      then mapM (\_ -> allocVar) [asPackedTy]
+      else return []
```
jeapostrophe, line 814: We avoid `if` and use `case` instead

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -810,7 +822,13 @@ solCom = \case
     rng_ty' <- solType rng_ty
     let rng_ty'mem = rng_ty' <> withArgLoc rng_ty
     f' <- addInterface (pretty f) dom'mem rng_ty'mem
-    let eargs = f' : as'
+    asPackedTy' <- if apiLongArgSplit then solType asPackedTy else return $ ""
```
jeapostrophe, line 825: This `if` is okay because it is one line

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1304,23 +1330,41 @@ apiDef who ApiInfo {..} = do
...
-                _ -> return args
+                  let apiLongArgSplit = length ts > apiArgPackCutoff
+                  let (bareArgs, packedArgs) = if apiLongArgSplit
+                        then splitAt apiArgPackCutoff args else (args, [])
+                  let packedArgLength = length ts - apiArgPackCutoff
+                  let tcDefLift =  pretty $ (show tc') <> " memory _apiArgsPack;"
+                  let fillLift = map (\(arg, i) -> solSet
+                                       ("_apiArgsPack.elem" <> pretty i)
+                                       arg)
+                                 $ zip bareArgs ([0 ..] :: [Int])
```
jeapostrophe, line 1354: `map f $ zip x y = zipWith f x y`

jeapostrophe (commented):
When you are ready with this, I'd like to see the `.sol` that gets generated (perhaps in a `gist`)

PR #769, opened 2022-04-07
CORE-1246: RPC token functions

mattaudesse:
Running `git diff -w` locally might be easier for reviewing since GitHub doesn't seem to be marking some indent changes correctly.


jeapostrophe (changes requested):
Is there something I'm missing about why you have `.id` everywhere? That's not the way that `tokenAccept` works in the normal library and we shouldn't do it here either

DanBurton (approved):

PR #774, opened 2022-04-08
Simulator: skip Interact functions which only return `Null`

chikeabuah:


hs/src/Reach/Simulator/Core.hs
```diff
@@ -546,7 +546,11 @@ instance Interp DLExpr where
             Just v -> return v
             Nothing -> possible $ "DLE_Interact: late api call for " ++ partName'
         Nothing -> do
-          suspend $ PS_Suspend (Just at) (A_Interact slcxtframes partName' str dltype args)
+          case dltype of
+            T_Null -> do
+              return V_Null
```
jeapostrophe, line 551: No log?

jeapostrophe (approved):

hs/src/Reach/Simulator/Core.hs
chikeabuah: if the function input is a variable/identifier in the program, it will show up in the local state details 
if it's a literal, its value is visible in the source code

chikeabuah:
i still have to update the other 21 tests :)

jeapostrophe (approved):

PR #775, opened 2022-04-08
remove pr link

erer1243:
^


jeapostrophe (changes requested):
... but we want the PR link when there is one

jeapostrophe (approved):
Actually, I'll approve for now and you'll fix it later

erer1243:
I didn't know if we wanted it or not. I came up with it because I wished it had that link. I'll put it back in later

PR #786, opened 2022-04-11
CORE-1225 NFT Pawn Shop Example

erer1243:


erer1243:
same 'random' unrelated failure happened twice in a row.. rebased to master in case it somehow branched off of a bad commit

jeapostrophe (changes requested):
The random error is not your fault.

Given that we don't have re-attaching yet, I think that you should translate this to use API calls so people don't have to leave their browser open

examples/nft-pawn-shop/index.mjs
```diff
@@ -26,32 +26,40 @@ const run = async (loanMade) => {
...
+        },
+        loanAmt,
+        interest,
+        deadline,
+        nft: nft.id,
+      }),
+      backend.Lender(ctcLender, {}),
+    ]);
+  } catch (e) {
+    if (e != 'loanMade') {
```
jeapostrophe, line 45: You should use `!==`

examples/nft-pawn-shop/index.rsh
```diff
@@ -30,15 +34,24 @@ export const main = Reach.App(() => {
...
+      .assume(() => check(this == B))
+      .throwTimeout(relativeTime(deadline), null);
+    check(this == B);
+    k(null);
+    transfer(collateral).to(B);
+    transfer(loanAmt + interest).to(L);
+  } catch (_) {
+    // B timed out on paying back the loan
+    const [[], k] = call(LA.takeCollateral)
+      .assume(() => check(this == L));
```
jeapostrophe, line 50: You actually don't need this... which is part of the reason you didn't include it in the consensus step. Anyone can give Elle the collateral... but probably only she will.

examples/nft-pawn-shop/index.rsh
```diff
@@ -30,15 +34,24 @@ export const main = Reach.App(() => {
...
-    });
-  transfer(collateral).to(B);
-  transfer(loanAmt + interest).to(L);
+
+  try {
+    const [[], k] = call(LA.payBack)
+      .pay(() => loanAmt + interest)
+      .assume(() => check(this == B))
+      .throwTimeout(relativeTime(deadline), null);
+    check(this == B);
```
jeapostrophe, line 43: Maybe the same argument applies here... we should be okay with someone else paying B's loan :)

jeapostrophe: And if we do that, then we could do a weird thing and not use any APIs, but just have four different participants... kind of a weird, but maybe we discovered a new way to write Reach programs.

jeapostrophe: Even if we did want to enforce that it was B, you could make a new participant and set them to be B

jeapostrophe (changes requested):

examples/nft-pawn-shop/index.rsh
erer1243: It was an accidental exclusion. I have been wondering what is better practice, to be as lenient or strict as possible with triggering timeout actions. I just did it this way to stay on what seems intuitively like the safe side.

examples/nft-pawn-shop/index.rsh
erer1243: Would doing it that way not require someone to be attached the whole time still? I don't understand much about participant internals and what happens when you set them.

examples/nft-pawn-shop/index.rsh
jeapostrophe: I don't believe there is a strictness metric where you can decide to be more or less strict. There's just the right thing for the application... I think about it in terms of benefits. Like...

1. If X acts and then action B happens, which decides whether X "won" or not, then B must not be done by X, so we want to constrain it so someone else does it.
2. If action C spends resources of Y, then Y must be the only one that does it.
3. If action D returns resources to Z, then anyone can do it, but probably only Z will.

I think this program is like situation 3.

examples/nft-pawn-shop/index.rsh
jeapostrophe: First, to be clear about what I'm talking about... I'm suggesting that there are four participants:

1. Borrower
2. Lender
3. Repayer
4. Claimant

The program is...

1. Borrower deposits collateral
2. Lender gives loan to Borrower
3. Repayer repays loan and gives collateral and interest to Borrower
4. OR, Claimant waits until deadline and gives collateral to Lender

If Repayer and Claimant were respectively Borrower and Lender, then they'd have to run the program forever, because there's no way to "reattach". "To run the Borrower" would mean "launch the program, deposit collateral, wait, do something". That means that if the Borrower accidentally closed the window, they couldn't get back to the "do something" step.

The way I'm proposing... the Jon would run the program first as Borrower and then close it. Later, they would open it and run the program as Repayer (which would observe the previous actions of the Borrower and Lender) and finish it. Sara could do the same thing as Lender and then Claimant. OR, someone else could run the second parts if they wanted to pay for someone else's lunch.

erer1243: I like this alternative style for this program. it makes the code shorter and cleaner. Though, if there were more publishes to do, we would have to add an interact function and throws. Would it make sense to have a wrapper for that functionality (throwing and catching for intentional disconnects) in the js stdlib?

examples/nft-pawn-shop-participants/index.mjs
```diff
@@ -0,0 +1,68 @@
...
+
+  console.log("Starting balances:");
+  await printBalances();
+  const ctcBorrower = accBorrower.contract(backend);
+  const ctcLender = accLender.contract(backend, ctcBorrower.getInfo());
+
+  console.log(`Borrower puts up the ${nft.name} nft as collateral`);
+  try {
+    await Promise.all([
+      backend.Borrower(ctcBorrower, {
```
jeapostrophe, line 32: You should rewrite to use `ctcBorrower.p.Borrower`

examples/nft-pawn-shop-participants/index.mjs
```diff
@@ -0,0 +1,68 @@
...
+          console.log(`Borrower receives a loan of ${fmt(loanAmt)} ${stdlib.standardUnit}`);
+          await printBalances();
+          throw 'loanMade';
+        },
+        loanAmt,
+        interest,
+        deadline,
+        nft: nft.id,
+      }),
+      backend.Lender(ctcLender, {}),
```
jeapostrophe, line 43: Similar

examples/nft-pawn-shop-api/index.mjs
```diff
@@ -42,7 +42,7 @@ const run = async (borrowerPaysBack) => {
       backend.Lender(ctcLender, {}),
```
jeapostrophe, line 42: See my other comment about directly looking at `backend`

jeapostrophe (changes requested):

examples/nft-pawn-shop/index.rsh
jeapostrophe: I'm not sure how to do that wrapper in a way that is shorter than just writing out the normal JS pattern... maybe

```
withDisconnect(async (disconnect) => {
 ....
 ready: {
  disconnect();
 }
});
```

where `withDisconnect` is basically `try/catch` and `disconnect` is throw?

erer1243: That's pretty much what I was thinking. I would do it like
```js
stdlib.disconnect = () => throw "reach disconnect";
stdlib.withDisconnect = async (prom) => {
  try { await prom() }
  catch (e) { if (e !== "reach disconnect") throw e;
}

await stdlib.withDisconnect(backend.Alice(ctcAlice, { ready: stdlib.disconnect });
```

erer1243: Seems like a common pattern in testing. Not sure if it would be a common pattern in real usage.

erer1243:
Why switch from backend.Alice style to ctcAlice.p.Alice?

jeapostrophe (approved):

jeapostrophe:
> Why switch from backend.Alice style to ctcAlice.p.Alice?

It is the newer style and you won't make a mistake using backend1 when you meant to use backend2... because ctcAlice includes backend in it

examples/nft-pawn-shop/index.rsh
jeapostrophe: Make a Jira card for it, BUT... don't use that string... you should make an object and use object equality to get a ptr to it

PR #792, opened 2022-04-12
Core 1349 give more information about compilation status to developers

hamirmahal:
This screen recording provides some context for why we want this change, up until around 2:05, after which it goes into what this change actually does.

It also has some audio if you want to hear additional information. You can watch it at 2x speed if your browser supports it.

The goal of this change is to provide a lot of useful information regarding compilation in a way that respects users' attention.

https://user-images.githubusercontent.com/43425812/163056056-7e022e41-f1a5-4923-ad1d-89f233588c16.mp4

### DIFF
--- a/vsce/client/package-lock.json
+++ b/vsce/client/package-lock.json
@@ -12,17 +12,17 @@
 				"vscode-languageclient": "^7.0.0"
 			},
 			"devDependencies": {
-				"@types/vscode": "^1.52.0",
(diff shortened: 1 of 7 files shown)


jeapostrophe (approved):
This is a great change and I love that you make these easy to digest videos.

hamirmahal:
Thank you for the kind words!

PR #793, opened 2022-04-13
CORE-1380 document and expose getTimeSecs

erer1243:


js/stdlib/ts/ETH_like.ts
```diff
@@ -1176,6 +1176,7 @@ const ethLike = {
   setMinMillisBetweenRequests,
   setCustomHttpEventHandler,
   setSigningMonitor,
+  getTimeSecs,
```
jeapostrophe, line 1179: Don't you need to change `ETH` and `CFX` too? I'm pretty sure you do. You should write a test program that uses it (or stick it in an existing program)

jeapostrophe (changes requested):

js/stdlib/ts/ETH_like.ts
erer1243: ah whoops. trying to move too fast today

examples/getTimeSecs-demo/index.mjs
```diff
@@ -0,0 +1,19 @@
...
+const ctcAlice = accAlice.contract(backend);
+await ctcAlice.p.Alice({
+  giveTime: async (time) => {
+    console.log(`The contract was launched at network time: ${time}`);
+    console.log(`which is network seconds: ${await stdlib.getTimeSecs(time)}`);
+  }    
+});
+console.log(`The program started at network time ${startTime}`);
+console.log(`which is network seconds: ${await stdlib.getTimeSecs(startTime)}`);
+await stdlib.wait(100);
```
jeapostrophe, line 16: Eh... I am actually about to make a rule that we almost never call `wait` in tests, because it causes random failures with other tests. This is going to force the network forward 100 blocks which is going to mess up timeouts for other people. I think it is okay to just delete 16-19

jeapostrophe (changes requested):

examples/getTimeSecs-demo/index.mjs
erer1243: Got it.

jeapostrophe (approved):

PR #794, opened 2022-04-13
Update index.md

Craig-Elre:
Add a note that `index.mjs` should not be in the folder when trying to use `index.js` with `reach run`

### DIFF
--- a/docs/src/tut/rps/index.md
+++ b/docs/src/tut/rps/index.md
@@ -1529,7 +1529,9 @@ Furthermore, MetaMask does not support multiple active accounts, so if you want 
 To complete this section we'll use the `index.rsh` you've already written and create an `index.js` file from scratch which replaces `index.mjs`.
 
 :::note
-`index.mjs` needs to be moved or deleted from the folder in order for `index.js` to work as expected.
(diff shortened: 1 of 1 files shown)


docs/src/tut/rps/index.md
```diff
@@ -1528,6 +1528,10 @@ Furthermore, MetaMask does not support multiple active accounts, so if you want
 
 To complete this section we'll use the `index.rsh` you've already written and create an `index.js` file from scratch which replaces `index.mjs`.
 
+:::note
+`index.mjs` needs to be moved or deleted from the folder in order for `index.js` to work as expected.
```
jeapostrophe, line 1532: I want a little bit more...

If you do not replace `index.mjs`, then you will get an error when you run `reach react`. You can avoid this error by renaming the file to `index.mjs.bak` or by creating a totally new directory with a copy of the old `index.rsh`. We do the second option.

Then when we first tell them to run `reach react`, put in a note like this again... "Did it error? Do you still have `index.mjs`?"

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #795, opened 2022-04-13
Update the consensus page with a `while` example

Craig-Elre:
Add a code example of `while` to be explained later.

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -153,7 +153,7 @@ Read about finding [loop invariants](##guide-loop-invs) in the Reach guide.
 :::
 
 ```reach
-load: /examples/chickenrace/index.rsh
+load: /examples/chicken-race/index.rsh
 range: 43-64
 ```
 
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -148,6 +148,33 @@ if `{!rsh} COND_EXPR` is true, then the block executes,
 and if not, then the loop terminates and control transfers to the continuation of the while statement.
 The identifiers bound by `{!rsh} LHS` are bound within `{!rsh} DEFINE_BLOCK`, `{!rsh} INVARIANT_EXPR`, `{!rsh} COND_EXPR`, `{!rsh} BLOCK`, and the tail of the while statement.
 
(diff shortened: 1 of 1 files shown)


docs/src/rsh/consensus/index.md
```diff
@@ -148,6 +148,33 @@ if `{!rsh} COND_EXPR` is true, then the block executes,
 and if not, then the loop terminates and control transfers to the continuation of the while statement.
 The identifiers bound by `{!rsh} LHS` are bound within `{!rsh} DEFINE_BLOCK`, `{!rsh} INVARIANT_EXPR`, `{!rsh} COND_EXPR`, `{!rsh} BLOCK`, and the tail of the while statement.
 
+```reach
```
jeapostrophe, line 151: You should sample this from a file

docs/src/rsh/consensus/index.md
```diff
@@ -148,6 +148,33 @@ if `{!rsh} COND_EXPR` is true, then the block executes,
...
+        const [ da, db ] = isAlice ? [ 1, 0 ] : [ 0, 1 ];
+        each([Alice, Bob], () => {
+          interact.roundWinnerWas(isAlice); });
+        [ keepGoing, as, bs ] = [ true, as + da, bs + db ];
+        continue;
+      }
+```
+
 :::note
 Read about finding [loop invariants](##guide-loop-invs) in the Reach guide.
```
jeapostrophe, line 179: I also think it should go under this note

jeapostrophe (changes requested):

docs/src/rsh/consensus/index.md
Craig-Elre: In our DCT team meeting yesterday, we discussed how we could ensure that the coding samples did not get modified and change the line numbers that we referenced in the examples folder. So Njoku is creating a doc-samples folder to be inside of the examples folder with a document where we can put all of the samples we use with a section for each term in alphabetical order with room for 15 lines. However, the issue was brought up about what if the coding sample is longer than 15 lines. For that, it was decided that we would place the few code samples that are longer than 15 lines directly to the document. This code sample is around 20 lines, so I put it directly in the document.

jeapostrophe (approved):

PR #811, opened 2022-04-18
CORE-1418 fix tokenMetadata example races

erer1243:
Added events to each racey example contract to sync the indexer. As a bonus, threw events into the rpc server. Inferred from surrounding code how to do it; maybe it's not a good implementation. If so I will redo the rpc examples and remove that change. 

I feel the fact that the stdlib can fall out of sync with itself, like in `max-supply-precision`, would be surprising from a user's perspective. It bubbles up the implementation details to viable race conditions in algo frontends. Strikes me as bad form.

### DIFF
--- a/examples/max-supply-precision/index.mjs
+++ b/examples/max-supply-precision/index.mjs
@@ -1,4 +1,5 @@
 import { loadStdlib } from "@reach-sh/stdlib";
+import * as backend from './build/index.main.mjs';
 const stdlib = loadStdlib();
 
 const sbal = stdlib.parseCurrency(100);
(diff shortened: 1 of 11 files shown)


examples/max-supply-precision/index.mjs
```diff
@@ -14,8 +15,14 @@ const assertEq = (a, e) => {
   }
 };
 
+{ // sync indexer with latest blocks, so it can observe token metadata
+  const ctc = acc.contract(backend);
+  await ctc.p.Alice({});
+  await ctc.e.sync.next();
+}
```
jeapostrophe, line 22: This program could also work by adding a way to explicitly wait on the indexer. The current `wait` function just probes the node

jeapostrophe (approved):

PR #813, opened 2022-04-18
CORE-652: Allow APIs To Occur Multiple Times In A Program

chrisnevers:
Allow APIs to be called many times during a program. Each call must be in a different consensus step.

Eval/Core will keep a map of how many times an API is called. This map will be used to ensure each API call is in a different consensus step in EPP. Each API call will generate its own EPP program. The frontend and backend will emit a wrapper that jumps to the correct instance based on the state of the application.


hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3065,6 +3120,18 @@ cmeth sigi = \case
       cconcatbs_ (const $ return ()) $ zipWith f tys [1 ..]
       doWrap
       code "b" [handlerLabel which]
+  -- This API occurs in multiple places throughout the program.
+  -- Jump to the correct handler
+  CApi who sig _ _ _ _ _ wls -> do
+    block_' (bunpack who) $ do
+      comment $ LT.pack $ "API: " <> sig
+      comment $ LT.pack $ " ui: " <> show sigi
```
jeapostrophe, line 3128: You got rid of `sigDump`

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3065,6 +3120,18 @@ cmeth sigi = \case
...
+      forM_ wls $ \ (w, l) -> do
```
jeapostrophe, line 3129: You should use a BLT, look at the view code

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3197,13 +3265,17 @@ compile_algo env disp pl = do
             lr_at = ch_at
             lr_what = "Step " <> show i
         (_, C_Loop {}) -> Nothing
-  let a2lr (p, ApiInfo {..}) = LabelRec {..}
+  let a2lr qualify (p, ApiInfo {..}) = LabelRec {..}
         where
-          lr_lab = apiLabel p
+          (prefix, suffix) = bool ("", "") ("_", show ai_which) qualify
+          lr_lab = prefix <> apiLabel p <> LT.pack suffix
```
jeapostrophe, line 3271: You have this pattern three times I believe... you need to make it one function

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1357,15 +1361,22 @@ apiDef who ApiInfo {..} = do
...
           ty <- makeConsensusArg args
-          return $ (ty, argDefns, [], args)
-  (ty, argDefns, tyLifts, args) <- go ai_compile
+          return $ (ty, argDefns, [], args, m_arg_ty)
+  go ai_compile
+
+apiDef :: SLPart -> Bool -> ApiInfo -> App Doc
+apiDef who qualify ApiInfo {..} = do
+  let (prefix, suffix) = bool ("", "") ("_", show ai_which) qualify
+  let who_s = prefix <> bunpack who <> suffix
```
jeapostrophe, line 1377: Here's that pattern again... I think it's not something that is just in ALGO, but something that should be a shared DLBase resource

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1387,9 +1396,37 @@ apiDef who ApiInfo {..} = do
...
+  let defL = M.toList defs
+  defs' <- concat <$> mapM (\ (p, ms) -> do
+                              let qualify = M.size ms > 1
+                              ds <- mapM (\ (_, a) -> apiDef p qualify a) $ M.toAscList ms
+                              case qualify of
+                                True  -> do
+                                  d <- genApiJump p ms
+                                  return $ d : ds
+                                False -> return ds
+                            ) defL
```
jeapostrophe, line 1428: My taste is to define this as a `let go =` or do `defs' <- forM defL $ \(p, ms) -> do ....` and then `return $ vsep $ concat defs'`

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1495,12 +1532,26 @@ solEB args (DLinExportBlock _ mfargls (DLBlock _ _ t r)) = do
...
+createAPIRng :: ApiInfos -> App Doc
 createAPIRng env =
   case M.null env of
     True -> return ""
     False -> do
-      fields <- mapM (\(k, v) -> (pretty (bunpack k),) <$> solType_ v) $ M.toAscList env
+      fields <- concat <$> mapM (\(k, ms) -> do
+          let qualify = M.size ms > 1
+          let k' = bunpack k
+          fs <- mapM (\ (w, ai) -> do
```
jeapostrophe, line 1543: Good application of `forM`

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1495,12 +1532,26 @@ solEB args (DLinExportBlock _ mfargls (DLBlock _ _ t r)) = do
...
+                  let (prefix, suffix) = bool ("", "") ("_", show w) qualify
+                  let n = pretty $ prefix <> k' <> suffix
```
jeapostrophe, line 1545: And the pattern!

hs/src/Reach/EPP.hs
```diff
@@ -746,14 +778,29 @@ epp (LLProg at (LLOpts {..}) ps dli dex dvs das alias devts s) = do
...
-        let isAPI = S.member ee_who sps_apis
+  as <- readIORef be_api_steps
+  -- Ensure an API is called at most once in a given consensus step
+  forM_ (M.toAscList as) $ \ (k, v) -> do
+      forM_ (groupOn fst v) $ \ vs -> do
+        when (length vs /= 1) $ do
+          let apiAt = snd $ fromMaybe (impossible "api empty") $ headMay vs
+          expect_thrown apiAt $ Err_API_Twice k
+  -- When the same API call occurs in multiple steps,
+  -- make a seperate `EPProg` for each one.
```
jeapostrophe, line 789: separate

hs/src/Reach/EPP.hs
```diff
@@ -746,14 +778,29 @@ epp (LLProg at (LLOpts {..}) ps dli dex dvs das alias devts s) = do
...
+  let sps_ies' = M.foldrWithKey (\ k v acc ->
+          case M.lookup k as of
+            Just ns
+              | length ns > 1 -> foldr (\ (x,_) acc' -> M.insert (k, Just x) v acc') acc ns
+            _ -> M.insert (k, Nothing) v acc
+        ) mempty sps_ies
```
jeapostrophe, line 795: I really hate `()`s in Haskell code like this and like to define a helper function beforehand

hs/src/Reach/Eval/Core.hs
```diff
@@ -1896,6 +1898,14 @@ evalForm f args = do
       [de, e] -> return (de, e)
       _ -> illegal_args 2
 
+recordApiCall :: SLPart -> App ()
+recordApiCall who = do
+  apiCalls <- asks e_apiCalls
+  liftIO $ modifyIORef apiCalls $ \ env -> do
+    case M.lookup who env of
+      Just n  -> M.insert who (n + 1) env
+      Nothing -> M.insert who 1 env
```
jeapostrophe, line 1907: I think this is a good use `M.alter`

js/stdlib/ts/ALGO.ts
```diff
@@ -1499,6 +1499,15 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
       };
     };
 
+    const getCurrentStep_ = async (getC:GetC): Promise<BigNumber> => {
+      const { getAppState, getGlobalState } = await getC();
+      const appSt = await getAppState();
+      if ( !appSt ) { throw Error(`getCurrentStep_: no appSt`); }
+      const gs = await getGlobalState(appSt);
+      if ( !gs ) { throw Error(`getCurrentStep_: no gs`); }
+      return gs[0];
```
jeapostrophe, line 1508: I would look at what the view code does, because it already needs to learn this. At the very least you should make that code call this function

hs/src/Reach/APICut.hs
```diff
@@ -318,8 +317,8 @@ many_ = getFirst . mconcat . map First
 _many :: (a -> App (Maybe a)) -> [a] -> App (Maybe a)
 _many f l = many_ <$> mapM f l
 
-apc :: HasCounter a => a -> SLPart -> EPProg -> IO EPProg
-apc hc eWho = \case
+apc :: HasCounter a => a -> (SLPart, Maybe Int) -> EPProg -> IO EPProg
+apc hc (eWho, _) = \case
```
jeapostrophe, line 321: I need to be convinced more about the changes to this program... how do you cut the EPProg at the right part? Is it that EPP has "pre-cut" it for you? If so, then I think you could still leave in those `err` cases, but you'd turn them into `impossible`

hs/src/Reach/Backend/JS.hs
```diff
@@ -879,8 +879,44 @@ jsMapDefns varsHuh = do
...
+  , interactChk who
+  , "const ctc = ctcTop._initialize();"
+  , "const stdlib = ctc.stdlib;" ]
+
+jsApiWrapper :: B.ByteString -> [Int] -> App Doc
+jsApiWrapper p whichs = do
+  let who = pretty $ bunpack p
+  let chk_which w = "step == " <> pretty w
+  let chk_st = concatWith (\ l r -> l <> " || " <> r) $ map chk_which whichs
+  let assertStep = "stdlib.assert" <> parens (chk_st <> ", 'API called in the wrong state. Currently in state: ' + step ") <> semi
```
jeapostrophe, line 894: You could say which steps are allowed

hs/src/Reach/Backend/JS.hs
```diff
@@ -879,8 +879,44 @@ jsMapDefns varsHuh = do
...
+  let jmps = map (\ which -> do
+          let inst = "_" <> who <> pretty which
+          "if" <+> parens ("step" <+> "==" <+> pretty which) <+> braces ("return " <> inst <> parens "ctcTop, interact" <> semi)
+        ) whichs
+  let body = vsep $
+        setupPart who
+        <> [ "const step = await ctc.getCurrentStep()"
```
jeapostrophe, line 901: This function `getCurrentStep` should go into the `interfaces.ts` and you should increment the `backendVersion` in the `JS.hs` file

jeapostrophe: And make a note in the changelog

hs/src/Reach/EPP.hs
```diff
@@ -413,10 +427,18 @@ be_m = \case
...
   DL_Only at (Left who) l -> do
     ic <- be_inConsensus <$> ask
+    w <- asks be_which
     l'l <- ee_t l
+    which <- multipleApiCalls who >>= \case
+      -- Not an API or API only called in one place
+      False -> return $ Nothing
+      -- Take the `only(() => interact.in())` from preceding step
+      -- and the `only(() => interact.out())` from the correct step
+      True  -> return $ Just $ f $ w
```
jeapostrophe, line 437: The third `$` isn't needed

jeapostrophe (changes requested):

js/stdlib/ts/ETH_like.ts
```diff
@@ -765,7 +765,13 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account>
         return balanceOfNetworkAccount(networkAccount, mtok);
       }
 
-      return { getContractInfo, getContractAddress, getBalance, sendrecv, recv, getState, apiMapRef };
+      const getCurrentStep = async () => {
+        const ethersC = await getC();
+        const [ cs, _ ] = await ethersC["_reachCurrentState"]();
```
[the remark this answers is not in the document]
chrisnevers: `getState` ensures that the current step matches an expectation. I did abstract the uses of `ethersC["_reachCurrentState"]()` in to `getGlobalState()` to be similar to the ALGO runtime.

hs/src/Reach/APICut.hs
chrisnevers: Yes, EPP will essentially pre-cut the programs. The `EPProg` associated with `(who, step)` will have removed every other API call that is not in `step`. 

I don't think the err cases should be impossible. For example, if there are two paths to calling `interact.out` (from a conditional), `eSeenOut` can be true

hs/src/Reach/Backend/JS.hs
chrisnevers: I thought methods on a `Contract` get annotated in `shared_impl` and only top level things from the runtime are in `interfaces`?

hs/src/Reach/APICut.hs
jeapostrophe: First thing, that's cool.

Second thing, you're taking something that used to be an error path and making it okay. The error was there because something bad would happen. The pre-cut argument suggests that the bad thing I'm worried about is impossible. Thus, I suggest impossible. Now, you suggest that maybe we've enabled some new behavior that would not be bad. If that's the case, I want a test case that demonstrates that new feature. My taste is to turn it into an impossible, come up with a test case that triggers it, decide if we want to allow the program, remove the impossible, and ensure that the program does what we think it should

hs/src/Reach/Backend/JS.hs
jeapostrophe: https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Backend/JS.hs#L1073 is used to ensure that the compiled object's expectations match the things provided by the stdlib and we don't get weird undefined errors. It has to match here --- https://github.com/reach-sh/reach-lang/blob/master/js/stdlib/ts/ALGO.ts#L159

I pointed to the wrong file. You need to add it to `IContractCompiled` --- https://github.com/reach-sh/reach-lang/blob/master/js/stdlib/ts/shared_impl.ts#L179

hs/src/Reach/AST/DLBase.hs
```diff
@@ -1237,3 +1238,8 @@ arraysLength arrays = do
   case allEqual sizes of
     Right s -> s
     _ -> impossible "Inconsistent array sizes."
+
+
```
jeapostrophe, line 1242: Heresy!

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3138,13 +3136,9 @@ cmeth sigi = \case
...
-      comment $ LT.pack $ " ui: " <> show sigi
-      forM_ wls $ \ (w, l) -> do
-        gvLoad GV_currentStep
-        cint $ fromIntegral w
-        op "=="
-        code "bnz" [l]
-      op "err"
+      comment $ LT.pack $ sigDump sigi
+      let go _ _ = gvLoad GV_currentStep
+      cblt "api" go $ bltM $ M.fromList wls
```
jeapostrophe, line 3141: This isn't correct. `cblt` assumes the target is on the stack and it will call the function for the entry. I think you want
```
gvLoad GV_currentStep
let go _ l = code "b" [l]
cblt "api" go $ bltM $ M.fromList wls
```

jeapostrophe: If the code you wrote didn't crash CI, then you need better tests

jeapostrophe (changes requested):

hs/src/Reach/APICut.hs
chrisnevers: okay Ill look into that

hs/src/Reach/Connector/ALGO.hs
chrisnevers: wrote a better test that uses remote calls

hs/src/Reach/Backend/JS.hs
```diff
@@ -1093,7 +1129,7 @@ jsPIProg cr (PLProg _ _ dli dexports ssm (EPPs {..}) (CPProg _ vi _ devts _)) =
     local (\e -> e {ctxt_maps = dli_maps}) $
       jsViews vi
   mapsp <- jsMaps dli_maps
-  let partMap = flip M.mapWithKey epps_m $ \p _ -> pretty $ bunpack p
+  let partMap = M.foldrWithKey (\ (p, _) _ acc -> M.insert p (pretty $ bunpack p) acc) mempty epps_m
```
jeapostrophe, line 1132: The paren hate is why I flipped here :)

jeapostrophe: Also, this foldr is really an `M.mapWithKey`. I don't think you should ever do a fold when you mean map

hs/src/Reach/EPP.hs
```diff
@@ -428,6 +450,13 @@ be_m = \case
   where
     nop at = retb0 $ const $ return $ DL_Nop at
 
+multipleApiCalls :: SLPart -> BApp Bool
+multipleApiCalls who = do
+  apis <- asks be_api_calls
+  case M.lookup who apis of
+    Just n  -> return $ n > 1
+    Nothing -> return $ False
```
jeapostrophe, line 458: This function is copied in all of these fields but using different environment accessors and the previous comment would make it irrelevant

jeapostrophe (changes requested):

hs/src/Reach/Backend/JS.hs
chrisnevers: I'm using `fold` because I'm changing the keys

jeapostrophe: Yuck

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1374,7 +1374,6 @@ apiArgs ApiInfo {..} = do
 apiDef :: SLPart -> Bool -> ApiInfo -> App Doc
 apiDef who qualify ApiInfo {..} = do
   let who_s = adjustApiName (bunpack who) ai_which qualify -- ("", "") ("_", show ai_which) qualify
```
jeapostrophe, line 1376: What's this comment mean?

hs/src/Reach/EPP.hs
```diff
@@ -413,10 +427,18 @@ be_m = \case
            return $ DL_MapReduce at mri ans x z b a f')
   DL_Only at (Left who) l -> do
     ic <- be_inConsensus <$> ask
+    w <- asks be_which
     l'l <- ee_t l
+    which <- multipleApiCalls who >>= \case
+      -- Not an API or API only called in one place
+      False -> return $ Nothing
```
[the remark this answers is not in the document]
jeapostrophe: Boom

hs/t/n/API_OutBeforeIn.txt
```diff
@@ -3,11 +3,11 @@ Verifying for generic connector
   Verifying when ALL participants are honest
   Verifying when NO participants are honest
 Checked 4 theorems; No failures!
-reachc: error[RAPI0001]: API "Writer_f" calls interact.out() before interact.in()
+reachc: error[RAPI0000]: API "Writer_f" does not occur in program
```
jeapostrophe, line 6: This seems like a weird error? Is it because EPP didn't see the `in` and so it didn't separate it out?

jeapostrophe (approved):

PR #819, opened 2022-04-18
DCT-296

Craig-Elre:
Add docs-wisdom-2 and docs-wisdom-3 to the repo.


TheChronicMonster (changes requested):
CI failed on the `examples sink`
I believe it's because of the `/` at the end of `docs-wisdom-3/`

jeapostrophe:
The failure was because of me, something I'm about to fix

jeapostrophe (commented):
Is there supposed to be changes made to the files?

TheChronicMonster:
I think his only changes were to add the two new directories.

From Craig Re the PR:
"I added two new folders to the repo for WFS to see how Jay responds to
this creation method."

Craig-Elre:
> Is there supposed to be changes made to the files?

There are changes in these files. The two commits with "Update" in front is when I changed the `index.mjs` file to match the changes made in the tutorial.

examples/docs-wisdom-2/index.mjs
```diff
@@ -0,0 +1,32 @@
+import { loadStdlib } from '@reach-sh/stdlib';
+import * as backend from './build/index.main.mjs';
+
+if (process.argv.length < 3 || ['seller', 'buyer'].includes(process.argv[2]) == false) {
+  console.log('Usage: reach run index [seller|buyer]');
+  process.exit(0);
+}
```
jeapostrophe, line 7: Something you need to think about and do with these programs is that the CI won't actual do any tests if you build it this way. You should do something like `rps-8-interact` does to script the two different versions to interact and run automatic tests. @DanBurton Please create a Jira card for this for Oliver to help Craig with on the CORE board.

jeapostrophe (approved):

PR #820, opened 2022-04-18
CORE-1434 add compiler flag to quit after first verification fail

erer1243:
flag is `--first-fail-quit`. better name? should it also count knowledge verification fails?


jeapostrophe (changes requested):
You need to update the docs and the changelog

erer1243:
:+1: didn't realize we had cli flags documented

jeapostrophe (approved):

PR #821, opened 2022-04-18
CORE-1435 stdlib.withDisconnect

erer1243:
added `stdlib.withDisconnect` & `stdlib.disconnect`. Updated a few examples to use them instead of `try`. Did not write any docs - is this going to be an examples/tests only feature, or should I document it like it's public?

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -1136,3 +1136,27 @@ The default is no clawback address.
 
 For more information on Algorand-only options, see
 https://developer.algorand.org/docs/get-details/transactions/transactions/#asset-parameters.
+
+---
+@{ref("js", "withDisconnect")}@{ref("js", "disconnect")}
+```js
+withDisconnect<T>(f: () => Promise<T>) => Promise<T>
+disconnect(t: any) => void
+```
+`withDisconnect` calls the given function `f` such that any calls to `disconnect` within `f` will cause `withDisconnect` to return immediately.
+`withDisconnect` returns the value passed to `disconnect`.
+
+`disconnect` causes the surrounding call to `withDisconnect` to immediately return `t`.
+`disconnect` must be called from within a function passed to `withDisconnect`,
+otherwise an exception will be thrown.
+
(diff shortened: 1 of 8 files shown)


js/stdlib/ts/interfaces.ts
```diff
@@ -134,6 +134,8 @@ export interface Stdlib_User_Shared {
   numberToFixedPoint: (n: number) => FixedPoint
   parseInt: (n: {sign: boolean, i: num}) => number
   numberToInt: (n: number) => {sign: boolean, i: num}
+  withDisconnect: (p: Promise<any>) => Promise<void>
+  disconnect: () => undefined
```
jeapostrophe, line 138: You should make it take an argument that withDisconnect returns

js/stdlib/ts/interfaces.ts
```diff
@@ -134,6 +134,8 @@ export interface Stdlib_User_Shared {
   numberToFixedPoint: (n: number) => FixedPoint
   parseInt: (n: {sign: boolean, i: num}) => number
   numberToInt: (n: number) => {sign: boolean, i: num}
+  withDisconnect: (p: Promise<any>) => Promise<void>
```
jeapostrophe, line 137: I think it should take a lambda not a promise

jeapostrophe (changes requested):
It should be documented and exposed

docs/src/frontend/index.md
```diff
@@ -1136,3 +1136,27 @@ The default is no clawback address.
...
+`withDisconnect` and `disconnect` are intended as a utility to exit participant frontends early, like such:
+```js
+await stdlib.withDisconnect(() => ctcAlice.participants.Alice({
+  ready: () => {
+    console.log("Ready!");
+    stdlib.disconnect(null); // causes withDisconnect to immediately return null
+  }
+}));
+```
+Note that once a participant disconnects from the contract, they cannot rejoin as the same participant.
```
jeapostrophe, line 1162: Note, you should never start a sentence with note. It is a sign of bad writing. You should make a note ;) to fix this sentence, but I am merging now.

jeapostrophe (approved):

PR #833, opened 2022-04-20
DCT-320

EBEREGIT:
Error Message 123

Changed `Flower` to `Flower_girl` In the incorrect example

### DIFF
--- a/docs/src/rsh/errors/index.md
+++ b/docs/src/rsh/errors/index.md
@@ -2163,7 +2163,7 @@ For example, the code below erroneously uses the same name multiple times:
 
 ```reach
 const A = Participant('Flower_girl', {});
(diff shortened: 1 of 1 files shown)


TheChronicMonster (approved):

docs/src/rsh/errors/index.md
```diff
@@ -2163,7 +2163,7 @@ For example, the code below erroneously uses the same name multiple times:
 
 ```reach
 const A = Participant('Flower_girl', {});
-const B = API('Flower', { girl: Fun([UInt], Null) });
+const B = API('Flower_girl', { girl: Fun([UInt], Null) });
```
jeapostrophe, line 2166: This is not a correct change. The API scope is part of the names that get generated. Your change would make the function called `Flower_girl_girl`

PR #836, opened 2022-04-21
DCT-224

Craig-Elre:
Add a coding sample and a description to the `transfer` section.

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -372,6 +372,14 @@ The amount transfered must evaluate to less than or equal to the balance of the 
 
 A transfer expression may only occur within a consensus step.
 
+```reach
+load: /examples/simple-nft-auction/index.rsh
+range: 59-60
+```
+
+The first `transfer` example from the Simple NFT Auction pays the `lastPrice` to the NFT `Creator`.
+The second `transfer` example gives the sale quantity of the NFT to the `Participant` that bid the highest price.
+
 ### `require`
 
 @{ref("rsh", "require")}
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -377,8 +377,8 @@ load: /examples/simple-nft-auction/index.rsh
 range: 59-60
 ```
 
-The first `transfer` example from the Simple NFT Auction pays the `lastPrice` to the NFT `Creator`.
-The second `transfer` example gives the sale quantity of the NFT to the `Participant` that bid the highest price.
+The first `{!rsh}transfer` example from the Simple NFT Auction pays the `lastPrice` to the NFT `Creator`.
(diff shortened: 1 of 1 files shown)


docs/src/rsh/consensus/index.md
```diff
@@ -372,6 +372,14 @@ The amount transfered must evaluate to less than or equal to the balance of the
 
 A transfer expression may only occur within a consensus step.
 
+```reach
+load: /examples/simple-nft-auction/index.rsh
+range: 59-60
+```
+
+The first `transfer` example from the Simple NFT Auction pays the `lastPrice` to the NFT `Creator`.
+The second `transfer` example gives the sale quantity of the NFT to the `Participant` that bid the highest price.
```
TheChronicMonster, line 381: `transfer` and `Participant` need the `{!rsh}` prefix.

docs/src/rsh/consensus/index.md
```diff
@@ -372,6 +372,14 @@ The amount transfered must evaluate to less than or equal to the balance of the
 
 A transfer expression may only occur within a consensus step.
 
+```reach
+load: /examples/simple-nft-auction/index.rsh
+range: 59-60
+```
+
+The first `transfer` example from the Simple NFT Auction pays the `lastPrice` to the NFT `Creator`.
```
TheChronicMonster, line 380: `transfer` needs `{!rsh}` prefix.

TheChronicMonster (changes requested):

docs/src/rsh/consensus/index.md
```diff
@@ -370,15 +370,15 @@ where `{!rsh} PAY_AMOUNT_EXPR` is an expression that evaluates to a [pay amount]
 performs a transfer of network tokens or non-network tokens from the contract to the named participant.
 The amount transfered must evaluate to less than or equal to the balance of the network and non-network tokens in the contract account.
 
-A transfer expression may only occur within a consensus step.
+A `{!rsh} transfer` expression may only occur within a consensus step.
```
TheChronicMonster, line 373: I don't think `transfer expression` should be monospaced, because the first mention of it is in standard format. (See paragraph above this one)
IDK why. Perhaps @jeapostrophe could educate us on the nuance of not monospacing "transfer expression"?

TheChronicMonster (changes requested):

TheChronicMonster (approved):

jeapostrophe (approved):

docs/src/rsh/consensus/index.md
jeapostrophe: The subject of this sentence is the difference between using a word in its English meaning and talking about a particular word, like "subject", itself.

There is a English meaning to "transfer" and when we're writing a sentence like "Alice did a transfer of funds to the contract", we're not talking about the Reach function `transfer`, but we are talking about that when we say, "This block of code shows an example of using `transfer` to transfer funds."

PR #837, opened 2022-04-21
improve error handling when updating via extension

hamirmahal:
# Without this change

If something goes wrong during the

```ts
execSync(
  'docker pull "reachsh/reach-cli:latest" ' +
  `&& curl https://docs.reach.sh/reach -o "${
    pathToScript
  }" && chmod +x "${pathToScript}"`
);
```

step, the user interface just hangs.

https://user-images.githubusercontent.com/43425812/164503220-39f2ad99-9ece-4991-9ef2-81d5767523ba.mp4

# With this change

The user interface should update with an error notification if an error occurs, instead of just "hanging". I also split that big

```ts
execSync(
  'docker pull "reachsh/reach-cli:latest" ' +
  `&& curl https://docs.reach.sh/reach -o "${
    pathToScript
  }" && chmod +x "${pathToScript}"`
);
```

command into its three pieces for a more fine-grained approach, to hopefully help us troubleshoot things.

Both screen recordings should have audio, so feel free to unmute them for additional information.

https://user-images.githubusercontent.com/43425812/164503236-1f4a8e95-b23a-47fa-8dd2-444bd5de21f5.mp4

### DIFF
--- a/vsce/client/src/util/update.ts
+++ b/vsce/client/src/util/update.ts
@@ -96,14 +96,35 @@ const ASYNC_UPDATE_USING = (
   });
 };
 
+/**
+ * @param pathToScript the path to download the script to
+ * @returns an error if something went wrong
+ */
 const UPDATE_SCRIPT_AND_CLI_IMAGE_USING = (
   pathToScript: string
-): Buffer => execSync(
-  'docker pull "reachsh/reach-cli:latest" ' +
-  `&& curl https://docs.reach.sh/reach -o "${
-    pathToScript
-  }" && chmod +x "${pathToScript}"`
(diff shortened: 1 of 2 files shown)


mattaudesse (approved):
I think this is an improvement but could be made even better by not forcing `execSync` on what feels like a more natural use case for JS' default asynchronicity (which would probably require less code, too).

There's also some unnecessary repetition in your `try`/`catches` either way.

mattaudesse (approved):
IMHO it's still a lot more characters than necessary to achieve what it's doing but certainly a step up for smoothness in the UI.

PR #845, opened 2022-04-22
DCT-317 done

EBEREGIT:
### Error Message 119

Added Code Samples

### DIFF
--- a/docs/src/rsh/errors/index.md
+++ b/docs/src/rsh/errors/index.md
@@ -2116,6 +2116,22 @@ Maybe(UInt).Some(5).match({
 This error indicates that you have inspected the details about a publication, such as via `{!rsh} didPublish()`, before there has been a publication.
 This is impossible, so the expression must be moved after the first publication.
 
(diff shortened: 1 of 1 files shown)


docs/src/rsh/errors/index.md
```diff
@@ -2116,6 +2116,22 @@ Maybe(UInt).Some(5).match({
 This error indicates that you have inspected the details about a publication, such as via `{!rsh} didPublish()`, before there has been a publication.
 This is impossible, so the expression must be moved after the first publication.
 
+For example, the code below erroneously puts `{!rsh} didPublish()` before `{!rsh} publish`:
+
+```reach
+  Alice.didPublish(handAlice);
```
jeapostrophe, line 2122: This program is broken for a different reason... you can only call `didPublish` inside of an `only` block. The real way to get this error is to call it before `publish` but after `init`

jeapostrophe (changes requested):

PR #850, opened 2022-04-25
CORE-1408

jeapostrophe:

PR #851, opened 2022-04-25
CORE-1487: teach `reach run` a new `--entrypoint` option

mattaudesse:
https://github.com/reach-sh/reach-lang/pull/850/files#r857859713

### DIFF
--- a/hs/app/reach/Main.hs
+++ b/hs/app/reach/Main.hs
@@ -1187,11 +1187,12 @@ run' :: Subcommand
 run' = command "run" . info f $ d <> noIntersperse
   where
     d = progDesc "Run a simple app"
-    f =
-      go <$> switchIsolate
+    f = go
+        <$> switchIsolate
+        <*> strOption (long "entrypoint" <> value "main" <> help "Override default \"main\" entrypoint")
         <*> argAppOrDir
         <*> manyArgs "APP"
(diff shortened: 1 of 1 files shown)


jeapostrophe (commented):
This could work. Could we do even better... could we teach `reach compile` to do nothing if the build file is newer? Could we have `reach compile` tell `reach run` what files it will make? We don't have to... I'm asking us to think about it...

In the specific program I'm working with, it has 4 different output files.

mattaudesse:
> This could work. Could we do even better... could we teach `reach compile` to do nothing if the build file is newer? Could we have `reach compile` tell `reach run` what files it will make? We don't have to... I'm asking us to think about it...
> 
> In the specific program I'm working with, it has 4 different output files.

I definitely think it could be improved, yeah (and even has to be for CORE-168 + vaguely related stuff we know we want to work better).

Right now all this really gives us is the "build file is newer" part for `reach run`.
https://github.com/reach-sh/reach-lang/blob/9985882165f79f4fefc599245c262bce5d88e80e/hs/app/reach/Main.hs#L1245-L1252

Wanna wait for a better, more generalized solution?

jeapostrophe:
I think we should wait... I can't imagine someone really meticulously setting `--entrypoint`. Thanks for jumping right on my comment :)

PR #852, opened 2022-04-25
DCT 317

EBEREGIT:
### Error Message 119

Added Code Samples

### DIFF
--- a/docs/src/rsh/errors/index.md
+++ b/docs/src/rsh/errors/index.md
@@ -2120,14 +2120,24 @@ For example, the code below erroneously puts `{!rsh} didPublish()` before `{!rsh
 
 ``` reach
 load: /hs/t/n/Err_NotAfterFirst.rsh
+range: 6 - 10
 ```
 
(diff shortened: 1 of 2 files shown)


docs/src/rsh/errors/index.md
```diff
@@ -2116,6 +2116,18 @@ Maybe(UInt).Some(5).match({
...
+For example, the code below erroneously puts `{!rsh} didPublish()` before `{!rsh} publish`:
+
+``` reach
+load: /hs/t/n/Err_NotAfterFirst.rsh
+```
+
+This error can be corrected by placing `{!rsh} publish` before `{!rsh} didPublish()`:
+
+```reach
+load: /hs/t/n/j610.rsh
```
jeapostrophe, line 2128: This program is also incorrect though, so I don't think you should use this example. (All of the programs in `n` don't work and the ones in `y` do)

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #854, opened 2022-04-26
CORE-1379 *Prog to records

erer1243:
Committed just position -> record syntax changes. Want to run these names by you before I finish changing all the references to them. Otherwise I might have to repeat a bunch of work.. :P

### DIFF
--- a/hs/src/Reach/AST/PL.hs
+++ b/hs/src/Reach/AST/PL.hs
@@ -137,8 +137,8 @@ data EPProg = EPProg
   deriving (Eq)
 
 instance Pretty EPProg where
-  pretty (EPProg _ _ ie et) =
-    pretty ie <> semi <> hardline <> pretty et
+  pretty (EPProg _ _ epp_interactEnv epp_tail) =
+    pretty epp_interactEnv <> semi <> hardline <> pretty epp_tail
 
 data CTail
   = CT_Com DLStmt CTail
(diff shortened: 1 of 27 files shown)


hs/src/Reach/AST/PL.hs
```diff
@@ -128,8 +128,12 @@ instance Pretty ETail where
...
       cm l = parens (hsep $ punctuate comma $ l)
 
-data EPProg
-  = EPProg SrcLoc Bool InteractEnv ETail
+data EPProg = EPProg
+  { epp_at :: SrcLoc
+  , epp_isApi :: Bool
+  , epp_interactEnv :: InteractEnv
+  , epp_tail :: ETail
+  }
```
jeapostrophe, line 136: Don't forget `CPProg` in this file

jeapostrophe (commented):
The names are good. When you are updating references, use this rule...

1. If the names are bound in a block and only some are looked at, then use e.g. `EPProg {..}`
2. If the names are bound in a block and we expect all to be looked at (like in the optimizer, pretty printer, or type collector), then use `EPProg at isApi interactEnv tail` (explicitly write out the names and use the same names as the fields... but don't go crazy if this requires you to rename everything. The point of using positional arguments is to ensure that if we add a field in the future, it isn't forgotten about in something that is supposed to look at everything
3. If you are constructing one without previously binding it, use `EPProg {..}` and name the fields with `let epp_at = whatever`. This is convenient.
4. If you are constructing one WITH having bound it previously, then if you did 2 then use positions, but if you did 1 then follow 3.

jeapostrophe (approved):

Issue #856, opened 2022-04-26
[Suggestion] expose signTxns on window.algorand (or some workaround to access it)


grzracz:
It's needed to access TX signature without actually sending it to chain

DanBurton:
I believe the workaround you're looking for is something like:

```
const w = reach.walletFallback({
  providerEnv: 'TestNet',
  MyAlgoConnect,
});
w.signTxns(txns);
```

jeapostrophe:
We believe that does not exist because our fallback only implements signAndPostTxns but could easily implement that... or am I wrong?

DanBurton:
Ohhhh I see. I forgot we don't currently expose that, but that optional interface is in the ARC and is super easy for us to expose. Opened a PR.

PR #868, opened 2022-04-27
CORE-1504 - core-1504-signTxns

DanBurton:
Addresses #856 by exposing the optional methods `signTxns` and `postTxns` as specified by the wallet ARCs.


jeapostrophe (approved):
Approved given CI success

PR #882, opened 2022-05-01
correct syntax error for reach command

kcelestinomaria:
``./reach rpc-server`` instead of ``reach rpc-server``


jeapostrophe:
This is not a correct change. The command is `reach rpc-server`. It is typical to document commands this way, although you can only run it that way if you `reach` in your `PATH`. If you don't have it in your `PATH`, then you would run the executable directly by specifying the path as in `/a/b/c/reach rpc-server`. `./reach` is just what you do when it is in the same directory.

PR #887, opened 2022-05-02
DCT 205

EBEREGIT:
### `init`

- Added Code Sample 
- Added a note saying that it's preferable to use `API` over `ParticipantClass`


docs/src/rsh/appinit/index.md
```diff
@@ -25,6 +25,18 @@ changed because that name was misleading.
...
+In the example below, see how `{!rsh} init();` is used to finalize the available `{!rsh} Participant` and `{!rsh} ParticipantClass`. 
+After which a local step is introduced:
+
+```reach
+load: ./examples/nft-auction/index.rsh
+range: 23 - 30
+```
+
+:::note
+Since `{!rsh} ParticipantClass` is being deprecated, it is preferable to use `{!rsh} API`.
```
jeapostrophe, line 37: I think it would be better to just not show an example that uses a participant class

jeapostrophe (changes requested):

TheChronicMonster:
@Njoku Samson Ebere ***@***.***> we can use the example
API-Twice using the range 4-15
https://github.com/reach-sh/reach-lang/blob/master/examples/api-twice/index.rsh

PR #896, opened 2022-05-03
improve documentation tooltips

hamirmahal:
We don't want readers to think tooltips are real links, and then click on them, and lose their place on the page.

Actually preventing clicking on the tooltips causing the page to jump to the top is something we will have to address later; straightforward fixes like doing `event.preventDefault()`, `event.stopPropagation()`, and overriding tooltips' `onclick` handlers with `return false;` don't appear to work.

This change just changes the tooltip color, hopefully discouraging users from clicking on them and losing their place on the page.

### Without this change

https://user-images.githubusercontent.com/43425812/166551666-6613b8f6-f949-45a1-a539-b60bb144cd0c.mp4

### With this change

https://user-images.githubusercontent.com/43425812/166538160-c5168571-d6b7-4f82-bada-be1897a7e5a2.mp4

### DIFF
--- a/docs/dev/generator.js
+++ b/docs/dev/generator.js
@@ -420,7 +420,7 @@ const processMd = async ({baseConfig, relDir, in_folder, iPath, oPath}) => {
   };
 
   const tooltip = (text, hoverText) => {
-    return `<a class="ui-tooltip" title="${hoverText}"><span style="cursor">${text}</span></a>`;
+    return `<a class="ui-tooltip" title="${hoverText}"><span style="cursor: help">${text}</span></a>`;
   };
 
   const makeDirectiveWithClass = (classString) => {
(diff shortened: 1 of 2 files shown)


Craig-Elre (commented):
I really like the change to the color and hover color. Thanks for your work on this!

hamirmahal:
You're welcome! I hope Jay likes this as much as you do.

TheChronicMonster:
This is a good first step. Perhaps there is logic in the generator that causes tooltips to act like hyperlinks?

jeapostrophe:
They "act like tooltips" because they are `a` tags. I think you should add a tooltip icon with `::after` and `content:` in the CSS style.

hamirmahal:
This screen recording has audio.

https://user-images.githubusercontent.com/43425812/166725560-f981c3cc-e6c6-42ce-9544-034d5bd9aa67.mp4

jeapostrophe (approved):

PR #904, opened 2022-05-04
CORE-1320: Shared Local/Consensus Checks

chrisnevers:
Introduce fork/parallelReduce `.api_` branch which allows local checks, pay, and consensus expressions to be expressed in one consolidated function. I didn't make an equivalent for case because I think it would turn out ugly; e.g specifying `{ msg, when }` as an element of the return value. But, I can add it if needed

Add `.check(() => {})` to `publish`.

Add `.check((...dom) => {})` to api calls.

Add a check argument to `.case` in fork/parallelReduce.


vsce/snippets.json
```diff
@@ -73,6 +73,13 @@
 		],
 		"description": "while"
 	},
+	"API": {
+		"prefix": "Api",
+		"body": [
+			"API('${1:NAME}', { ${0} })"
+		],
+		"description": "Api constructor"
```
jeapostrophe, line 81: API

hs/src/Reach/Eval/Core.hs
```diff
@@ -1760,7 +1773,11 @@ evalForm f args = do
         go p' = retV $ public $ SLV_Form $ SLForm_parallel_reduce_partial $ p' {slpr_mode = Nothing}
         makeTimeoutArgs mode aa = Just (mode, fst aa, snd aa)
         retTimeout prm aa = go $ p {slpr_mtime = makeTimeoutArgs prm aa}
-    SLForm_Part_ToConsensus p@(ToConsensusRec {..}) ->
+    SLForm_Part_ToConsensus p@(ToConsensusRec {..}) -> do
+      let mustBeThunk = \case
```
jeapostrophe, line 1777: Maybe use `deconstructFun`? And then check that formals are empty?

hs/src/Reach/Eval/Core.hs
```diff
@@ -2432,6 +2456,15 @@ doBalanceUpdate mtok op = \case
     let up_rator = SLV_Prim $ SLPrim_PrimDelay at (SLPrim_op op) [] [(Public, rhs)]
     -- Find the current balance of mtok
     bsv <- getBalanceOf mtok
+    -- Assume we can pay into contract
+    when (op == S_ADD) $ do
```
jeapostrophe, line 2460: We can also assume that subtraction works, because we know we're already going to check that the balance is sufficient for the transfer

hs/src/Reach/Eval/Core.hs
```diff
@@ -4939,11 +4983,23 @@ doApiCall lhs (ApiCallRec {..}) = do
...
+  let ss = [callOnly [onlyThunk], es pub5, assignLhs, es setDetails]
   return ss
 
-doForkAPI2Case :: [JSExpression] -> App [JSExpression]
-doForkAPI2Case args = do
+prependFunStmts :: [JSStatement] -> JSExpression -> App JSExpression
+prependFunStmts stmts f = do
+  (fargs, fs) <- deconstructFunStmts f
+  return $ jsArrowStmts a fargs $ stmts <> fs
+  where a = jsa f
```
jeapostrophe, line 4993: I like `let`

hs/src/Reach/Eval/Core.hs
```diff
@@ -4939,11 +4983,23 @@ doApiCall lhs (ApiCallRec {..}) = do
...
+
+prependFunArgs :: [JSExpression] -> JSExpression -> App JSExpression
+prependFunArgs xtra f = do
+  (args, stmts) <- deconstructFunStmts f
+  return $ jsArrowStmts a (xtra <> args) stmts
+  where a = jsa f
```
jeapostrophe, line 4999: I think that these should be the same function or at least the same impl and you just pass `mempty` to one of the options

hs/src/Reach/Eval/Core.hs
```diff
@@ -4955,55 +5011,91 @@ doForkAPI2Case args = do
...
-  let mkx xp = jsThunkStmts xpa [jsConst xpa dom (readJsExpr "declassify(interact.in())"), e2s (jsCall xpa xp [dotdom]), JSReturn xpa (Just $ jsObjectLiteral xpa $ M.fromList [("msg", dom)]) sp]
+  let jsInlineCall f fargs =
+        case f of
+          JSExpressionParen _ e _ -> jsInlineCall e fargs
+          JSArrowExpression aargs a' s -> do
+            let pargs = parseJSArrowFormals at aargs
+            let JSBlock _ ss _ = jsArrowBodyToRetBlock s
+            let ss' = [jsConst a' (jsArrayLiteral a' pargs) (jsArrayLiteral a fargs)]
+            return $ ss' <> ss
+          _ -> expect_ $ Err_Fork_ConsensusBadArrow f
```
jeapostrophe, line 5022: Maybe another use of `deconstructFun`?

hs/src/Reach/Eval/Core.hs
```diff
@@ -5105,9 +5197,24 @@ doFork ks (ForkRec {..}) = locAt slf_at $ do
           _ -> expect_ $ Err_Fork_ConsensusBadArrow after_e
           where
             afterAt = srcloc_jsa "consensus block" (jsa after_e)
+  let augWithChecks chk e = do
+        (_, chk_ss) <- deconstructFunStmts chk
+        (e_args, e_ss) <- deconstructFunStmts e
+        return $ jsArrowStmts (jsa e) e_args $ chk_ss <> e_ss
```
jeapostrophe, line 5203: This is that prepend stmts function

jeapostrophe (changes requested):

hs/t/y/many_itxns.txt
```diff
@@ -4,4 +4,5 @@ Verifying for generic connector
   Verifying when NO participants are honest
 Checked 517 theorems; No failures!
 WARNING: Compiler instructed to emit for Algorand, but we can statically determine that this program will not work on Algorand, because:
- * Step 0 uses 261 inner transactions, but the limit is 256. Step 0 starts at ./many_itxns.rsh:13:5:dot.
+ * Step 0 uses 264 inner transactions, but the limit is 256. Step 0 starts at ./many_itxns.rsh:13:5:dot.
+ * Too many scratch slots
```
jeapostrophe, line 8: Why do you think this program changed?

jeapostrophe (approved):

hs/t/y/many_itxns.txt
chrisnevers: The program makes 1 pay and 256 transfers. So there are 256 assumptions, that balance doesn't overflow/wraparound, added via `doBalanceUpdate`

jeapostrophe: I get why there are too many scratch slots. I mean, I don't know where the 3 extra inner txns came from. Maybe from cost?

chrisnevers: I think so, I don't quite understand the TEAL code inside of the publish handler, but it seems like it's making three more itxns to the companion contract

jeapostrophe (commented):
This looks good sans the `remote-rsh` error. I'm not sure what's going on there.

jeapostrophe:
I'm approving preemptively if you can see that the remote-rsh problem is a fluke and not caused by you.

hs/src/Reach/Optimize.hs
```diff
@@ -393,13 +394,15 @@ instance Optimize DLExpr where
     DLE_Claim at fs t a m -> do
       a' <- opt a
       let meh = return $ DLE_Claim at fs t a' m
-      case (t, a') of
-        (CT_Possible, _) -> meh
-        (_, DLA_Literal (DLL_Bool True)) -> do
+      isSim <- asks eSimulate
```
jeapostrophe, line 397: You're going to have to consider some operations to be impure under simulation, but I believe that is going to be part of ac, not opt, so that's pretty groddy.

jeapostrophe (approved):

PR #906, opened 2022-05-04
DCT-246

Craig-Elre:
Add the `getAddress` code sample and description.

### DIFF
--- a/docs/src/rsh/compute/index.md
+++ b/docs/src/rsh/compute/index.md
@@ -1701,6 +1701,15 @@ getAddress()
 The @{defn("getAddress")} primitive returns the `{!rsh} Address` value of the deployed contract's account.
 This function may not be called until after the first publication (which creates the contract).
 
+```reach
+load: /examples/ctc-address/index.rsh
+range: 15-22
+```
+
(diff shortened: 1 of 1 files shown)


jeapostrophe (approved):

docs/src/rsh/compute/index.md
```diff
@@ -1701,6 +1701,15 @@ getAddress()
 The @{defn("getAddress")} primitive returns the `{!rsh} Address` value of the deployed contract's account.
 This function may not be called until after the first publication (which creates the contract).
 
+```reach
+load: /examples/ctc-address/index.rsh
+range: 15-22
+```
+
+In this code sample, Alice publishes to create the contract, and then an object named `info` is created to reference `{!rsh} getContract` that pulls in the contract information.
```
TheChronicMonster, line 1709: In the sample above, Alice publishes to `{!rsh} deploy` the contract, 
and then creates an object `info` to reference the `{!rsh} getContract` primitive, which returns the `{!rsh} Contract` value. 
A second object, `addr` references `{!rsh} getAddress`, which returns Alice's account address. 
The contract moves into Alice's local step and both `info` and `addr` are called in `{!rsh} interact` expressions.

PR #909, opened 2022-05-05
CORE-1549: Allow Multiple Defines in Parallel Reduce

chrisnevers:


hs/src/Reach/Eval/Core.hs
```diff
@@ -1763,7 +1763,7 @@ evalForm f args = do
           go $ p {slpr_mpay = Just x}
         Just PRM_Def -> do
           x <- one_arg
-          go $ p {slpr_mdef = Just x}
+          go $ p {slpr_defs = x : slpr_defs}
```
jeapostrophe, line 1766: This puts it at the front, but I think it should put it at the end, so that they occur in the output in the order of the program. You should have an example where the second refers to something bound in the first and you'll see that it won't work

hs/t/n/Err_ParallelReduce_DefineBlock.txt
```diff
@@ -1,8 +1,8 @@
-reachc: error[RE0113]: `define` expects an argument of the form: `() => DEFINE_BLOCK`, where `DEFINE_BLOCK` is a statement block.
+reachc: error[RE0035]: Invalid function application. Cannot apply: UInt
```
jeapostrophe, line 1: This error is worse. The location is bad

hs/src/Reach/Eval/Core.hs
```diff
@@ -5566,11 +5566,9 @@ doParallelReduce lhs (ParallelReduceRec {..}) = locAt slpr_at $ do
   let commit_s = JSMethodCall (jid "commit") a JSLNil a sp
   let while_body = [commit_s, fork_s]
   let while_s = JSWhile a a while_e a $ JSStatementBlock a while_body a sp
-  block_s <- case pr_mdef of
-    Just (JSArrowExpression _ _ b) -> return $ jsArrowBodyToStmt b
-    Just ow -> locAtf (srcloc_jsa "define" $ jsa ow) $ expect_ Err_ParallelReduce_DefineBlock
-    Nothing -> return $ JSStatementBlock a [] a sp
-  let pr_ss = [var_s, block_s, inv_s, while_s]
+  block_ss <- flip concatMapM pr_defs $ fmap snd . deconstructFunStmts
```
jeapostrophe, line 5569: The root of that error is that `deconstructFunStmts` evals the argument when it is not an arrow/etc

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #930, opened 2022-05-07
Core 1114 + CORE 1235

phoniks:
Graph shows action taken on hover

https://user-images.githubusercontent.com/7213762/167275844-7e827f58-4a4d-47a2-9faf-32f0e3c6aff0.mp4

### DIFF
--- a/examples/simulator_ui/src/App.tsx
+++ b/examples/simulator_ui/src/App.tsx
@@ -1,5 +1,5 @@
 import * as c from "@reach-sh/simulator-client";
-import { useEffect, useState, useReducer } from "react";
+import { useEffect, useState, useReducer, createRef } from "react";
 import { createGlobalStyle } from "styled-components";
 import getObjectView from "./reach/ObjectView";
 import LeftPanel from "./componenents/organisms/LeftPanel";
@@ -11,6 +11,7 @@ import styled from "styled-components";
 import VisualizerPanel from "./componenents/organisms/Visualizer";
 import { IUserNode, IUserEdge } from "@antv/graphin";
 import { Participant } from "./types";
+import Graphin from '@antv/graphin'
 
 export const GlobalStyles = createGlobalStyle`
   html {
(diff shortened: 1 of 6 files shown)


chikeabuah (changes requested):
the workflow here seems a bit off to me. the task is the show the action taken on hover, but i don't see a receive action happening in the recording. if the action shown is just a placeholder (i.e. it would always show "Receive" no matter what) then I don't feel comfortable passing on this. the receive actions should happen in the response UI. if that isn't complete yet, then your tasks are being completed in the wrong order.

this ui needs to match the programmer art ui functionally before new features are possible imho.

chikeabuah:
technically, what you're hovering there is the "Init" step, which isn't an "Action" in Sim Lang, but the graph should show "Init" there.
regardless, the UI needs to be able to do "Action"s before the graph can visualize them

phoniks:
> the workflow here seems a bit off to me. the task is to show the action taken on hover, but i don't see a receive action happening in the recording. if the action shown is just a placeholder (i.e. it would always show "Receive" no matter what) then I don't feel comfortable passing on this. the receive actions should happen in the response UI. if that isn't complete yet, then your tasks are being completed in the wrong order.
> 
> this ui needs to match the programmer art ui functionally before new features are possible imho.

Nah, that Receive is dynamic and comes from the client ( in other words, I've never actually typed "Receive" in my code).  I did feel like it was a weird way to label that action tbh, but since there was a wager in the init, I assumed it meant that the contract received the wager.  My question, extending from your second comment, is does the client ever tell me "this is an init step" or is that just something that I need to implement in the UI based on knowing that we just hit the init button?

chikeabuah:
> > the workflow here seems a bit off to me. the task is to show the action taken on hover, but i don't see a receive action happening in the recording. if the action shown is just a placeholder (i.e. it would always show "Receive" no matter what) then I don't feel comfortable passing on this. the receive actions should happen in the response UI. if that isn't complete yet, then your tasks are being completed in the wrong order.
> > this ui needs to match the programmer art ui functionally before new features are possible imho.
> 
> Nah, that Receive is dynamic and comes from the client ( in other words, I've never actually typed "Receive" in my code). I did feel like it was a weird way to label that action tbh, but since there was a wager in the init, I assumed it meant that the contract received the wager. My question, extending from your second comment, is does the client ever tell me "this is an init step" or is that just something that I need to implement in the UI based on knowing that we just hit the init button?

the latter

PR #931, opened 2022-05-09
DCT-227

EBEREGIT:
### Token minting

Added code samples


docs/xrefs.json
```diff
@@ -19,6 +19,58 @@
     "title": "Guarantees",
```
jeapostrophe, line 19: Don't commit this file

hamirmahal: Don't we want it to always be up to date?

hamirmahal: Ah, I see. #935 should take care of that.

PR #940, opened 2022-05-09
DCT-228

EBEREGIT:
### Mappings: creation and modification

Added code samples

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -599,17 +599,17 @@ If `{!rsh} KEY_TYPE_EXPR` is not specified, it will default to `{!rsh} Address`.
 
 For example the code below contains only the `{!rsh} VAL_TYPE_EXPR`:
 ```reach
-load: examples/splice1/index.rsh
+load: /examples/splice1/index.rsh
 range: 7 - 7
 ```
 
 While the following code contains both the `{!rsh} KEY_TYPE_EXPR` and `{!rsh} VAL_TYPE_EXPR`:
 ```reach
-load: examples/map-arbitrary-key/index.rsh
+load: /examples/map-arbitrary-key/index.rsh
 range: 24 - 24
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/rsh/consensus/index.md
```diff
@@ -597,6 +597,18 @@ delete bidsM[this];
...
+range: 7 - 7
+```
+
+While the following code contains both the `{!rsh} KEY_TYPE_EXPR` and `{!rsh} VAL_TYPE_EXPR`:
+```reach
+load: examples/map-arbitrary-key/index.rsh
+range: 24 - 24
+```
+
 This returns a value which may be used to dereference particular mappings via `{!rsh} map[EXPR]`.
```
jeapostrophe, line 612: I think the interjection of the examples makes the use of "this" in this sentence confusing and you should revise it

jeapostrophe (changes requested):

docs/src/rsh/consensus/index.md
TheChronicMonster: Also, `examples/` directory needs to be prefixed with a `/`

docs/src/rsh/consensus/index.md
```diff
@@ -597,7 +597,19 @@ delete bidsM[this];
...
+range: 7 - 7
+```
+
+While the following code contains both the `{!rsh} KEY_TYPE_EXPR` and `{!rsh} VAL_TYPE_EXPR`:
+```reach
+load: /examples/map-arbitrary-key/index.rsh
+range: 24 - 24
+```
+
+Such new mapping of linear state exemplified in the code samples above, returns a value which may be used to dereference particular mappings via `{!rsh} map[EXPR]`.
```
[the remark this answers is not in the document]
TheChronicMonster: These examples return a value which may be used to dereference particular mappings via `{!rsh} map[EXPR]`.

TheChronicMonster (changes requested):

TheChronicMonster (approved):

jeapostrophe (approved):

PR #941, opened 2022-05-09
CORE-1567: Use `.api_` In More Examples

chrisnevers:
Also, tweak the `.api_` code to handle PAY_EXPRs that are arrays


jeapostrophe (approved):
Approve given CI

PR #943, opened 2022-05-10
DCT-229

EBEREGIT:
### Sets: creation and modification

Added code samples

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -628,6 +628,7 @@ only be constructed in a consensus step.
 
 A `{!rsh} Set` may be modified by writing `{!rsh} s.insert(ADDRESS)` to install `{!rsh} ADDRESS` in the
 set, `{!rsh} s`, or `{!rsh} s.remove(ADDRESS)` to remove the `{!rsh} ADDRESS` from the set.
+Such modifications may only occur in a consensus step.
 
 The following example shows the usage of `{!rsh} s.insert(ADDRESS)`:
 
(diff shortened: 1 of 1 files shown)


docs/src/rsh/consensus/index.md
```diff
@@ -628,9 +628,31 @@ only be constructed in a consensus step.
 
 A `{!rsh} Set` may be modified by writing `{!rsh} s.insert(ADDRESS)` to install `{!rsh} ADDRESS` in the
 set, `{!rsh} s`, or `{!rsh} s.remove(ADDRESS)` to remove the `{!rsh} ADDRESS` from the set.
+
+The following example shows the usage of `{!rsh} s.insert(ADDRESS)`:
+
+```reach
+load: /hs/t/y/merit-badge.rsh
```
jeapostrophe, line 635: Do I recall correctly that you can't have a leading `/`? Or is that you need one? @TheChronicMonster

docs/src/rsh/consensus/index.md
```diff
@@ -628,9 +628,31 @@ only be constructed in a consensus step.
...
+```
+
+While the example below shows the usage of  `{!rsh} s.remove(ADDRESS)`
+
+```reach
+load: /examples/dominant-assurance/index.rsh
+range: 143 - 143
+```
+
 Such modifications may only occur in a consensus step.
```
jeapostrophe, line 646: This is confusing because of the interjection of the example

jeapostrophe (changes requested):

docs/src/rsh/consensus/index.md
```diff
@@ -644,7 +644,7 @@ load: /examples/dominant-assurance/index.rsh
 range: 143 - 143
 ```
 
-To check whether the address is in the set, `{!rsh} s.member(ADDRESS)` is used.
+To check whether the `{!rsh} ADDRESS` is in the set, `{!rsh} s.member(ADDRESS)` is used.
```
jeapostrophe, line 647: I actually like the original, because we're talking about the "idea" of an address. I do think the phrasing should be "an address"

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #945, opened 2022-05-10
docs: DCT-230

EBEREGIT:
### Types

Added code samples


docs/src/rsh/compute/index.md
```diff
@@ -411,23 +411,69 @@ Reach's @{defn("type")}s are represented in programs by the following identifier
 + @{ref("rsh", "Bool")} `{!rsh} Bool`, which denotes a boolean.
 + @{ref("rsh", "UInt")} `{!rsh} UInt`, which denotes an unsigned integer.
   @{ref("rsh", "UInt.max")} `{!rsh} UInt.max` is the largest value that may be assigned to a `{!rsh} UInt`.
+
+The following example shows the usage of `{!rsh} Null`, `{!rsh} Bool` and `{!rsh} UInt`:
```
jeapostrophe, line 415: I don't think these examples work well. This is supposed to be a concise list of all the types and now the flow of that list is greatly interrupted by a bunch of examples. I think you and @TheChronicMonster should think about how to do it differently.

jeapostrophe (changes requested):

PR #949, opened 2022-05-10
CORE-1488: Ensure API Return Functions Are Called

chrisnevers:


hs/src/Reach/APICut.hs
```diff
@@ -137,6 +150,11 @@ interactIn = interactX "in"
 interactOut :: DLExpr -> Bool
 interactOut = interactX "out"
 
+setApiDetails :: SLPart -> DLExpr -> Bool
+setApiDetails who = \case
```
jeapostrophe, line 154: I think `is_setApiDetails` or something is better... the word `set` feels like it is an effect

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #953, opened 2022-05-11
DCT-233

EBEREGIT:
### Numeric Casting

Added code samples


docs/src/rsh/compute/index.md
```diff
@@ -654,8 +654,37 @@ Reach has different unsigned integer types, `{!rsh} UInt` and `{!rsh} UInt256`.
...
+load: /examples/uint256/index.rsh
+range: 88 - 88
+```
+
+```reach
+load: /examples/uint256/index.rsh
+range: 92 - 93
+```
+
+Similarly, if `y` is a `{!rsh} UInt256`, then `{!rsh} UInt(y)` is `y` with as many bits as `{!rsh} UInt` has.
```
jeapostrophe, line 674: Same thing as I've said before

jeapostrophe (changes requested):

PR #973, opened 2022-05-13
CORE-1505: Support Multiple `invariant`s and Optional Labels

chrisnevers:


hs/src/Reach/AST/DL.hs
```diff
@@ -104,7 +105,7 @@ data DLSStmt
   | DLS_While
       { dls_w_at :: SrcLoc
       , dls_w_asn :: DLAssignment
-      , dls_w_inv :: DLSBlock
+      , dls_w_inv :: [(DLSBlock, Maybe ByteString)]
```
jeapostrophe, line 108: You should make a type for this. A question I have now is whether any bindings are shared (I don't think they need to be, but I'm curious)

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -951,6 +951,8 @@ solCom = \case
   DL_Let _ (DLV_Let _ dv) (DLE_PrimOp _ GET_COMPANION []) -> do
     addMemVar dv
     solLargeArg dv $ mdaToMaybeLA T_Contract Nothing
+  DL_Let _ (DLV_Eff) (DLE_GetUntrackedFunds {}) ->
+    return ""
```
jeapostrophe, line 955: Why this?

jeapostrophe: An impossible? Did you add a test case?

hs/src/Reach/Eval/Core.hs
```diff
@@ -5511,11 +5511,11 @@ doParallelReduce lhs (ParallelReduceRec {..}) = locAt slpr_at $ do
   let want lab = \case
         Just x -> return x
         Nothing -> expect_ $ Err_ParallelReduceIncomplete $ "missing " <> lab
-  inv_e <- want "invariant" pr_minv
```
jeapostrophe, line 5514: The docs say it is allowed multiple times (which I read as 1+) but I think you might have implemented many (0+)

hs/t/n/Err_ParallelReduceIncomplete.txt
```diff
@@ -1,4 +1,4 @@
-reachc: error[RE0083]: Parallel reduce incomplete: missing invariant
+reachc: error[RE0083]: Parallel reduce incomplete: missing while
```
jeapostrophe, line 1: This is still good I think

jeapostrophe (changes requested):
You should look at examples and see if there are any cases where we have an && in the invariant and use this new feature

hs/src/Reach/Connector/ETH_Solidity.hs
chrisnevers: I discovered this while tweaking your duoswap PR to use `.api_`. When the shared checks and bindings are duplicated into PAY_EXPR, the result of `getUntrackedFunds` is unused and updated to `DLV_Eff`. Maybe, the purity of `getUntrackedFunds` should be handled specially during optimization so it doesn't make it this far?

hs/src/Reach/AST/DL.hs
chrisnevers: OK & the bindings aren't shared

jeapostrophe (approved):

PR #974, opened 2022-05-13
DCT-425

Craig-Elre:
Update `digest` in the Computations page with an example and description.


jeapostrophe (changes requested):
You need to rebase

PR #983, opened 2022-05-15
convert tabs to spaces in `hs/Makefile`

hamirmahal:


mattaudesse:
@hamirmahal this should be closed - Makefiles are supposed to have tabs.
https://www.gnu.org/software/make/manual/html_node/Rule-Introduction.html

jeapostrophe:
Yea, it's horrible :'(

PR #986, opened 2022-05-16
prefixing directories used in docs

TheChronicMonster:
Directories that contain files that are used in documentation are prefixed with `docs`.


jeapostrophe:
I don't want to do this.

I believe the goal is to more easily know that if we change code, then the docs may need to be adjusted. I think that a better solution is the following:

If a file is referenced from the docs, the docs has to include a checksum of the file and it will error if the actual checksum doesn't match. This will cause a build failure and we'll know that it needs to be fixed.

Specifically:

```
load: /examples/rps-8-interact/index.rsh
```

will change to

```
load: /examples/rps-8-interact/index.rsh
md5: ee287e712cdfe8d91bbb038c383d25d3
```

What do you think?

TheChronicMonster:
I like this suggestion. Including a checksum feels more elegant.
I'm assuming that adding an md5 checksum to files can be automated and it
will be the author's responsibility to paste in the md5 string?

jeapostrophe:
Getting the md5 comes from a command line tool `md5sum`.

We could have @willghatch or @hamirmahal implement the feature and then add them all in

jeapostrophe:
If you like it, will you make a DCT task?

TheChronicMonster:
https://reachsh.atlassian.net/browse/DCT-429?atlOrigin=eyJpIjoiMDExOGNhOGEyYTlmNGFhZGI4OGEwNGNkYzk5YzQ5ODkiLCJwIjoiaiJ9

PR #989, opened 2022-05-16
DCT-248

EBEREGIT:
### `lastConsensusSecs`

Added code sample

### DIFF
--- a/docs/src/rsh/compute/index.md
+++ b/docs/src/rsh/compute/index.md
@@ -1840,6 +1840,11 @@ lastConsensusSecs()
 
 @{defn("lastConsensusSecs")} is like `{!rsh} lastConsensusTime`, except it returns the network seconds.
 
+```reach
+load: /examples/realtime/index.rsh
+range: 18 - 18
+```
+
 ### `thisConsensusTime` and `thisConsensusSecs`
 
 @{ref("rsh", "thisConsensusTime")}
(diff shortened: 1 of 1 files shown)


docs/src/rsh/compute/index.md
```diff
@@ -1840,6 +1840,11 @@ lastConsensusSecs()
 
 @{defn("lastConsensusSecs")} is like `{!rsh} lastConsensusTime`, except it returns the network seconds.
 
+```reach
+load: /examples/realtime/index.rsh
+range: 18 - 18
```
TheChronicMonster, line 1845: Same thing here. Samples should provide more context than the syntax reference. 

Let's find a different example so we're not showing a duplicate sample right after the other.

TheChronicMonster (changes requested):

TheChronicMonster (approved):

jeapostrophe:
Re: transition statement

TheChronicMonster:
> 

@jeapostrophe we want to avoid a string of "In the example below" statements in the documentation. 
However, perhaps we can structure the entries as:
syntax reference
precise explanation
sample
intuitive explanation referencing sample

What do you think?

jeapostrophe:
That's good for me.

TheChronicMonster:
> > 
> 
> @jeapostrophe we want to avoid a string of "In the example below" statements in the documentation. However, perhaps we can structure the entries as: syntax reference precise explanation sample intuitive explanation referencing sample
> 
> What do you think?

@EBEREGIT Please update this, and the other PRs with explanations of the samples.

jeapostrophe (approved):

PR #994, opened 2022-05-17
CORE-1458 More precise Algod queries

erer1243:


scripts/devnet-algo/algorand_data/config.json
```diff
@@ -1,4 +1,5 @@
 {
+    "MaxAPIResourcesPerAccount": 0,
```
jeapostrophe, line 2: Why?

js/stdlib/ts/ALGO.ts
```diff
@@ -1426,13 +1433,10 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
...
         const getLocalState = async (a:Address): Promise<AppStateKVs|undefined> => {
-          const ai = await getAccountInfo(a);
-          debug(`getLocalState`, ai);
-          const alss = ai['apps-local-state'] || [];
-          const fmtApplicationID = bigNumberToBigInt(ApplicationID);
-          const als = alss.find((x) => (x.id === fmtApplicationID));
-          debug(`getLocalState`, als);
-          return als ? als['key-value'] : undefined;
+          const client = await getAlgodClient();
+          const query = client.accountApplicationInformation(a, bigNumberToNumber(ApplicationID)) as unknown as ApiCall<AccountApplicationInfo>;
```
jeapostrophe, line 1437: You need to have a fallback that doesn't use `algod` and can work with just the indexer. You will test by setting the NODE_WRITE_ONLY to true and having a new function that is like `getAccountInfo` and will dispatch to either one.

js/stdlib/ts/ALGO.ts
```diff
@@ -2225,17 +2229,37 @@ const balancesOfM = async (acc: Account, tokens: Array<Token|null>): Promise<Arr
...
 };
 
+const balanceOfM = async (acc: Account, token?: Token): Promise<BigNumber | false> => {
+  if (token == null) {
+    return (await balancesOfM(acc, [null]))[0];
+  } else {
+    const tokenId = bigNumberToNumber(bigNumberify(token));
+    const addr = extractAddr(acc);
+    const client = await getAlgodClient();
+    const query = client.accountAssetInformation(addr, tokenId) as ApiCall<AccountAssetInfo>;
```
jeapostrophe, line 2252: Same comment. This cannot work on real nodes that don't allow you to read

js/stdlib/ts/ALGO.ts
```diff
@@ -211,11 +211,18 @@ type AppSchema = {
...
-  'assets'?: Array<AccountAssetInfo>,
+  'assets'?: Array<AssetHolding>,
   'apps-local-state'?: Array<AppState>,
   'apps-total-schema'?: AppSchema,
   'created-apps'?: Array<AppInfo>
 };
+type AccountAssetInfo = {
+  'asset-holding'?: AssetHolding
+};
+type AccountApplicationInfo = {
```
jeapostrophe, line 222: The naming scheme would be to have this be `Res` thing, because the node returns things tagged with the `round` and then have another type that is used for both the indexer and the node query result

jeapostrophe (changes requested):

scripts/devnet-algo/algorand_data/config.json
erer1243: This makes balancesOf never need the indexer, and so theoretically more accurate in less time

erer1243: p.s. 0 = infinity.

erer1243:
In this latest change, `contract::getLocalState` still has the potential to fail due to api rate limiting. `indexer:v2/accounts/{address}/apps-local-state` does not seem to conform to the spec defined in the api docs. its `apps-local-states` field is always returned as null, despite being marked as required. I commented out code using this, but left it in as it would be a superior implementation if the spec was followed.

Other than that, the code was tested with nodeCanRead and !nodeCanRead and works.

erer1243:
Having solved the confusion related to null responses to `apps-local-states` queries, I believe this is now finished. I made a pr on the indexer to fix this behavior. Tangentially, I tested updating our indexer version. That appeared to cause other bugs (or random failures, I didn't bother looking into it; if we still want to consider updating, it should be a new ticket anyway) so I didn't include any of the work I did on that front.

jeapostrophe (approved):

jeapostrophe:
Thanks!

PR #1002, opened 2022-05-17
CORE-1422 Verification --error-format-json

erer1243:
Added an SMTError that is used in Verify/SMT::display_fail to display verification errors as json. Json versions are output to stderr, with a prefix `error: `. The message is passed exactly as it would be printed, except never with color codes. Also created a new error code "RV0000" as a catch-all for verification errors.

Example comparison of outputs:
image


jeapostrophe (approved):

hamirmahal (approved):
It looks like the extension can properly utilize verification errors in this format.

hamirmahal:
https://user-images.githubusercontent.com/43425812/168916493-15a22952-4d22-4c37-b483-a11cbab27c5f.mp4

PR #1008, opened 2022-05-18
CORE-1823

jeapostrophe:


hs/app/reach/Main.hs
```diff
@@ -1112,13 +1105,7 @@ compile = command "compile" $ info f d
 
           (cd "$$HS" && make expand)
```
mattaudesse, line 1106: I think we want
```suggestion
          (cd "$$HS" && make hs-build)
```
because `expand` is a prerequisite of `hs-build` and `stack exec` doesn't force a (re-)build.

jeapostrophe: Go for it

jeapostrophe: You own this branch now? :)

PR #1015, opened 2022-05-19
DCT-279

EBEREGIT:
### Error code 56

Added code samples

### DIFF
--- a/docs/src/rsh/errors/index.md
+++ b/docs/src/rsh/errors/index.md
@@ -1032,9 +1032,31 @@ for the function you are attempting to use and provide it with the correct argum
 ## {#RE0056} RE0056
 
 This error indicates that you are attempting to create a variable, although another
-variable in the scope uses the same name. In Reach, identifier shadowing is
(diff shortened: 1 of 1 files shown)


docs/src/rsh/errors/index.md
```diff
@@ -1032,10 +1032,32 @@ for the function you are attempting to use and provide it with the correct argum
...
+range: 3 - 8
+```
+
+In Reach, identifier shadowing is not allowed. 
+You can fix this issue by renaming your variable or moving one of the
 variable declarations to another scope where it does not conflict with the other.
 
+Example:
+
+```reach
```
jeapostrophe, line 1050: You should make this a file in `/hs/t/y`

jeapostrophe (changes requested):

docs/src/rsh/errors/index.md
TheChronicMonster: @EBEREGIT if you need help making a file then let's talk.

PR #1017, opened 2022-05-19
DCT-363

Craig-Elre:
Update `newAccountFromSecret` with a code sample and a description

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -339,7 +339,7 @@ range: 18-25
 ```
 
 In this code, if the user chooses to not create a new account, then the DApp uses `{!rsh} newAccountFromSecret` to create a new account.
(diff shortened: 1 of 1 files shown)


docs/src/frontend/index.md
```diff
@@ -331,6 +331,16 @@ stdlib.newAccountFromSecret(secret: string) => Promise<acc>
...
 
+Example:
+
+```js
+load: /examples/rps-8-interact/index.mjs
+range: 18-25
+```
+
+In this code, if the user chooses to not create a new account, then the DApp uses `{!rsh} newAccountFromSecret` to create a new account.
+The `acc` object `{!rsh} awaits` the user to input the secret and then sets the provided account to itself.
```
jeapostrophe, line 342: s outside `

jeapostrophe (changes requested):

PR #1021, opened 2022-05-20
DCT-283

EBEREGIT:
### Error code 73

Added code samples


docs/src/rsh/errors/index.md
```diff
@@ -1287,8 +1287,24 @@ f(Maybe(UInt).Some(1));
...
+Example:
+
+```reach
+load: /hs/t/n/Err_Switch_DoubleCase.rsh
+range: 10 - 16
+```
+
 You can fix this error by deleting one of the branches, ensuring there is only one branch per variant.
 
+```reach
```
jeapostrophe, line 1299: You should make this an example in `/hs/t/y`

jeapostrophe (changes requested):

PR #1022, opened 2022-05-20
DCT-300

EBEREGIT:
### Error code 74

Added code samples


docs/src/rsh/errors/index.md
```diff
@@ -1294,9 +1294,23 @@ You can fix this error by deleting one of the branches, ensuring there is only o
...
 
+Example:
+
+```reach
+load: /hs/t/n/Err_Switch_MissingCases.rsh
+range: 11 - 13
+```
+
 You can fix this issue by adding the missing cases listed
 in the error message.
```
jeapostrophe, line 1305: It would be better to show a correct version of the exact same code... is that what that example is?

PR #1025, opened 2022-05-20
DCT-275

Craig-Elre:
Added a code sample for issue `RE0003` and how to fix it.

### DIFF
--- a/docs/src/rsh/errors/index.md
+++ b/docs/src/rsh/errors/index.md
@@ -158,6 +158,20 @@ Array.iota(arr.length).map((i) => {
 This error indicates that a statement block returns a non-null value although a `{!rsh} null` value is expected.
 The block should either use a `{!rsh} return null;` statement or no return statement at all.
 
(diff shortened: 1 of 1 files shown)


docs/src/rsh/errors/index.md
```diff
@@ -158,6 +158,20 @@ Array.iota(arr.length).map((i) => {
...
+
+``` reach
+load: /hs/t/n/Err_Block_NotNull.rsh
+range: 3 - 6
+```
+
+This error can be corrected by removing the `{rsh} return` statement:
+
+```reach
+load: /examples/api-multiple-calls/index.rsh
```
TheChronicMonster, line 171: I suggest we use [data_null](https://github.com/reach-sh/reach-lang/blob/master/hs/t/y/data_null.rsh) for the correct example.

TheChronicMonster (changes requested):

TheChronicMonster (approved):

PR #1028, opened 2022-05-21
Dct 429 docs must include a checksum of a referenced file

hamirmahal:
This change adds `md5`s for about 223 examples in the documentation, to hopefully more easily catch documentation that is out-of-sync with `examples/`, as discussed in https://github.com/reach-sh/reach-lang/pull/986#issuecomment-1127884025.

This change also adds testing to our continuous integration that should catch examples added without an `md5sum`, or when an example linked from the documentation does not have an `md5sum` that matches the `md5` in the documentation.

The audio in this screen recording might be a bit loud, so you may want to lower your volume a bit before listening to it.

https://user-images.githubusercontent.com/43425812/169669105-63155536-7b7a-4128-8eb4-c30cd276c9d7.mp4

### DIFF
--- a/.circleci/config.pre.yml
+++ b/.circleci/config.pre.yml
@@ -145,6 +145,7 @@ jobs:
         git rm -r .
         cp -r /tmp/workspace/docs/* .
         [ -f index.html ] || exit 1
+        touch .nojekyll
         git add .
         # https://stackoverflow.com/questions/8123674/how-to-git-commit-nothing-without-an-error
         git diff-index --quiet HEAD || (git commit -m "[ci skip] docs for ${CIRCLE_SHA1}" && touch .did_change)
--- a/docs/dev/generator.js
+++ b/docs/dev/generator.js
@@ -1,7 +1,6 @@
 import { setTimeout } from 'timers/promises';
 import CleanCss from 'clean-css';
 import { createHash } from 'crypto';
-import { readFileSync, writeFileSync } from 'fs';
 import fs from 'fs-extra';
 import * as htmlMinify from 'html-minifier';
 import path from 'path';
@@ -295,7 +294,7 @@ const processHtml = async ({iPath, oPath}) => {
 const processJs = async ({iPath, oPath}) => {
   const input = await fs.readFile(iPath, 'utf8');
   const output = false ? { code: input } : new UglifyJS.minify(input, {});
-  if (output.error) throw output.error;
+  if (output.error) { throw output.error };
   await writeFileMkdir(oPath, output.code);
 }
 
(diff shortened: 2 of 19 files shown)


Craig-Elre (dismissed):
The checksums are accurate to my generations. Good job!

hamirmahal:
Thanks, Craig.

.circleci/config.pre.yml
```diff
@@ -168,6 +168,23 @@ jobs:
               }
             }]
           }
+  "docs-check-checksums":
```
jeapostrophe, line 171: No, this should be part of the generator itself. It should just error if a checksum doesn't match

scripts/checkChecksums.ts
```diff
@@ -0,0 +1,63 @@
...
+  );
+  if (r1) {
+    const [ _fullMatch, programPath ] = r1;
+    const nextLine: string = a[index + 1];
+    const r2: RegExpMatchArray | null = nextLine.match(
+      /md5: (.+)/
+    );
+    if (r2) {
+      const topDir: string = dirname(dirname(__filename));
+      const actualMD5WithNewline: string = execSync(`md5sum ${
```
jeapostrophe, line 26: Don't run the `md5sum` program. Node already has the `crypto` module with the `createHash` function (https://nodejs.org/api/crypto.html#cryptocreatehashalgorithm-options) and `'MD5'` is one of the options.

scripts/checkChecksums.ts
```diff
@@ -0,0 +1,63 @@
...
+        join(topDir, programPath)
+      } | awk '{print $1}'`).toString();
+      const actualMD5: string = actualMD5WithNewline.slice(0, -1);
+      const [ _fullMatch, md5InDocs ] = r2;
+      if (actualMD5 !== md5InDocs) {
+        errorString += programPath + '\n';
```
jeapostrophe, line 32: I wouldn't bother with errors like this. Throw JSON blobs like the other errors do

jeapostrophe (changes requested):

scripts/checkChecksums.ts
hamirmahal: Oh, neat! I didn't know `node` had that. Thanks for the tip!

.circleci/config.pre.yml
hamirmahal: Okay. The newest commits get rid of the continuous integration test.

hamirmahal: Okay. The new commits don't have a continuous integration test.

scripts/checkChecksums.ts
hamirmahal: I'm not exactly sure what you mean by
> Throw JSON blobs like the other errors do

[5b0bfdf](https://github.com/reach-sh/reach-lang/pull/1028/commits/5b0bfdfecff32a370199177eb6969ee4d8460754) just `throw`s an `Error` if an `md5` is missing or doesn't match on `make`.

How far is that from what we want?

jeapostrophe (approved):

jeapostrophe:
Thank you Hamir! I made a small tweak

hamirmahal:
You're welcome, Jay! I'm happy I could help.

PR #1033, opened 2022-05-23
DCT-310

EBEREGIT:
### Error Message 101

Added code sample

### DIFF
--- a/hs/t/y/Fix_Throw_No_Catch.rsh
+++ b/hs/t/y/Fix_Throw_No_Catch.rsh
@@ -1,0 +1,13 @@
+'reach 0.1';
+
+export const main =
+  Reach.App(
+    {},
+    [Participant('A', {})],
+    (A) => {
+      try {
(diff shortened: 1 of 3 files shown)


hs/t/y/Fix_Throw_No_Catch.rsh
```diff
@@ -0,0 +1,13 @@
+'reach 0.1';
+
+export const main =
+  Reach.App(
+    {},
+    [Participant('A', {})],
+    (A) => {
+      try {
```
TheChronicMonster, line 8: If we add
```
A.publish();
commit();
```
Just before the `try` then that will remove warning RW0004 and will make a cleaner `txt` output.

TheChronicMonster (changes requested):

jeapostrophe:
`hs-test` still fails

TheChronicMonster (approved):

PR #1035, opened 2022-05-23
Fix JS for Maps With Hashed Keys

chrisnevers:

### DIFF
--- a/docs/src/changelog/index.md
+++ b/docs/src/changelog/index.md
@@ -7,7 +7,6 @@ Versions and changes-within-versions are listed in reverse-chronological order: 
 
 Version 0.1.10 is the current Reach release candidate version.
 
-+ 2022/05/23: The backend interface to the compiled objects was updated, so you'll need to recompile for this release.
 + 2022/05/14: Added `{!cmd} reach support` sub-command.
 + 2022/05/13: Support custom violation messages for `{!rsh} invariant`s.
 + 2022/05/13: Support multiple `{!rsh} invariant`s for `{!rsh} while` loops and `{!rsh} parallelReduce`.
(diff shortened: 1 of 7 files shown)


examples/map-tuple-key/index.mjs
```diff
@@ -0,0 +1,35 @@
+import { loadStdlib } from '@reach-sh/stdlib';
+import * as backend from './build/index.main.mjs';
+
+const stdlib = await loadStdlib();
+
+if(stdlib.connector == 'ALGO') {
```
hamirmahal, line 6:
```suggestion
if(stdlib.connector === 'ALGO') {
```

docs/src/changelog/index.md
```diff
@@ -7,6 +7,7 @@ Versions and changes-within-versions are listed in reverse-chronological order:
 
 Version 0.1.10 is the current Reach release candidate version.
 
++ 2022/05/23: The backend interface to the compiled objects was updated, so you'll need to recompile for this release.
```
jeapostrophe, line 10: This is not true, because you didn't change the version or interface to them

hs/src/Reach/Backend/JS.hs
```diff
@@ -324,6 +324,28 @@ jsArg_m = \case
...
+  T_Bytes {} -> True
+  T_Array {} -> True
+  T_Data {} -> True
+  T_Tuple {} -> True
+  T_Object {} -> True
+  T_Struct {} -> True
+
+jsMapKey :: DLArg -> App Doc
+jsMapKey k =
+  case shouldHashMapKey $ argTypeOf k of
```
jeapostrophe, line 345: I would consider making this just always true... why would it be bad to always use digests?

jeapostrophe (changes requested):

hs/src/Reach/Backend/JS.hs
chrisnevers: Map lookups would fail because the connectors only hash keys that containers or strings

jeapostrophe (changes requested):
Can you remove the submodule?

PR #1044, opened 2022-05-24
simplify `gatherSearchData` in `dev/generator.js`

hamirmahal:
`gatherSearchData` doesn't `await` anything, so it doesn't need to be `async`hronous.


jeapostrophe:
I want to leave it the way it is so we can add asynchrony in the future if we need to.

PR #1048, opened 2022-05-25
DCT-282

Craig-Elre:
Add an example of an issue and how to fix it for `RE0069`

### DIFF
--- a/docs/src/rsh/errors/index.md
+++ b/docs/src/rsh/errors/index.md
@@ -2248,7 +2248,7 @@ a `{!rsh} Participant` or `{!rsh} ParticipantClass`.
 For example, the program below erroneously uses `{!rsh} Anybody.publish()` without declaring any `{!rsh} Participant` or `{!rsh} ParticipantClass`:
 
 ```reach
-load: ./hs/t/n/Err_No_Participants.rsh
+load: /hs/t/n/Err_No_Participants.rsh
 range: 3 - 9
 ```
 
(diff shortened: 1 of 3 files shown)


hs/t/y/underscore_unread.rsh
```diff
@@ -0,0 +1,11 @@
+'reach 0.1';
```
TheChronicMonster, line 1: Let's generate the output and add that to the PR, as well.

TheChronicMonster (changes requested):

docs/src/rsh/errors/index.md
```diff
@@ -2236,7 +2248,7 @@ a `{!rsh} Participant` or `{!rsh} ParticipantClass`.
 For example, the program below erroneously uses `{!rsh} Anybody.publish()` without declaring any `{!rsh} Participant` or `{!rsh} ParticipantClass`:
 
 ```reach
-load: /hs/t/n/Err_No_Participants.rsh
+load: ./hs/t/n/Err_No_Participants.rsh
```
jeapostrophe, line 2251: Doesn't this need to be changed?

jeapostrophe (changes requested):

docs/src/rsh/errors/index.md
jeapostrophe: Maybe you need to rebase?

Craig-Elre: Fixing it caused issues with merging for Hamir(who already had a fix), so I switched it back.

jeapostrophe: Right now, if we were to merge yours, then it would be wrong

Craig-Elre: Fixed it again

PR #1049, opened 2022-05-25
Algorand - support 16+ arg APIs.

willghatch:
This supersedes PR #767.  This is basically a squashed version of what was in that that only deals with Algorand and not ETH.

I'm making a PR though I'm not 100% certain all the CI tests pass, because I don't actually know how to make the CI run without making a PR.  Anyway, my test is working but I don't know yet that I didn't break other things while rebasing.

### DIFF
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -3229,8 +3229,7 @@ cmeth sigi = \case
 bindFromArgs :: [DLVarLet] -> App a -> App a
 bindFromArgs vs m = do
   let goSingle (v, i) = sallocVarLet v False (code "txna" ["ApplicationArgs", texty i] >> cfrombs (varLetType v))
-  let goSingles singles k =
-        foldl' (flip goSingle) k (zip singles [(1 :: Integer) ..])
+  let goSingles singles k = foldl' (flip goSingle) k (zip singles [(1 :: Integer) ..])
   case splitArgs vs of
     (vs', Nothing) -> do
       goSingles vs' m
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -2316,7 +2316,6 @@ splitArgs l =
     False -> (l, Nothing)
     True -> (before, Just after) where (before, after) = splitAt 14 l
 
-
 signatureStr :: Bool -> String -> [DLType] -> Maybe DLType -> String
 signatureStr addr2acc f args mret = sig
   where
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -2120,26 +2120,39 @@ ce = \case
         cbs $ sigStrToBytes sig
         makeTxn1 "ApplicationArgs"
         accountsR <- liftIO $ newCounter 1
-        forM_ as $ \a -> do
-          ca a
-          let t = argTypeOf a
-          ctobs t
-          case t of
-            -- XXX This is bad and will not work in most cases
-            T_Address -> do
-              incResource R_Account a
-              let m = makeTxn1 "Accounts"
-              case r_addr2acc of
-                False -> do
-                  op "dup"
-                  m
-                True -> do
-                  i <- liftIO $ incCounter accountsR
-                  m
-                  cint $ fromIntegral i
-                  ctobs $ T_UInt UI_Word
-            _ -> return ()
-          makeTxn1 "ApplicationArgs"
+        let processArg a = do
+              ca a
+              let t = argTypeOf a
+              ctobs t
+              case t of
+                -- XXX This is bad and will not work in most cases
+                T_Address -> do
+                  incResource R_Account a
+                  let m = makeTxn1 "Accounts"
+                  case r_addr2acc of
+                    False -> do
+                      op "dup"
+                      m
+                    True -> do
+                      i <- liftIO $ incCounter accountsR
+                      m
+                      cint $ fromIntegral i
+                      ctobs $ T_UInt UI_Word
+                _ -> return ()
+        let processArg' a = do
+              processArg a
+              makeTxn1 "ApplicationArgs"
+        let processArgTuple tas = do
+              --cconcatbs_ processArg $ map (\a -> (argTypeOf a, ca a)) tas
+              cconcatbs_ (const $ return ()) $
+                map (\a -> (argTypeOf a, processArg a)) tas
+              makeTxn1 "ApplicationArgs"
+        case splitArgs as of
+          (_, Nothing) -> do
+            forM_ as processArg'
+          (as14, Just asMore) -> do
+            forM_ as14 processArg'
+            processArgTuple asMore
         -- XXX If we can "inherit" resources, then this needs to be removed and
         -- we need to check that nnZeros actually stay 0
         forM_ (r_assets <> nnRecv) $ \a -> do
@@ -2294,6 +2307,15 @@ multipleApiCalls w = do
   case M.lookup w apiCalls of
     Just n  -> return $ n > 1
     Nothing -> return False
+
+splitArgs :: [a] -> ([a], Maybe [a])
+splitArgs l =
+  -- If there are more than 15 args to an API on ALGO,
+  -- args 15+ are packed as a tuple in arg 15.
+  case 15 < (length l) of
+    False -> (l, Nothing)
+    True -> (before, Just after) where (before, after) = splitAt 14 l
+
 
 signatureStr :: Bool -> String -> [DLType] -> Maybe DLType -> String
 signatureStr addr2acc f args mret = sig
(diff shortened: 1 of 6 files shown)


willghatch:
Also, I'm confused by the `typeSig_` function recurring by calling `typeSig` (no underscore) instead of `typeSig_` (with underscore).  I don't know what the purpose of the `addr2acc` flag is, but I would naively suspect that it should be passed through.  Why isn't it?  This may mean there is a bug for generating signatures of 16+ args because it creates a tuple then maps the typeSig_ function over the list that now includes a tuple instead of all raw arguments.

willghatch:
Well... the CI never ran.  What was different this time than previous PRs?

willghatch:
The `build-sink` test that failed seems spurious to me.  That said, I don't understand what it's supposed to be.  What is `build-sink` for?

jeapostrophe:
`build-sink` is what prints the message in Slack. It was:

*:warning: FAIL* willghatch/<https://github.com/reach-sh/reach-lang/tree/algo-arg-many-rebase|algo-arg-many-rebase>/<https://github.com/reach-sh/reach-lang/pull/1049|PR> tests > 4 of 1278 failed! <https://circleci.com/gh/reach-sh/reach-lang/132558|more...>\n- *hs-test* 1: <https://circleci.com/gh/reach-sh/reach-lang/132547|more...>\n- *ETH* 1: *<https://output.circle-artifacts.com/output/job/3e1efdff-2c0f-4e06-82d2-d7ebbec2d26a/artifacts/12/tmp/artifacts/ETH.api-arg-many.gz|api-arg-many>*\n- *ALGO* 1: *<https://output.circle-artifacts.com/output/job/e594335a-4da8-45e6-a585-69da34c012b5/artifacts/3/tmp/artifacts/ALGO.many-args.gz|many-args>*\n- *CFX* 1: *<https://output.circle-artifacts.com/output/job/f55241fe-fab7-4131-87f3-2e47ee2c8864/artifacts/2/tmp/artifacts/CFX.api-arg-many.gz|api-arg-many>*\n*:warning: DANGER MERGE*: `8ebb5be0`

jeapostrophe:
> Also, I'm confused by the `typeSig_` function recurring by calling `typeSig` (no underscore) instead of `typeSig_` (with underscore). I don't know what the purpose of the `addr2acc` flag is, but I would naively suspect that it should be passed through. Why isn't it?

It is because of https://docs.reach.sh/rsh/consensus/#p_77

We can't change `address` into `account` inside of a tuple, because we don't map over the argument values to convert them (yet)

> This may mean there is a bug for generating signatures of 16+ args because it creates a tuple then maps the typeSig_ function over the list that now includes a tuple instead of all raw arguments.

The relevant compile time code is https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Connector/ALGO.hs#L2132-L2140

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3199,9 +3225,20 @@ cmeth sigi = \case
...
-  foldl' (flip go) m (zip vs [(1::Integer) ..])
+  let goSingle (v, i) = sallocVarLet v False (code "txna" ["ApplicationArgs", texty i] >> cfrombs (varLetType v))
+  let goSingles singles k =
+        foldl' (flip goSingle) k (zip singles [(1 :: Integer) ..])
+  -- When there are more than 15 args, the 15th arg is a tuple of args 15+
+  case 15 < (length vs) of
+    False -> do
+      goSingles vs m
+    True -> do
+      let (vs14, vsMore) = splitAt 14 vs
```
jeapostrophe, line 3236: This pattern is so common. I think I want a function like:
```
splitArgs :: [a] -> ([a], Maybe [a])
splitArgs l =
 case 15 < (length l) of
   False -> (l, Nothing)
   True -> (before, Just after) where (before, after) = splitAt 14 l
```
and then adapt the code to only branch on the Maybe rather than the list. I think it will de-duplicate some code

jeapostrophe (changes requested):
One comment, plus the CI is failing on your example

willghatch:
The reason it was failing that test was due to tuple-izing the signature.  The latest patch makes it use a tuple only conditionally.  I'm still of the opinion that we should leave the signature alone in all cases (IE that's still my favorite interpretation of the spec).

jeapostrophe:
> > > Also, I'm confused by the `typeSig_` function recurring by calling `typeSig` (no underscore) instead of `typeSig_` (with underscore). I don't know what the purpose of the `addr2acc` flag is, but I would naively suspect that it should be passed through. Why isn't it?
> > 
> > 
> > It is because of https://docs.reach.sh/rsh/consensus/#p_77
> > We can't change `address` into `account` inside of a tuple, because we don't map over the argument values to convert them (yet)
> 
> So... are tokens represented by accounts? I think I'm still confused.
> 
> Or did you mean to link what is now [p75](https://docs.reach.sh/rsh/consensus/#p_75)? Mutable links are frustrating.
> 
> Do you mean that we're not doing conversion on the receiving end of an API, and that's why we don't want to map the signature conversion over a tuple?

There's an option called `opts.addressToAccount`. This makes it so that when call `f :: Address -> UInt` with `f(Jay)` it turns it into `f' :: Account -> UInt` and does `f'(1)` where `Accounts[1] = Jay`. It is possible to write `g' :: (Account, UInt) -> UInt` and have `g'((1, 42))` behave like `g((Jay, 42))` but Reach doesn't support that. So, we don't recursively change `Address` into `Account` when you enable `addressToAccount`.

> > > This may mean there is a bug for generating signatures of 16+ args because it creates a tuple then maps the typeSig_ function over the list that now includes a tuple instead of all raw arguments.
> > 
> > 
> > The relevant compile time code is https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Connector/ALGO.hs#L2132-L2140
> 
> In my PR I've mapped that so that despite being in a tuple, that conversion is still respected. So... do I need to adjust the argument handling on the receiving side? I'm a little worried about the TODO comments about how this conversion is bad and doesn't work in most cases, since I'm not really sure what the rationale behind this conversion is.
> 
> Anyway, for the moment I'll write that splitting function.

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2295,11 +2308,24 @@ multipleApiCalls w = do
...
+    False -> (l, Nothing)
+    True -> (before, Just after) where (before, after) = splitAt 14 l
+
+
+signatureStr :: Bool -> Bool -> String -> [DLType] -> Maybe DLType -> String
+signatureStr addr2acc forRemote f args mret = sig
   where
     rets = fromMaybe "" $ fmap typeSig mret
-    sig = f <> "(" <> intercalate "," (map (typeSig_ addr2acc) args) <> ")" <> rets
+    tupledArgs = case (forRemote, splitArgs args) of
```
jeapostrophe, line 2324: As discussed in call, this is wrong.

jeapostrophe (changes requested):

willghatch:
Ok, here is a version that does not change signatures.  (The way I wanted to in the beginning! :p)

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2295,6 +2308,15 @@ multipleApiCalls w = do
...
 
+splitArgs :: [a] -> ([a], Maybe [a])
+splitArgs l =
+  -- If there are more than 15 args to an API on ALGO,
+  -- args 15+ are packed as a tuple in arg 15.
+  case 15 < (length l) of
+    False -> (l, Nothing)
+    True -> (before, Just after) where (before, after) = splitAt 14 l
+
+
```
jeapostrophe, line 2319:
```suggestion
```

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3199,9 +3229,18 @@ cmeth sigi = \case
 
 bindFromArgs :: [DLVarLet] -> App a -> App a
 bindFromArgs vs m = do
-  -- XXX deal with >15 args
-  let go (v, i) = sallocVarLet v False (code "txna" ["ApplicationArgs", texty i] >> cfrombs (varLetType v))
-  foldl' (flip go) m (zip vs [(1::Integer) ..])
+  let goSingle (v, i) = sallocVarLet v False (code "txna" ["ApplicationArgs", texty i] >> cfrombs (varLetType v))
+  let goSingles singles k =
+        foldl' (flip goSingle) k (zip singles [(1 :: Integer) ..])
```
jeapostrophe, line 3234:
```suggestion
  let goSingles singles k = foldl' (flip goSingle) k (zip singles [(1 :: Integer) ..])
```

jeapostrophe (changes requested):

jeapostrophe:
Waiting for CI

willghatch:
> but Reach doesn't support that

Is it something we would want to support in the future that is just not worth changing now, or is there a reason we don't want to support it?

jeapostrophe:
> > but Reach doesn't support that
> 
> Is it something we would want to support in the future that is just not worth changing now, or is there a reason we don't want to support it?

It is hard to do and I think that Algorand will drop the need to do it at all

willghatch:
I'm pretty sure the failure is flakiness on the part of that particular test.  It passed for my commit before your formatting, and only failed post-formatting.  I have not had it fail for me locally yet.  But the error is that an API call is made in a wrong state, and the api-multiple-calls had an extra publish/commit pair that thus gave rise to an extra state that was not yet ready for an API call, but one that the Alice participant could publish and pass before any API call was made... usually.  So I think this fixes it.  Anyway I hope so.

willghatch:
To add evidence:  Here is the dot graph of the state diagram before my fix (note that it was failing in state 1):

api-multiple-calls_old

And here is the dot graph after my commit:

api-multiple-calls_new

jeapostrophe:
Awesome!

PR #1051, opened 2022-05-25
DCT-302

Craig-Elre:
Add an example of the error to `RE0077` and an example of how to fix it.


docs/src/rsh/errors/index.md
```diff
@@ -1353,9 +1353,21 @@ assert(2 == 2, "5th assertion")
...
+```reach
+load: /hs/t/n/Err_RecursionDepthLimit.rsh
+range: 3 - 5
+```
+
 You can fix this issue by re-writing your recursive function into an iterative set of
 statements, e.g. `{!rsh} while` loop.
 
+```reach
+load: /hs/t/y/while_preamble.rsh
```
TheChronicMonster, line 1367: I would prefer to include a working version of `Err_RecursionDepthLimit` or an erroneous version of `while_preamble` so that the files are more directly relatable to one another.

TheChronicMonster (changes requested):

PR #1086, opened 2022-06-01
DCT-370

Craig-Elre:
Add example and description of `getAddress`

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -444,6 +444,18 @@ acc.getAddress() => string
 
 Returns the account's address as a string. The format of this string varies across connectors.
 
+Example:
+
+```js
+load: /examples/interact-this/index.mjs
+md5: 34eaa215660697ce8c5ba87387c8f0f1
+range: 14-19
+```
+
+In this code snippet, `{!js} getAddress` is being called in order to obtain the individual addresses of `Alice` and `Bob` that were created with the `{!js} newTestAccounts` code.
+The addresses are then written to the console so that each participant can see both addresses.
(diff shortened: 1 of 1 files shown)


docs/src/frontend/index.md
```diff
@@ -444,6 +444,18 @@ acc.getAddress() => string
...
 
+Example:
+
+```js
+load: /examples/interact-this/index.mjs
+md5: 34eaa215660697ce8c5ba87387c8f0f1
+range: 14-19
+```
+
+In this code snippet, `{!js} getAddress` is being called in order to obtain the individual addresses of `Alice` and `Bob` that were created with the `{!js} newTestAccounts` code.
```
TheChronicMonster, line 455: We can tighten up the language:
In this snippet, `{!js} getAddress` obtains the addresses of `Alice` and `Bob`.
Next, each participant's address is logged.

TheChronicMonster (changes requested):

docs/src/frontend/index.md
Craig-Elre: I think simplifying it that much makes it a little more difficult for newer programmers to understand, but I agree it is a bit wordy. How about this:

In this snippet, `{!js} getAddress` obtains the individual addresses of `Alice` and `Bob` created by `{!js} newTestAccounts`.
Next, each participant's address is logged.

TheChronicMonster: Sounds good. 
You'll need to rebase, also.

TheChronicMonster (dismissed):

TheChronicMonster (approved):

PR #1087, opened 2022-06-02
DCT-302

Craig-Elre:
Change examples to be from our repo instead of hard coded. Didn't really make sense to have both hard-coded and repo examples in this particular error, since the repo examples added no additional information.


TheChronicMonster (changes requested):
In order to make properly, every `load` in this file will need to be paired with an `md5`

PR #1094, opened 2022-06-03
DCT-394

EBEREGIT:
### Frontends > Utilities > protect

Added code sample

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -992,7 +992,7 @@ range: 349 -352
 
 This code tests to see if variables that are supposed to be `null` returns `null`.
 Line 350 `{!rsh} asserts` that `null` must return a result that is of type `T_Null`.
-Line 351 `{!rsh} asserts` that type `T_Null` must be returned by`undefined`.
+Line 351 `{!rsh} asserts` that type `T_Null` must be returned by `undefined`.
 
 ---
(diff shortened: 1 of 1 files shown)


docs/src/frontend/index.md
```diff
@@ -982,6 +982,18 @@ stdlib.protect(t, x) => x
...
+Example:
+
+```reach
+load: /examples/js-tests/index.mjs
+md5: 389bd098127db61d413f7586e81b561f
+range: 349 -352
+```
+
+This code tests to see if variables that are supposed to be `null` returns `null`.
+Line 350 `{!rsh} asserts` that `null` must return a result that is of type `T_Null`.
```
TheChronicMonster, line 994: Line 350 `{!rsh} asserts` that `null` has Reach type `T_Null`.
Line 351 `{!rsh} asserts` that `undefined` has Reach type `T_Null`.

TheChronicMonster (changes requested):

docs/src/frontend/index.md
```diff
@@ -982,6 +982,18 @@ stdlib.protect(t, x) => x
...
 
+Example:
+
+```reach
+load: /examples/js-tests/index.mjs
+md5: 389bd098127db61d413f7586e81b561f
+range: 349 -352
+```
+
+This code tests to see if variables that are supposed to be `null` returns `null`.
```
hamirmahal, line 993:
```suggestion
This code tests to see if variables that are supposed to be `null` return `null`.
```

TheChronicMonster (changes requested):

TheChronicMonster (approved):

PR #1095, opened 2022-06-03
make M1 Macs be able to generate correct output

hamirmahal:
This should fix an issue where running `scripts/gen-test-output.sh` on MacOS was generating incorrect output, like that in the original file in 4e7608335edb385ecba32f1682f268f591b1cf28.

I believe `brew install gnu-sed` will be necessary to run the macOS version of `gen-test-output.sh`, which is `scripts/gen-test-output-MacOS.sh`, @EBEREGIT and @TheChronicMonster.

### DIFF
--- a/scripts/gen-test-output.sh
+++ b/scripts/gen-test-output.sh
@@ -2,7 +2,7 @@
 # this script replicates the behavior of testCompileOut in /hs/test/Reach/Test_Compiler.hs
 case "$1" in
   *.rsh) RSH="$1" ;;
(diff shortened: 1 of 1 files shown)


jeapostrophe (changes requested):
It would be better to change the other program to use a variable for `sed` like `$SED` and have a check to see whether it has GNU or BSD sed

jeapostrophe:
Excellent

hamirmahal:
Thanks!

PR #1110, opened 2022-06-07
DCT-411

EBEREGIT:
### Frontends > Util > formatAddress

Added code sample

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -1207,6 +1207,18 @@ Formats the address in the way the user would expect to see it.
 + On Ethereum, it is a hex-encoded string starting with `{!js} '0x'`.
 + On Algorand, it is a base32-encoded string, ending with the checksum.
 
+Example:
+
+```js
+load: /examples/getUntrackedFunds1/index.mjs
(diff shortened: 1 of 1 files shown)


docs/src/frontend/index.md
```diff
@@ -1207,6 +1207,18 @@ Formats the address in the way the user would expect to see it.
...
+Example:
+
+```js
+load: /examples/getUntrackedFunds1/index.mjs
+md5: d743dd6357719dfd1ba3fcd540ab4e49
+range: 32 - 37
+```
+
+This code is a `{!rsh} function` that takes in `addr` on line 32.
+Line 33 formats `addr` based on the consensus network, in the way the user would expect to see it.
```
TheChronicMonster, line 1219: Line 33 formats `addr` based on the consensus network's encoding. 
We can remove "in the way the user would expect to see it" 
This phrase is vague and adds little value. Also, the same phrase was used just above the example. 

To compliment this suggestion, Line 1205, "Formats the address in the way the user would expect to see it:"
Should end with a colon ":" rather than a period since it then presents how the user would expect to see the address.

TheChronicMonster (approved):

PR #1111, opened 2022-06-07
DCT-366

Craig-Elre:
Add a code sample and description for `fundFromFaucet`


docs/src/frontend/index.md
```diff
@@ -384,11 +384,20 @@ stdlib.fundFromFaucet(account, balance) => Promise<void>
 ```
 
 Adds the given balance of network tokens to a Reach account abstraction.
-This can only be used in private testing scenarios,
-as it uses a private faucet to issue network tokens,
-as well as certain public TestNet scenarios.
+This can only be used in private testing scenarios, as it uses a private faucet to issue network tokens, as well as certain public TestNet scenarios.
```
TheChronicMonster, line 387: This can only be used in private testing scenarios, because it uses a private faucet to issue network tokens. `{!js} fundFromFaucet` may also fund from certain public TestNet scenarios. 

What do you think of this alteration? The current format is a bit long and requires a couple read-throughs. 

I wonder if we have any idea what "Certain public TestNet scenarios" means. Be good to inquire in #help and see if we can get insight to add clarity to this statement.

docs/src/frontend/index.md
```diff
@@ -384,11 +384,20 @@ stdlib.fundFromFaucet(account, balance) => Promise<void>
...
 
+Example:
+
+```js
+load: /examples/js-tests/index.mjs
+md5: 389bd098127db61d413f7586e81b561f
+range: 324 - 325
+```
+
+This code creates an account, and then funds that account with 100000 units from a private faucet.
```
TheChronicMonster, line 398: We now have this beautiful functionality where we can add "_" underscores to numbers that are longer than three digits. :)
I suggest you update this source code with that feature.
In the docs, use a comma. `100,000`

docs/src/frontend/index.md
```diff
@@ -384,11 +384,20 @@ stdlib.fundFromFaucet(account, balance) => Promise<void>
...
+This is used in cases where funding from a normal faucet is difficult or impossible, such as testing large token quantities.
```
TheChronicMonster, line 399: Suggest adding 'when'
such as when testing...

TheChronicMonster (changes requested):

TheChronicMonster (approved):

PR #1115, opened 2022-06-08
improve example for `stdlib`'s `parseInt` function

hamirmahal:


docs/src/frontend/index.md
```diff
@@ -1126,12 +1126,19 @@ Parses a signed `{!rsh} Int` into a JavaScript number.
 Example:
 
 ```js
-load: /examples/programmer-art/src/index.mjs
-md5: 0779cd5bfb7d45de6a7f46b9cd6c7164
-range: 394 - 402
+import { loadStdlib } from "@reach-sh/stdlib";
```
jeapostrophe, line 1129: I think you should add this as a "test-case" to `/examples/js-tests` and then cite the lines

jeapostrophe (changes requested):

examples/js-tests/parseInt/index.mjs
```diff
@@ -0,0 +1,6 @@
+import { loadStdlib } from "@reach-sh/stdlib";
```
jeapostrophe, line 1: You should not make a new file. We want the tests to actually run and `js-tests` already runs

docs/src/frontend/index.md
hamirmahal: This is what `frontend/#js_parseInt` looks like after [062ef7c](https://github.com/reach-sh/reach-lang/pull/1115/commits/062ef7c27079f984589ac0cfc4616119f45e1972).

image

jeapostrophe (changes requested):

docs/src/frontend/index.md
hamirmahal: This is what `frontend/#js_parseInt` looks like after [f16706b](https://github.com/reach-sh/reach-lang/pull/1115/commits/f16706becb3514660854cfe52cb417d072cdc271).

image

jeapostrophe:
Merge after CI

PR #1128, opened 2022-06-09
reachpc

erer1243:
PR to keep track of reachpc development.

Open questions about reachpc:
* Configuration style - I like having a toml file that reachpc reads, but that users can override with environment variables (because my understanding is the equivalent to a .bashrc type thing is using the registry on windows, and I know nothing about that). Jay I think you like the env style we have now.
* Hashing algorithm - this is a more important question for those designing the cloud.

spec which lays out a basic english description of the program https://reachsh.atlassian.net/l/c/aWS190kH


hs/app/reachpc/Main.hs
```diff
@@ -0,0 +1,125 @@
...
+  return ()
+
+execRemote :: [(String, String)] -> (String, [String]) -> IO ()
+execRemote env (cmd, args) = do
+  putStrLn "Executed on remote:"
+  putStrLn $ cmd <> " " <> show args
+  print env
+
+helpMessage :: Text
+helpMessage = [text|
```
jeapostrophe, line 56: Doesn't the CLI parser do this for us?

hs/app/reachpc/Main.hs
```diff
@@ -0,0 +1,125 @@
...
+ where
+  emptyCliOptions = CliOptions ("", []) Nothing Nothing
+  go args opts = case args of
+    ("-h":args')         -> go ("--help" : args') opts
+    ("-e":args')         -> go ("--env"  : args') opts
+    ("--help":_)         -> go ["internal help"] opts
+    ("--env":list:args') -> go args' opts{ co_envVars = Just $ splitOn "," list }
+    (arg@('-':_):_)      -> opts{ co_error = Just $ "Unknown or bad use of argument " <> arg }
+    (cmd:args')          -> opts{ co_command = (cmd, args') }
+    []                   -> opts{ co_error = Just "No command was specified.\nRun with -h to see local commands, or run with the `help` command to see remote commands." }
```
jeapostrophe, line 99: I think what opt-applicative does is show the help one you don't match one of the cases

hs/app/reachpc/Main.hs
```diff
@@ -0,0 +1,125 @@
...
+parseCliOptions :: IO CliOptions
+parseCliOptions = flip go emptyCliOptions <$> getArgs
+ where
+  emptyCliOptions = CliOptions ("", []) Nothing Nothing
+  go args opts = case args of
+    ("-h":args')         -> go ("--help" : args') opts
+    ("-e":args')         -> go ("--env"  : args') opts
+    ("--help":_)         -> go ["internal help"] opts
+    ("--env":list:args') -> go args' opts{ co_envVars = Just $ splitOn "," list }
+    (arg@('-':_):_)      -> opts{ co_error = Just $ "Unknown or bad use of argument " <> arg }
```
jeapostrophe, line 97: I think you need to have more general arguments like `--project` and maybe `--instance`

hs/app/reachpc/Main.hs
```diff
@@ -0,0 +1,125 @@
...
+    (cmd:args')          -> opts{ co_command = (cmd, args') }
+    []                   -> opts{ co_error = Just "No command was specified.\nRun with -h to see local commands, or run with the `help` command to see remote commands." }
+
+version :: IO ()
+version = putStrLn "reachpc version todo"
+
+config :: IO ()
+config = do
+  _localOrCloud <- askUser "Will you use Reach Cloud or Reach Local?" ["cloud", "local"]
+  _connector <- askUser "What connector will you use?" ["algo", "cfx", "eth"]
```
jeapostrophe, line 107: Stylize like docs... capitals!

hs/app/reachpc/ReachPC/ConfigDir.hs
```diff
@@ -0,0 +1,56 @@
...
+inConfigDir :: FilePath -> IO FilePath
+inConfigDir path = (</> path) <$> getXdgDirectory XdgConfig "reach"
+
+configCodec :: T.TomlCodec Config
+configCodec = Config 
+  <$> T.string "cloud-or-local" T..= cfg_cloudOrLocal
+  <*> T.string "connector-mode" T..= cfg_connectorMode
+
+defaultConfigToml :: Text
+defaultConfigToml = [text|
```
jeapostrophe, line 45: Look at how we do file inclusion in the other modules

hs/stack.yaml
```diff
@@ -3,6 +3,8 @@ resolver: lts-19.7
 ghc-options:
   "$everything": -haddock
 
+system-ghc: true
```
jeapostrophe, line 6: What does this do? Does it break my machine?

hs/app/reachpc/ReachPC/DirScan.hs
```diff
@@ -0,0 +1,108 @@
...
+    fileHashes <- mapM hashFile files'
+    let fileHashes' = zip files' (map FH_File fileHashes)
+    let newFilesMap = M.fromList fileHashes'
+
+    -- Add remaining dirs to stack and hashes map
+    let dirStack' = dirs' <> dirStack
+    let newDirsMap = M.fromList (map (, FH_Directory) dirs')
+    
+    let hashTree' = M.unions [hashTree, newFilesMap, newDirsMap]
+    doScan ignores' dirStack' hashTree'
```
jeapostrophe, line 108: This makes it so the ignores in `/a/b/.reachignore` affect `/a/c/` if the stack is `/a/b : /a/c : []`

jeapostrophe (changes requested):

hs/app/reachpc/Main.hs
erer1243: I made this with opt-applicative and it was way longer and harder to follow. I found just writing it out was drastically nicer for such a simple application.

hs/app/reachpc/ReachPC/ConfigDir.hs
erer1243: thanks for the tip

hs/app/reachpc/Main.hs
erer1243: What will those flags take/do?

hs/stack.yaml
erer1243: Accidental commit. fixes stack bug that causes me grief

hs/app/reachpc/ReachPC/DirScan.hs
erer1243: this is not true. 
image

hs/app/reachpc/Main.hs
jeapostrophe: When you talk to the cloud, you need to be able to say what project (organization) you're talking about and eventually what instance it is; check out the REACHp spec

hs/stack.yaml
jeapostrophe: :)

hs/app/reachpc/ReachPC/DirScan.hs
jeapostrophe: If `/a/b/.reachignore` contains `*.js` but `/a/c/index.js` is a file and you're going to visit `/a/c` after you visit `/a/b`, then `ignores'` is going to have `*.js`, right?

jeapostrophe: If it doesn't, then it seems weird, because I WANT `/a/b/xyz/index.js` to be ignored... right? If it doesn't matter, then you're making `ignores'` get bigger and bigger and bigger rather than scoping it. I basically don't get why this is a queue rather than being a tree walk

hs/app/reachpc/Main.hs
erer1243: Oops I misread. I thought I based that line on what reach does already

hs/app/reachpc/ReachPC/DirScan.hs
erer1243: The filesystem is already inherently scoped. It's less efficient in that it in a big project it will check a lot of impossible patterns, but I think it's safe to assume it will be super rare for that to be noticable.

hs/app/reachpc/ReachPC/CommandLine.hs
```diff
@@ -0,0 +1,65 @@
...
+  go args opts = case args of
+    ("-h":args')            -> go ("--help" : args') opts
+    ("-e":args')            -> go ("--env"  : args') opts
+    ("--help":_)            -> go ["local-help"] opts
+    ("--env":list:args')    -> go args' opts{ cli_forwardedEnvVars = Just $ splitOn "," list }
+    ("--cloud":args')       -> go args' opts{ cli_cloudOrLocal = Just True }
+    ("--local":args')       -> go args' opts{ cli_cloudOrLocal = Just False }
+    ("--connector":c:args') -> go args' opts{ cli_connector = Just c }
+    ("--project":p:args')   -> go args' opts{ cli_project = Just p }
+    ("--organization":o:args') -> go args' opts{ cli_organization = Just o }
```
jeapostrophe, line 62: I think that `--org` should be a short-hand. But also, I think that when you specify a project like `@xbacked/staking`, then the `@xbacked` means the organization. I can see how there might be commands that need to know an org without a project, but I think that most of the time `--project` will be only thing used

hs/app/reachpc/ReachPC/CommandLine.hs
```diff
@@ -0,0 +1,65 @@
...
+parseCliOptions = flip go emptyCliOptions <$> getArgs
+ where
+  go args opts = case args of
+    ("-h":args')            -> go ("--help" : args') opts
+    ("-e":args')            -> go ("--env"  : args') opts
+    ("--help":_)            -> go ["local-help"] opts
+    ("--env":list:args')    -> go args' opts{ cli_forwardedEnvVars = Just $ splitOn "," list }
+    ("--cloud":args')       -> go args' opts{ cli_cloudOrLocal = Just True }
+    ("--local":args')       -> go args' opts{ cli_cloudOrLocal = Just False }
+    ("--connector":c:args') -> go args' opts{ cli_connector = Just c }
```
jeapostrophe, line 60: This is always passed in the env nowadays

hs/app/reachpc/ReachPC/CommandLine.hs
```diff
@@ -0,0 +1,65 @@
...
+emptyCliOptions = CliOptions ("", []) Nothing Nothing Nothing Nothing Nothing Nothing
+
+parseCliOptions :: IO CliOptions
+parseCliOptions = flip go emptyCliOptions <$> getArgs
+ where
+  go args opts = case args of
+    ("-h":args')            -> go ("--help" : args') opts
+    ("-e":args')            -> go ("--env"  : args') opts
+    ("--help":_)            -> go ["local-help"] opts
+    ("--env":list:args')    -> go args' opts{ cli_forwardedEnvVars = Just $ splitOn "," list }
```
jeapostrophe, line 57: Docker always you to write `docker -e X -e Y` but you make it so only the last one counts

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+                    env_project <|> cli_project <|> prj_project <|> glb_project
+
+  -- Remote organization
+  -- REACH_ORGANIZATION / --organization=... / organization = ".."
+  let env_organization = envVar "REACH_ORGANIZATION"
+  let cli_organization = Cli.cli_organization cliOpts
+  let prj_organization = rtml_organization projectRchToml
+  let glb_organization = rtml_organization globalRchToml
+  let cfg_organization = unwrapConfigOption "Unspecified remote organization." $
+                         env_organization <|> cli_organization <|> prj_organization <|> glb_organization
```
jeapostrophe, line 120: Abstraction... but also remember my comment about the org coming from the project

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+
+  print Config{..}
+
+  return Config{..}
+ where
+  boolToCloudOrLocal b = if b then Cloud else Local
+  unwrapConfigOption = fromMaybe . error . (++ "\nCheck <reach.sh/docs/project-config TODO page> for help\n")
```
jeapostrophe, line 127: I think these should be lets

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+-- instead of throwing an error and printing valid options. This can probably be fixed with T.match
+reachTomlCodec :: T.TomlCodec RchToml
+reachTomlCodec = RchToml
+  <$> T.dioptional (T.enumBounded "cloud-or-local") T..= rtml_cloudOrLocal
+  <*> T.dioptional (T.enumBounded "connector") T..= rtml_connector
+  <*> T.dioptional (T.arrayOf T._String "forward-env-vars") T..= rtml_forwardEnvVars
+  <*> T.dioptional (T.string "project") T..= rtml_project
+  <*> T.dioptional (T.string "organization") T..= rtml_organization
+
+-- TODO: put into separate file and include at compile time
```
jeapostrophe, line 168: You did that

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+                         env_cloudOrLocal <|> cli_cloudOrLocal <|> prj_cloudOrLocal <|> glb_cloudOrLocal
+
+  -- What connector to use (algo/cfx/eth/...)
+  -- REACH_CONNECTOR / --connector=... / connector = "algo" or "eth" or ...
+  let env_connector = readConnector <$> envVar "REACH_CONNECTOR"
+  let cli_connector = readConnector <$> Cli.cli_connector cliOpts
+  let prj_connector = rtml_connector projectRchToml
+  let glb_connector = rtml_connector globalRchToml
+  let cfg_connector = unwrapConfigOption "Unspecified what connector to use." $
+                      env_connector <|> cli_connector <|> prj_connector <|> glb_connector
```
jeapostrophe, line 94: Abstraction

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+  -- env var, command line flag, project reach.toml, global reach.toml.
+  
+  -- Whether to use reachd-cloud or reachd-local
+  -- REACH_CLOUD_OR_LOCAL / --cloud or --local / cloud-or-local = "cloud" or "local"
+  let env_cloudOrLocal = readCloudOrLocal <$> envVar "REACH_CLOUD_OR_LOCAL"
+  let cli_cloudOrLocal = boolToCloudOrLocal <$> Cli.cli_cloudOrLocal cliOpts
+  let prj_cloudOrLocal = rtml_cloudOrLocal projectRchToml
+  let glb_cloudOrLocal = rtml_cloudOrLocal globalRchToml
+  let cfg_cloudOrLocal = unwrapConfigOption "Unspecified whether to use Reach Cloud or Reach Local." $
+                         env_cloudOrLocal <|> cli_cloudOrLocal <|> prj_cloudOrLocal <|> glb_cloudOrLocal
```
jeapostrophe, line 85: Abstraction

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -0,0 +1,170 @@
...
+  , cfg_connector :: Connector
+  , cfg_forwardEnvVars :: [String]
+  , cfg_project :: String
+  , cfg_organization :: String
+  } deriving (Show)
+
+-- Represents a reach.toml file; Same fields as Config, except every field is optional (because they
+-- may also be specified by flags or env vars).
+data RchToml = RchToml
+  { rtml_cloudOrLocal :: !(Maybe CloudOrLocal)
```
jeapostrophe, line 59: Don't we use StrictData?

jeapostrophe (changes requested):

hs/app/reachpc/ReachPC/CommandLine.hs
erer1243: I don't think it hurts to have this flag. at the very least it's more ergonomic to us when e.g. testing an example on every connector

hs/app/reachpc/ReachPC/CommandLine.hs
erer1243: Thinking on it further, I'm not sure it even makes sense to have a project flag. Since a reach.toml file is required to recognize a project locally, it may as well just require the project be specified in it..

erer1243: I'll make these changes in the dedicated config pr

hs/app/reachpc/ReachPC/CommandLine.hs
jeapostrophe: It hurts because now there is inconsistency between the tools and you're elevating this particular environment variable above all the others. If I do `RECH_CONNECTOR_MODE=ALGO reach --connector ETH run` what environment variable does the stdlib see? Which connector gets loaded?

hs/app/reachpc/ReachPC/Config.hs
jeapostrophe: !

erer1243:
For some reason I can't reply directly to your comments.
About the connector flag, my understanding was that reachpc should be entirely in charge of the connector, since we want it to be configurable in a toml config file. And reachpc would pass it via a 'special' channel to reachd rather than just forwarding an env var. Even if we don't want this, the intended behavior is for the environment variable to take complete precedent anyway.

! that file is changed a lot in the general config pr

jeapostrophe:
> For some reason I can't reply directly to your comments. About the connector flag, my understanding was that reachpc should be entirely in charge of the connector, since we want it to be configurable in a toml config file. 

No, reachpc is merely a conduit for information, it is not in charge of the connector. The config file should facilitate setting configuration for Reach, so I see its job as making it is easy to have `REACH_CONNECTOR_MODE=ALGO` in a file and not having to remember to `source` it

> And reachpc would pass it via a 'special' channel to reachd rather than just forwarding an env var. Even if we don't want this, the intended behavior is for the environment variable to take complete precedent anyway.

I don't believe we want it

erer1243:
I changed REACH_CONNECTOR to REACH_CONNECTOR_MODE in the config branch. I believe all three reachpc related branches (the current branch, [this](https://github.com/reach-sh/reach-lang/pull/1186), and [this](https://github.com/reach-sh/reach-lang/pull/1191)) are ready to merge. Not sure if it's better to merge the two small ones into this branch first and then do a mega single master merge, or to merge all three separately into master. Either works for me when I tried it locally.

jeapostrophe:
I think you should merge everything into one branch and then into master

erer1243:
I am kind of sure I rebased the branches correctly. I did: rebase reachpc onto master, rebase reachpc-configuration onto reachpc, backup reachpc-core-1926, rebase reachpc-core-1926 onto reachpc-configuration, rename reachpc to reachpc-original, rename reachpc-core-1926 to reachpc, then force push that replacing origin/reachpc. It looks right to me.. I tried using the rebase squash functionality but it did not simplify any rebase fix; I had to fix conflicts 10 to 15 times during that, and it came out wrong. So I just tried it with -X ours and no squashing.

jeapostrophe (approved):

PR #1135, opened 2022-06-13
Dct 287 refresh

Craig-Elre:
Wisdom for Sale Tutorial


jeapostrophe:
@TheChronicMonster @Craig-Elre Please review my changes for the next time you do something like this

PR #1155, opened 2022-06-17
CORE-1906: REACHp PoC with stubbed server + client

mattaudesse:


hs/src/Reach/Proto.hs
```diff
@@ -0,0 +1,190 @@
...
+
+type V0_Stream = "v0"
+  :> "proc" :> Capture "pid" Pid
+  :> "fd"   :> Capture "fd"  FDescOut'
+  :> StreamGet NoFraming EventStream (SourceIO FDescOut)
+
+type Proto = V0_Sync :<|> V0_Stream
+
+runStubServer :: IO ()
+runStubServer = Warp.run stubPort $ serve (Proxy @Proto) ((r :<|> pin) :<|> pout) where
```
jeapostrophe, line 141: Do we imagine that is a demonstration of how to write a server or that this is the library that will actually do it? If library, then I think this should be a bit more parameterized, like `runServer :: MonadIO m => Config -> (Req -> m Resp) -> m ()` for a suitable part of `Req`/`Resp`

jeapostrophe (changes requested):
Can you add a little test script w/ sample output? I'm having a hard time understanding what this is :)

mattaudesse:
> Can you add a little test script w/ sample output? I'm having a hard time understanding what this is :)

I made these Makefile targets in lieu of a single script:
```Makefile
.PHONY: proto-stub-serve
proto-stub-serve: hs-build
        @stack exec -- proto-stub serve

.PHONY: proto-stub-client
proto-stub-client: hs-build
        @stack exec -- proto-stub cmd help
        @stack exec -- proto-stub cmd compile --verify-fail-once
        @stack exec -- proto-stub cmd run examples/ttt foo bar baz
        @stack exec -- proto-stub say
        @stack exec -- proto-stub listen
```

**Sample client run:**
[asciicast](https://asciinema.org/a/gJYQXxAMXWqSpxErfr4t5iKWA)

**Sample stub server run:**
[asciicast](https://asciinema.org/a/VSD0JuYnirc7nHmFkSQDzDsIu)

hs/src/Reach/Proto.hs
mattaudesse: I think of this function as more of a demo, but other elements of this module are the beginnings of a library.

In real-world servers it'd be more common to break stuff in `runStubServer` out into abstracted + distinct pieces, so it naturally becomes easier to parameterize.

I think `runServer :: Config -> ...` or `c <- readConfig` will be a common pattern for all our REACHp-speakers though, yup. I agree we could probably make a helper function to capture the pattern.

jeapostrophe (approved):

hs/src/Reach/Proto.hs
```diff
@@ -0,0 +1,189 @@
...
+ <> "event: " <> e <> "\n"
+ <> "data: " <> m <> "\n\n"
+
+instance ToSSE FDescOut where
+  toSSE = \case
+    Stdout n m -> toSSE' "stdout" n m
+    Stderr n m -> toSSE' "stderr" n m
+    Keepalive  -> ": keepalive\n\n"
+
+instance ToSSE a => MimeRender EventStream a where
```
DanBurton, line 106: We have `FlexibleInstances` on by default already? I don't feel strongly enough about it to object, but it makes me suspicious to see instances that are not defined on concrete types.

DanBurton (commented):
This PR takes the code base from 0 servant servers to 1 servant server and that server even responds to requests resembling what we actually want it to do eventually. This is exactly the kind of small step that I am thinking of when I say "small PRs = good".

hs/src/Reach/Proto.hs
mattaudesse: https://github.com/reach-sh/reach-lang/blob/b9c95e61ab571309b4c1a3616a3d5e0c82e7a0b7/hs/package.mo.yaml#L109

I don't really have an opinion so if you'd like to see this changed I think we can.

PR #1172, opened 2022-06-20
move Discussions logo to left instead of right

hamirmahal:
### Without this change
image

### With this change
image


jeapostrophe:
I was hoping it would be even more left :)

PR #1178, opened 2022-06-22
adopt mo

erer1243:
I think we should adopt mo. I believe the license permits this and it simplifies our scripts. mo license: https://github.com/tests-always-included/mo/blob/master/LICENSE.md

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -22,4 +22,3 @@ DOTNET_SDK_IMAGE=mcr.microsoft.com/dotnet/sdk:${DOTNET_SDK_VERSION}
 DOTNET_RUNTIME_IMAGE=mcr.microsoft.com/dotnet/runtime:${DOTNET_SDK_VERSION}
 PYTHON_IMAGE_VERSION=3.8-alpine3.15
 PYTHON_IMAGE=python:${PYTHON_IMAGE_VERSION}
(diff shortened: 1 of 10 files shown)


jeapostrophe:
I think it is a good idea. Why close?

erer1243:
It doesn't actually really reduce the total complexity of the project. If you want, I can still finish it.

DEPS
```diff
@@ -22,4 +22,3 @@ DOTNET_SDK_IMAGE=mcr.microsoft.com/dotnet/sdk:${DOTNET_SDK_VERSION}
 DOTNET_RUNTIME_IMAGE=mcr.microsoft.com/dotnet/runtime:${DOTNET_SDK_VERSION}
 PYTHON_IMAGE_VERSION=3.8-alpine3.15
 PYTHON_IMAGE=python:${PYTHON_IMAGE_VERSION}
-MO_URL=https://raw.githubusercontent.com/tests-always-included/mo/master/mo
```
jeapostrophe, line 25: I think it could be valuable to not download `mo` and instead `cp` it, but we'd still need all the `install-mo` things to get them into the Docker image and it would be a little bit MORE complex, because we need to get them into the right place in the filesystem. It might be possible to symlink all the uses of `mo` and then when the Docker build bundle is made, it will resolve to the real thing. If you want, you can make a card for that or just do it

PR #1186, opened 2022-06-23
CORE-1891 Config subcommand & remove default global reach.toml creation

erer1243:
A pretty, interactive `reach config` subcommand. Also commented the automatic creation of a global reach.toml. I'm not sure anymore if that makes sense to do. I'd like to spend some time looking at language tools I like for their user friendliness and investigate how they are configured on a project/global level, and then copy that behavior on this branch. Also would like opinions.


hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -63,12 +72,92 @@ data RchToml = RchToml
...
   } deriving (Show)
 
+emptyRchToml :: RchToml
+emptyRchToml = RchToml Nothing Nothing Nothing Nothing Nothing
+
+-- helper for interactiveConfigCreate
+data PromptAnswer = Unset | Unchanged | Str String
+
+-- Code for the `reach config` subcommand
+interactiveConfigCreate :: IO ()
```
jeapostrophe, line 82: I think it is better to have this code in Main. If we are to have two files (blech), then I think it makes sense for there to be one "data" file and the other "UI" file

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -63,12 +72,92 @@ data RchToml = RchToml
...
+  connector <- askUserEnum' ("\nWhat connector do you want to use? " <> showCurrent rtml_connector) (showEnum @Connector)
+               <&> \case
+                 Unset -> Nothing
+                 Unchanged -> rtml_connector rchTml
+                 Str s -> Just $ readConnector s
+  forwardEnvVars <- askUserString ("\nWhat environment variables do you want to forward? " <> showCurrent rtml_cloudOrLocal) "Env vars (comma separated list, e.g. VAR1,VAR2,VAR3): "
+                    <&> \case
+                      Unset -> Nothing
+                      Unchanged -> rtml_forwardEnvVars rchTml
+                      Str s -> Just . map (dropWhileEnd isSpace . dropWhile isSpace) . splitOn "," $ s
```
jeapostrophe, line 115: This code feels like you're calling a function a few times

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -63,12 +72,92 @@ data RchToml = RchToml
...
+  let newRchTml = RchToml
+        { rtml_project = project
+        , rtml_organization = organization
+        , rtml_cloudOrLocal = cloudOrLocal
+        , rtml_connector = connector
+        , rtml_forwardEnvVars = forwardEnvVars
+        }
+
+
```
jeapostrophe, line 124: Yuck

jeapostrophe:
I don't feel strongly about creating a global file by default

hs/app/reachpc/ReachPC/Config.hs
erer1243: how did you even see the separation? good eye

erer1243:
I remade the a bunch of the config module. The old way didn't really make sense. Allowing a user to e.g. set a global default project, or temporarily rename their project from the command line; I figure users would be smart enough to not do that, but for those who do get confused, it would probably lead to a bad first experience with reach.

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -1,27 +1,29 @@
-{- HLINT ignore "Use newtype instead of data" -}
-{-# LANGUAGE TemplateHaskell #-}
+{- HLINT ignore "Use if" -}
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
jeapostrophe, line 2: Why do you need this? It has a scary name. @DanBurton ?

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -42,113 +44,207 @@ readEnum :: forall a. (Show a, Read a, Enum a, Bounded a) => String -> String ->
...
-  <*> T.dioptional (T.arrayOf T._String "forward-env-vars") T..= rtml_forwardEnvVars
-  <*> T.dioptional (T.string "project") T..= rtml_project
-
--- TODO: put into separate file and include at compile time
-defaultReachToml :: ByteString
-defaultReachToml = $(makeRelativeToProject "app/reachpc/ReachPC/default_reach.toml" >>= embedFile)
+    if exists then return $ Just path else go paths
+
+globalsTomlPath :: IO FilePath
+globalsTomlPath = getXdgDirectory XdgConfig "reach" <&> (</> "global.toml")
```
jeapostrophe, line 149: I think it should be `config.toml` not "global"... it's global because of where it is, not its name, right?

hs/app/reachpc/ReachPC/Config.hs
```diff
@@ -42,113 +44,207 @@ readEnum :: forall a. (Show a, Read a, Enum a, Bounded a) => String -> String ->
...
+        }
+  path <- if alreadyExists then return oldTomlPath else getCurrentDirectory <&> (</> "reach.toml")
+  writeToml path reachTomlCodec newReachToml
+
+interactiveCreateGlobalsToml :: IO ()
+interactiveCreateGlobalsToml = do
+  oldToml <- readOrEmptyGlobalsToml
+  let showOld field = " Currently " <> maybe "unset" (\f -> "\"" <> show f <> "\"") (field oldToml)
+  let parseAnswer' field = parseAnswer $ field oldToml
+
```
jeapostrophe, line 197: I hate extra newlines... I can't keep track of when the function ends

hs/app/reachpc/ReachPC/Config.hs
erer1243: It allows `showEnum`. The compiler suggested it. I don't 100% understand it but it doesn't seem bad.

erer1243: https://stackoverflow.com/questions/45128492/what-is-allowambiguoustypes-and-why-is-it-needed-in-this-forall-example somewhat explains it for me

hs/app/reachpc/ReachPC/Config.hs
erer1243: true

hs/app/reachpc/ReachPC/Config.hs
erer1243: Hm, I hate dense code for the same type of reason. My eyes and mind can't focus on it well. But if you want, I can remove the spacing.

hs/app/reachpc/ReachPC/Config.hs
jeapostrophe: `showEnum` feels very powerful and a general tool separate from `ReachPC.Config`'s purview

erer1243: I don't think it's that crazy. It just makes a list of all the variants of the type you give it. Handily, when deriving all the needed instances, it gives you `Read`-able strings that can be substituted for user input. It just saves a few lines. I was inspired by Tomland.enumBounded. I don't think `AllowAmbiguousTypes` is that crazy either, my understanding is that ghc doesn't like inferring a type from an already inferred type. The flag makes that type inference lazier. It doesn't change program safety.

erer1243: If you want I can just make it take a Proxy or make them literals.

jeapostrophe: I want @DanBurton 's opinion on adding the extension (at the whole project level or just here) or doing the proxy

hs/app/reachpc/ReachPC/Config.hs
jeapostrophe: :cry:

hs/app/reachpc/ReachPC/Config.hs
DanBurton: I don't like it but I don't hate it. Seems harmless but feels icky. But having to write Proxy is ickier. Let's keep it enabled only at the module level rather than whole project. Maybe I will become desensitized to it over time and be ok with it being enabled for the whole project.

jeapostrophe: Ok

jeapostrophe (approved):

jeapostrophe:
Merge conflicts :'(

erer1243:
this was merged through the reachpc pr

PR #1191, opened 2022-06-24
reachpc CORE-1926 cache file hashes

erer1243:


hs/app/reachpc/ReachPC/DirScan.hs
```diff
@@ -55,54 +68,85 @@ compareHashTrees localHashes remoteHashes = M.union updatedFiles deletedFiles
...
+        let ignores = map (G.compile . (dir </>)) ignoresPatterns
+        return ignores
+
+  hashTreeToCacheTree = M.fromList . foldr foldHelper [] . M.toList
+  foldHelper (p, fh) lst =
+    case fh of
+      FH_File h -> (p, h) : lst
+      _ -> lst
+
+  cachePath = root </> ".reachcache"
```
jeapostrophe, line 143: We already have `.reach` directory... I think it would be best to use it rather than have too many `.reach` files

jeapostrophe (changes requested):

hs/app/reachpc/ReachPC/ConfigDir.hs
```diff
@@ -0,0 +1,56 @@
...
+# This config file contains default settings for Reach
+# They can be overridden with environment variables
+
+# Whether to use Reach Cloud or Reach Local. Can be "cloud" or "local"
+# Can be overridden by REACH_CLOUD_OR_LOCAL
+cloud-or-local = "cloud"
+
+# Which connector to use. Can be "algo" or "cfg" or "eth"
+# Can be overridden by REACH_CONNECTOR_MODE
+connector-mode = "algo"
```
jeapostrophe, line 55: It should also be capitalized

jeapostrophe (changes requested):

PR #1189, opened 2022-06-23
CORE-1985/1986/1997 REACHp: extend streaming API

mattaudesse:
Introduces new `withKeepalives` and `stubUnixProc` functions (the latter being for the purpose of demonstration), and terminates streams with process' exit code.

[asciicast](https://asciinema.org/a/KZ0TCUmNbiXqrDPuKkmkJ0Z7D)

[asciicast](https://asciinema.org/a/yW8b8jLW1c0oHCK6A0GrNPFLX)

PR #1193, opened 2022-06-24
CORE-1916/REACHp: create initial test harness

mattaudesse:
~Dependent upon #1189; if that gets merged to `master` I'll rebase this and flip it out of "draft" mode.~

```
$ make watch-proto-test
Requesting `reach compile`
  should return a `Res` (which we'll flesh out later)
`say`ing some freeform text to a `PID`
  should succeed

Finished in 0.5190 seconds
2 examples, 0 failures
```

The prospect of building up a test suite like this raises an obvious question: how will you account for the tight-coupling of runtime concerns like databases and proxies and remote containers, etc?

My hand-wavy answer is that orienting work around tests like this _early_ naturally lends the product to mocking + simpler abstraction in the style of "functional core; imperative shell". If we design mainly in terms of nouns and datatypes instead of verbs and opaque side-effects I think we'll better understand our code's innards, and using spec tests like this encourages the discipline.

This PR is deliberately small lest I'm the only one who thinks it's a good idea :-)

### DIFF
--- a/hs/Makefile
+++ b/hs/Makefile
@@ -154,6 +154,10 @@ watch-tsa: expand
 watch-proto-stub: expand
 	@$(call w,exe:proto-stub) -T Main.main
 
+.PHONY: watch-proto-test
+watch-proto-test: expand
+	@$(call w,test:reach-test) -T Reach.Test_Proto.main
+
 .PHONY: repl-reach-cli
 repl-reach-cli: expand
 	@stack ghci app/reach/Main.hs $(MEVEREST) \
--- a/hs/package.mo.yaml
+++ b/hs/package.mo.yaml
@@ -35,6 +35,7 @@ dependencies:
 - hexstring
 - hspec
 - hspec-smallcheck
+- hspec-wai
 - http-client
 - http-client-tls
 - http-conduit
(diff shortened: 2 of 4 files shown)


jeapostrophe (commented):
It's hard to tell what is different. You can have PRs onto non-`master` branches.

As for how to do tests like this longer term... I think something like this is fine for "micro" tests of the protocol handling, but I think the real proof will be in the pudding and we'll have just end-to-end tests of e.g. actually running `reach run` on something

mattaudesse:
> As for how to do tests like this longer term... I think something like this is fine for "micro" tests of the protocol handling, but I think the real proof will be in the pudding and we'll have just end-to-end tests of e.g. actually running `reach run` on something

Yeah, I think they compliment each other and I wouldn't feel like we'd finished the job without the kind of broader end-to-end tests you describe. Would you rather us not fuss with micro?

jeapostrophe (approved):

jeapostrophe:
> Would you rather us not fuss with micro?

Just don't try too hard :)

mattaudesse:
> > Would you rather us not fuss with micro?
> 
> Just don't try too hard :)

Haha - you got it.

PR #1197, opened 2022-06-24
DCT-375

Craig-Elre:
Remove a couple of existing words in the description, and add a code sample and description for `balanceOf`

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -535,7 +535,7 @@ md5: 69477201beedc87b2616001290ae29cf
 range: 10-10
 ```
 
-This code obtains the balance of network tokens for user, `{!js} who`, and sets it to the `getBalance` object.
+This code obtains the balance of network tokens for a user, `{!js} who`, and sets it to the `getBalance` object.
 This makes the account balance simpler to reference later in the code, such as displaying balances after transactions, proving that cancelled transactions did not transfer tokens, etc.
 
 ---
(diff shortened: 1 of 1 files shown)


docs/src/frontend/index.md
```diff
@@ -525,7 +525,18 @@ acc.balanceOf(token?) => Promise<BigNumber>
...
+
+Example:
+
+```js
+load: examples/raffle/index.mjs
+md5: 69477201beedc87b2616001290ae29cf
+range: 10-10
+```
+
+This code obtains the balance of network tokens for user, `{!js} who`, and sets it to the `getBalance` object.
```
hamirmahal, line 538:
```suggestion
This code obtains the balance of network tokens for a user, `{!js} who`, and sets it to the `getBalance` object.
```

docs/src/frontend/index.md
```diff
@@ -525,7 +525,18 @@ acc.balanceOf(token?) => Promise<BigNumber>
...
+This makes the account balance simpler to reference later in the code, such as displaying balances after transactions, proving that cancelled transactions did not transfer tokens, etc.
```
hamirmahal, line 539:
```suggestion
This makes the account balance simpler to reference later in the code, by doing things like displaying balances after transactions, proving that cancelled transactions did not transfer tokens, etc.
```

Craig-Elre: It doesn't actually do these things, just makes it simpler to do these things. I do not agree with this change.

jeapostrophe (approved):

PR #1210, opened 2022-06-28
exponential form - potential fix

chikeabuah:
note: this is currently restricted to the form of scientific notation which base Haskell supports 

```
ghci> read "-1.E34" :: Double
*** Exception: Prelude.read: no parse
ghci> read "10e+9" :: Double
1.0e10
```


hs/src/Reach/Eval/Core.hs
```diff
@@ -4264,9 +4265,11 @@ evalExpr e = case e of
     locAtf (srcloc_jsa "id ref" a) $
       evalId "expression" x
   JSDecimal a ns -> do
-    when ('e' `elem` ns || 'E' `elem` ns) $
-      expect_ $ Err_Eval_IllegalJS e
-    case splitOn "." ns of
+    let ns' = case ('e' `elem` ns || 'E' `elem` ns) of
+            True -> do
+              showFFloat Nothing (read ns :: Double) ""
```
jeapostrophe, line 4270: Cute. You need some tests though and to review the CI logs

jeapostrophe: And to update the docs... it says we don't support it

And make sure you update the changelog to say that we do now too

chikeabuah:
https://reachsh.slack.com/archives/CMY3E5B3J/p1656521443829079

DanBurton:
According to this: https://app.circleci.com/pipelines/github/reach-sh/reach-lang/7926/workflows/0cb4bcbf-6e78-4635-b63c-ccb9756bfbe0/jobs/142416

There's only one failure, which is hs-test. (Look at the "echo $RECORD_MESSAGE" step to see what it sent to slack)

And according to the link it provides: https://app.circleci.com/pipelines/github/reach-sh/reach-lang/7926/workflows/0cb4bcbf-6e78-4635-b63c-ccb9756bfbe0/jobs/142405

that failure is:

```
Test output was different from '/build/t/y/./scientific_notation.txt'. Output of ["diff","-u","/build/t/y/./scientific_notation.txt","/tmp/scientific_notation222-52.actual"]:
```

The diff looks trivial. It will pass if you just update the `scientific_notation.txt` file.

chikeabuah:
thank you @DanBurton

examples/new-contract1/build/child.main.mjs
```diff
@@ -2,7 +2,7 @@
 /* eslint-disable */
```
jeapostrophe, line 2: You should run `./scripts/restore-build.sh` when you accidentally change these files in meaningless ways

hs/t/n/ne1.rsh
```diff
@@ -1,2 +0,0 @@
-'reach 0.1';
```
jeapostrophe, line 1: Don't delete it... move it to hs/t/y

hs/t/y/scientific_notation.rsh
```diff
@@ -0,0 +1,11 @@
+'reach 0.1';
+
+export const main =
+  Reach.App(
+    {},
+    [Participant('A', {})],
+    (A) => {
+      A.only(() => {
+        const x = 10e+2;
```
jeapostrophe, line 9: You should add more examples of the things you expect to allow. Maybe include the exact example for ne1. Also, when you write new Reach examples, you should use the modern non-ternary style

jeapostrophe (changes requested):

chikeabuah:
> You should remove the build file changes, make it work with JS syntax that we can conceivably support, and add more tests, including that the values are parsed correctly.

i reverted the build changes 

i don't know what you mean by "JS syntax that we can conceivably support"

there are already tests that parse the values correctly. what other tests would you like to see?

chikeabuah:
@DanBurton any input on what Jay means above?

DanBurton:
re: "conceivably support": we could conceivably support any UInt literal that evaluates to a whole number.

Yes:

```
> 1.e2
100
> 1.1e2
110
```

No:

```
> 1.1e0
1.1
> 1e-1 // this is currently allowed but is bad
0.1
```

jeapostrophe:
We can support those examples Dan, because we already turn 0.1 into a fixed point number

DanBurton:
To amend my earlier suggestion in light of what Jay said about fixed point numbers, I think we should parse any valid scientific notation expression:

* as a fixed point number if it has a decimal
* else as a fixed point number if it has no decimal but has a negative exponent
* else as an Int if it is a negative (?)
* else as a UInt (in this case it is none of the above, which makes it a nonnegative whole number)

jeapostrophe:
Right on, I think we can basically support anything.

DanBurton:
superceded by https://github.com/reach-sh/reach-lang/pull/1326

PR #1216, opened 2022-06-28
set version to 0 in package.mo.yaml

willghatch:
Using version 0 in package.yaml seems fine as far as I can tell.  `reach --version` still gives the right answer and `build/index.main.mjs` files still have the right version inside.  As long as CI passes, I think this should be fine.

Since version is the only variable we substitute in, should we stop using `mo` on the `package.yaml` file completely?  We still need some generation to optionally include `package.closed.yaml`, which the makefile currently handles poorly (it doesn't check it for updates -- you need to touch `packages.mo.yaml` or something to get it to build properly).  I'm not sure what actually triggers new docker images or whatever it was we wanted to avoid by not changing the version number.

### DIFF
--- a/hs/Dockerfile.haskell-build-artifacts
+++ b/hs/Dockerfile.haskell-build-artifacts
@@ -42,7 +42,7 @@ RUN apk add make curl bash
 COPY .docker-root ./.docker-root
 COPY Makefile.docker ./Makefile
 RUN make ROOT=.docker-root install-mo
-COPY package.mo.yaml ./
+COPY package.open.yaml ./
 RUN make ROOT=.docker-root package.yaml
 
 COPY stack.yaml stack.yaml.lock ./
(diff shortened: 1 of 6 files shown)


willghatch:
The only test that failed is a test that checks that the version in Version.hs matches the cabal version (generated by `package.mo.yaml`).  So... I think I should just delete that test and then this is merge-able, though maybe we also want to edit the makefile.

jeapostrophe:
I agree with your conclusion. Re Makefile... if you can, go for it.

willghatch:
This now has the version fix, but also the second commit has a change to the makefile to properly but optionally depend on the `package.closed.yaml` file.  I believe I did it in a portable way (IE a way that should work on non-GNU make, like on MacOS or FreeBSD).  But I'm not 100% certain.

jeapostrophe (approved):

jeapostrophe:
Excellent

PR #1218, opened 2022-06-29
CORE-2011/REACHp: propose response types + dispatch/interpreter

mattaudesse:
This PR offers a sketch of how we might approach:
- [An open spec question](https://github.com/reach-sh/reach-everest/pull/5#discussion_r899189655)
- Keeping `reachpc`'s knowledge minimal
- Maintaining our current CLI features

Basically: boring old dispatch/interpret patterns. This isn't intended to be complete but I think we could use it as a foundation to continue building upon if there's rough consensus about the design.

*`Res` is short for "response"*
```haskell
data ResInterpret
  = ExitStdout Int Text
  | ExitStderr Int Text
  | AttachStreamJustListen Pid Integer
  | AttachStreamListenSpeak Pid Integer
  -- TODO spec further:
  --  * We'll want to batch PUTs rather than send them all at once
  --  * Once the cloud is satisfied it has everything we'll redirect the client
  | PutProjectUpdates ProjectContent
  deriving (Eq, Show, Generic, FromJSON, ToJSON)

data Res
  = Redirect String
  | Interpret ResInterpret
  deriving (Eq, Show, Generic, FromJSON, ToJSON)
```

`PutProjectUpdates` is how a cloud service might ask `reachpc` to push updated project modules up over the wire. The client may need to do this before receiving an `AttachStreamJustListen`, for instance.

[asciicast](https://asciinema.org/a/0sz2fuf5GrgEHzsMVn99Yiy8y)
[asciicast](https://asciinema.org/a/16q0HfwdKSRpjDC3NCa8QMqhu)

### DIFF
--- a/hs/test/Reach/Test_Proto.hs
+++ b/hs/test/Reach/Test_Proto.hs
@@ -22,11 +22,16 @@ import qualified Servant.Client.Streaming as S
 
 main :: IO ()
 main = hspec $ do
+  spec_account
   spec_headerParse
   spec_proto
 
 type MXP = Maybe XProject
 type EXP = Either String XProject
+
+spec_account :: Spec
+spec_account = describe "Parsing an `Account`" $ do
+  it "<has yet to be specified>" $ pendingWith "CORE-2029"
 
 spec_headerParse :: Spec
 spec_headerParse = describe "Parsing an" $ do
--- a/hs/test/Reach/Test_Proto.hs
+++ b/hs/test/Reach/Test_Proto.hs
@@ -1,43 +1,80 @@
 module Reach.Test_Proto
-  ( spec_proto
+  ( spec_headerParse
+  , spec_proto
   , main
   ) where
 
+import Control.Monad.Except
+import Data.Aeson
 import Network.HTTP.Client
 import Reach.Proto
 import Safe
 import Servant.API
 import Servant.Client
+import Servant.Types.SourceT
 import Test.Hspec
+import qualified Data.Map.Strict as M
 import qualified Data.Text as T
 import qualified Network.Wai.Handler.Warp as Warp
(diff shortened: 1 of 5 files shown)


hs/test/Reach/Test_Proto.hs
```diff
@@ -16,15 +18,26 @@ main :: IO ()
...
+spec_proto = around (Warp.testWithApplication app) $ do
   u <- runIO $ parseBaseUrl "http://localhost"
   m <- runIO $ newManager defaultManagerSettings
   let env p = mkClientEnv m $ u { baseUrlPort = p }
+  let run p x = runClientM x (env p) >>= either (fail . show) (pure . id)
 
   describe "Requesting `reach compile`" $ do
-    it "should return a `Res` (which we'll flesh out later)" $ \p ->
-      runClientM (cmd'' "compile" $ Req [] Nothing mempty) (env p) >>= (`shouldBe` Right Res)
+    it "with a bogus flag like --nope will fail" $ \p -> do
```
mattaudesse, line 28: (BTW, I don't think `--flag` specs really belong in Test_Proto.hs long-term... these just test my temporary stubs)

jeapostrophe (commented):
I think I am stupid and don't really understand what is different about this... and how the code relates to the things that you commented on in the message.The main thing is that there can be five different kinds of output, plus a redirect? How does the upload response get a second response after the files are uploaded? I feel like it would be simpler to just have a single kind where the stream pid is already closed/finished/complete

jeapostrophe (approved):

mattaudesse:
Thanks @jeapostrophe - the only change since your approval was to make CI not fail on a pending test.

jeapostrophe (dismissed):

PR #1224, opened 2022-06-29
Dct 489 debounce documentation site search

hamirmahal:
When working on [Docs site: Persistent search terms](https://reachsh.atlassian.net/browse/DCT-126), I noticed that debouncing documentation site search is necessary to prevent cluttering users' search histories.

Rather than have this code be part of that pull request, I thought it might make review easier to have this code separate.


docs/src/assets/scripts.js
```diff
@@ -188,7 +188,21 @@ const getWebpage = async (folder, hash, shallUpdateHistory) => {
...
+      let timer;
+      return (...args) => {
+        clearTimeout(timer);
+        timer = setTimeout(
+          () => { fn.apply(this, args); },
+          timeoutInMilliseconds
+        );
+      };
+    }
+    const processKeyUp = debounce(() => search());
```
jeapostrophe, line 204: This is a huge nitpick... debounce is generally useful, so you should define it at the top-level, not inside of `getWebpage` and you should define `processKeyUp` just before using it, or not at all, and just put `debounce(search)` directly into the `addEventListener` call

jeapostrophe (changes requested):

docs/src/assets/scripts.js
hamirmahal: That makes sense!

hamirmahal: [843d7a5](https://github.com/reach-sh/reach-lang/pull/1224/commits/843d7a5980433a887df42e754c3ee044e3a0610c) defines `debounce` at the top-level.

hamirmahal: [cf47db5](https://github.com/reach-sh/reach-lang/pull/1224/commits/cf47db533f297118ee84961dfc8586222c95e765) removes `processKeyUp` altogether.

docs/src/assets/scripts.js
```diff
@@ -236,7 +250,8 @@ const getWebpage = async (folder, hash, shallUpdateHistory) => {
         searchResultsList.append(e);
       });
       setClickFollowLink();
-    });
+    };
+    searchInput.addEventListener('keyup', debounce(() => search()));
```
jeapostrophe, line 254:
```suggestion
    searchInput.addEventListener('keyup', debounce(search));
```

docs/src/assets/scripts.js
```diff
@@ -16,6 +16,20 @@ const currentPage = {
...
+function debounce(fn, timeoutInMilliseconds = 250) {
+  let timer;
+  return (...args) => {
+    clearTimeout(timer);
+    timer = setTimeout(
+      () => { fn.apply(this, args); },
+      timeoutInMilliseconds
+    );
+  };
+}
```
jeapostrophe, line 31:
```suggestion
};
```

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #1241, opened 2022-07-07
CORE-2030 Use `Makefile` to check and update extension hover documentation

hirehamir:


jeapostrophe (changes requested):
I don't think this is the right way to do this. Instead, there should be a test in the CI that errors if the documentation is out of date and then an easy script that people can run to update it when that happens

jeapostrophe (changes requested):
I feel like I must be so aggravating Hamir, I'm sorry.

I don't like this either. I don't like needing to add a new build step and having to install npm and all that.

I think you should make a new target in `docs`'s `Makefile` for `check` and another for `check-accept`. `check` will run this test and `check-accept` will run it in a way that it does the update. (Essentially `check-accept` is running the update script and `check` runs `check-accept` but errors if there's a change.) These targets should use a Docker container for doing the execution, so you don't have to install anything and we can be consistent with our versions

jeapostrophe (approved):

hamirmahal:
It's okay! I'm glad we were able to get this through.

hamirmahal:
It's okay! I'm glad we were able to merge this.

PR #1253, opened 2022-07-12
js/ALGO: use limit(1) during verifyContract_

DanBurton:
Introduces the `limsug` argument to `deq`, `peq`, and `getTxns`. This argument allows the caller to suggest a limit to the number of transactions fetched. This feature is used to allow ALGO's `verifyContract_` to specify `limit(1)`.

Dan TODO: Needs testing. Idea: use the request monitor to show that when you call a view on a fresh contract handle, `limit=1` is used on the request, and the response has, in fact, one txn in the result, even though the ctc has multiple txns associated with it.


jeapostrophe (dismissed):
This is fine

DanBurton:
@jeapostrophe can you take a look at the new example? I assume you think it's also fine.

jeapostrophe (dismissed):
Ship it!

PR #1259, opened 2022-07-13
CORE-1411: Allow Explicit Token Tracking

chrisnevers:
Adds `Token.track`, which allows apps to track tokens published inside a container or received from an API call

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -460,7 +460,7 @@ range: 85 - 87
 
 ### {#ref-programs-consensus-token-minting} Token minting
 
-@{ref("rsh", "burn")}@{ref("rsh", "Token.burn")}@{ref("rsh", "destroy")}@{ref("rsh", "Token.destroy")}@{ref("rsh", "supply")}@{ref("rsh", "Token.supply")}@{ref("rsh", "destroyed")}@{ref("rsh", "Token.destroyed")}
+@{ref("rsh", "burn")}@{ref("rsh", "Token.burn")}@{ref("rsh", "destroy")}@{ref("rsh", "Token.destroy")}@{ref("rsh", "supply")}@{ref("rsh", "Token.supply")}@{ref("rsh", "destroyed")}@{ref("rsh", "Token.destroyed")}@{ref("rsh", "Token.track")}
 ```reach
 require(supply >= 2 * amt);
 const tok = new Token({ name, symbol, url, metadata, supply, decimals });
@@ -470,6 +470,9 @@ assert(tok.supply() == supply - amt);
 tok.burn();
 assert(tok.destroyed() == false);
 tok.destroy();
+
+const [ [tok2], k ] = call(API.getToken());
+Token.track(tok2);
 ```
 
 :::note
@@ -524,6 +527,10 @@ has been called on `{!rsh} tok` yet.
 ---
 
 `{!rsh} Token.supply(tok)`, or `{!rsh} tok.supply()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to query the current supply of tokens, i.e. the number of tokens which have not been burnt.
+
+---
+
+`{!rsh} Token.track(tok)`, or `{!rsh} tok.track()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to explicitly start tracking the balance of a Token in an application. This operation is useful for tracking a token that is received from an API call or published within a container, like an `{!rsh} Array` or `{!rsh} Data` object.
 
 ### Remote objects
 
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -530,7 +530,8 @@ has been called on `{!rsh} tok` yet.
 
 ---
 
-`{!rsh} Token.track(tok)`, or `{!rsh} tok.track()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to explicitly start tracking the balance of a Token in an application. This operation is useful for tracking a token that is received from an API call or published within a container, like an `{!rsh} Array` or `{!rsh} Data` object.
+`{!rsh} Token.track(tok)`, or `{!rsh} tok.track()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to explicitly start tracking the balance of a non-network token in an application.
+This operation is useful for tracking a token that is received from an API call or published within a container, like an `{!rsh} Array` or `{!rsh} Data` object.
 
 ### Remote objects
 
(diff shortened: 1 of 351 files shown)


hs/t/n/token_track_container_zero_bal.txt
```diff
@@ -0,0 +1,295 @@
...
+  when ALL participants are honest
+  of theorem: assert
+  msg: "balance zero at application exit"
+  at ./token_track_container_zero_bal.rsh:32:7:application
+
+  // Violation Witness
+
+  const UInt.max = 5;
+
+  const toks/753 = protect<Array(Token, 5)>("A".interact.getTokens());
```
chrisnevers, line 14: C++ errors be like:

docs/src/rsh/consensus/index.md
```diff
@@ -525,6 +528,10 @@ has been called on `{!rsh} tok` yet.
 
 `{!rsh} Token.supply(tok)`, or `{!rsh} tok.supply()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to query the current supply of tokens, i.e. the number of tokens which have not been burnt.
 
+---
+
+`{!rsh} Token.track(tok)`, or `{!rsh} tok.track()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to explicitly start tracking the balance of a Token in an application. This operation is useful for tracking a token that is received from an API call or published within a container, like an `{!rsh} Array` or `{!rsh} Data` object.
```
jeapostrophe, line 533:
```suggestion
`{!rsh} Token.track(tok)`, or `{!rsh} tok.track()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to explicitly start tracking the balance of a non-network token in an application.
This operation is useful for tracking a token that is received from an API call or published within a container, like an `{!rsh} Array` or `{!rsh} Data` object.
```

hs/src/Reach/Eval/Core.hs
```diff
@@ -2734,18 +2750,18 @@ evalPrim p sargs =
       let existingToks = st_toks st
       tokIsUniq <- tokIsUnique (map DLA_Var existingToks) $ DLA_Var tokdv
       doClaim CT_Assume tokIsUniq $ Just "New token is unique"
```
jeapostrophe, line 2752: This should be part of `trackToken`

hs/src/Reach/Eval/Core.hs
```diff
@@ -2734,18 +2750,18 @@ evalPrim p sargs =
...
-          }
+      setSt $ st { st_toks_c = S.insert tokdv (st_toks_c st) }
       let toka = DLA_Var tokdv
-      let mtok_a = Just $ toka
-      doBalanceInit' TM_Balance mtok_a supplya
+      trackToken toka supplya
       tokenMetaSet TM_Supply toka supplya False
       tokenMetaSet TM_Destroyed toka (DLA_Literal $ DLL_Bool False) False
       return $ public $ SLV_DLVar tokdv
+    SLPrim_Token_track -> do
```
jeapostrophe, line 2759: You need to call this in the `publish` code path that currently modifies the set of tokens, so we can delete code by doing this

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
chrisnevers: This assertion is specifically for assuming that the tokens we create via `new Token` are unique; it's not applicable to tracking tokens from an interact. Should I still move this to `trackToken` and conditionally run it for new tokens?

hs/src/Reach/Eval/Core.hs
```diff
@@ -4951,7 +4967,6 @@ doToConsensus ks (ToConsensusRec {..}) = locAt slptc_at $ do
         void $ locSt st_pure $ evalApplyVals' req_rator [public bv, public $ SLV_Bytes at $ "non-network tokens distinct"]
```
jeapostrophe, line 4967: I was referring to this one... when you start to track a token, you should ensure that it is distinct from what you've done so far

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
```diff
@@ -4957,14 +4963,11 @@ doToConsensus ks (ToConsensusRec {..}) = locAt slptc_at $ do
               , st_after_first = True
               }
       setSt st_recv
-      mapM_ (flip trackToken Nothing) tok_as
+      trackTokens $ map (,Nothing) toks
```
jeapostrophe, line 4966: Why did you change it to take a list? It seems way more complicated now?

hs/src/Reach/Eval/Core.hs
```diff
@@ -2654,21 +2654,28 @@ warnInteractType = \case
...
+      { st_toks = all_toks
+      , st_tok_pos = M.fromList all_toks_idx
       }
-  doBalanceInit' TM_Balance (Just tok) bal
+  forM_ toks_x_bals $ \ (tok, mBal) -> do
+    let bal = fromMaybe (DLA_Literal $ DLL_Int at UI_Word 0) mBal
+    doBalanceInit' TM_Balance (Just $ DLA_Var tok) bal
+  let req_rator = SLV_Prim $ SLPrim_claim CT_Require
+  -- Initialize and distinctize tokens
+  bv <- let f = map SLV_DLVar in evalDistinctTokens (f old_toks) (f toks)
```
jeapostrophe, line 2675: If you unwrap this function, you'll see that it just does an iteration anyways

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
chrisnevers: For `toConsensus` we can publish many tokens, and I thought that processing the distinct check for all the tokens at once would be more efficient. I can switch it back to handle one token at a time though

jeapostrophe: I think it will be the same amount of work but in a different order

hs/src/Reach/Eval/Core.hs
```diff
@@ -2654,29 +2654,44 @@ warnInteractType = \case
...
+--   let all_toks = old_toks <> toks
+--   let all_toks_idx = (flip zip [0 ..]) $ map DLA_Var all_toks
+--   setSt $
+--     st
+--       { st_toks = existingToks <> [tok]
+--       , st_tok_pos = M.fromList all_toks_idx
+--       }
+--   let bal = fromMaybe (DLA_Literal $ DLL_Int at UI_Word 0) mBal
+--   doBalanceInit' TM_Balance (Just $ DLA_Var tok) bal
+
```
jeapostrophe, line 2694: You should drop this

jeapostrophe (changes requested):

jeapostrophe (approved):
Approve, assuming CI is good

jeapostrophe (approved):

PR #1261, opened 2022-07-14
Fix website display for RE0033

Craig-Elre:
Fix website displaying and empty and excess paragraph


jeapostrophe (changes requested):
The correct fix is actually to put all of the `ref`s on the same line

jeapostrophe:
I'm about to push a fix for this

PR #1263, opened 2022-07-14
CORE-2031: Remove scrypt Dep

chrisnevers:


DanBurton (approved):
If CI says it's good then 👍

PR #1266, opened 2022-07-15
Add a code sample of the issue to `RE0066`

Craig-Elre:
Modify an existing `index.rsh` file to create a fork that ends in two different states of the program, and add it to the `/n` folder. Then create a code sample for the `RE0066` issue


jeapostrophe:
CI is busted

PR #1279, opened 2022-07-21
Eth api stack too deep

willghatch:
At this point this works, but I'm not sure quite what to do about errors.  In this PR I just use `error` in the ETH_Solidity file, but I think we want error codes like in the rest of the compiler, and I'm vaguely aware that at some point errors can be turned into JSON messages.  In a previous commit that was never in a PR I raised the error at the Eval layer to be able to use the error infrastructure that's already there (though I wasn't getting the connector info correctly).  The ALGO connector has some infrastructure for errors, but when I try to follow it it has several layers of indirection that I really don't understand, so I'm not in a great position to try to replicate it for ETH or try to abstract it into common infrastructure or anything.

Anyway, a primer on how errors work in the compiler would be great, I guess.


willghatch:
And I didn't add golden files because I'm confident that the error raised (and thus the golden file) needs to change before this is merged.

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1688,7 +1710,7 @@ solPLProg PLProg {plp_opts = plo, plp_init = dli, plp_cpprog = CPProg { cpp_at =
           let body'' = vsep $ body' <> [illegal]
           return $
             (,) (s2t k, Aeson.String $ s2t vk_) $
-              vsep $ defns <> [solFunction vk dom' ret body'']
+              vsep $ defns <> [solFunction vkWrapped [mayMemSol wrappedDomTy <> " _a"] retWrapped body''] <> [wrapper]
```
jeapostrophe, line 1713: Can you post an example (gist) of what the generated sol looks like?

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1654,6 +1661,10 @@ solPLProg PLProg {plp_opts = plo, plp_init = dli, plp_cpprog = CPProg { cpp_at =
           let vk = pretty $ vk_
           let (dom, rng) = itype2arr t
           args <- mapM (allocDLVar at) dom
+          when (length args > apiMaxArgs) $
+            error $ "The ETH connector supports Views that have up to " <>
+            show apiMaxArgs <> " arguments, but " <> vk_ <> " uses " <>
+            show (length args) <> "."
```
jeapostrophe, line 1667: Define something like

https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Connector/ALGO.hs#L55-L71

Probably there's just one constructor `Err_SolTooManyArgs String String Int` and this is

```
expect_throw Nothing at $ Err_SolTooManyArgs "Views" vk_ (length args)
```

https://github.com/reach-sh/reach-lang/blob/5124cfb30e147afe07c9a8a1cb8fdab74b3606a0/hs/src/Reach/AST/Base.hs#L214

And you have to add entries to the error code index --- https://github.com/reach-sh/reach-lang/blob/057eca2aba25a4d398f905fae4ab4a5611a54873/docs/src/rsh/errors/index.md

And a test case with the same name

jeapostrophe (changes requested):

hs/src/Reach/Connector/ETH_Solidity.hs
willghatch: https://gist.github.com/willghatch/da4f54ccbb635098872c7a8bdf1048cc#file-eth-api-view-stack-depth-mainserver-sol-L696

jeapostrophe: Cool

hs/src/Reach/Connector/ETH_Solidity.hs
willghatch: Ok, thanks.  I had started writing stuff like in the ALGO connector, but then I looked at how those errors were used and instead of `expect_throw` it used some other stuff defined in that file, so I was under the impression that `expect_throw` wasn't available and something else was needed.

willghatch:
Before implementing this I had really wanted to support an arbitrary number of arguments by building the struct manually with inline assembly, but then I read about Solidity's inline assembly and found that it doesn't allow direct stack manipulation.  That was a real disappointment.

jeapostrophe (approved):

jeapostrophe:
Excellent!

PR #1280, opened 2022-07-22
CORE-2067

jeapostrophe:


docs/dev/generator.js
```diff
@@ -504,8 +504,9 @@ const processMd = async ({baseConfig, relDir, in_folder, iPath, oPath}) => {
   const verOld = (ver) => `Version ${ver} is an old Reach release version.`;
   const verCur = (ver) => `Version ${ver} is the current Reach release version.`;
   const verRC = (ver) => `Version ${ver} is the current Reach release candidate version.`;
+  const rcHead = (ver) => `_Available in release candidate ${ver} and earlier..._`;
```
DanBurton, line 507:
```suggestion
  const rcHead = (ver) => `_Available in release candidate ${ver} and later..._`;
```

examples/launchToken/index.mjs
```diff
@@ -0,0 +1,12 @@
...
+
+const stdlib = loadStdlib();
+const [ accA, accB, accC, accD ] = await stdlib.newTestAccounts(4, stdlib.parseCurrency(100));
+
+await stdlib.launchToken(accA, 'Gil', 'GIL', {
+  clawback: accB,
+  freeze: accC,
+  reserve: accD,
+  defaultFrozen: true,
+});
```
DanBurton, line 12: Can this test then check the on-chain metadata of the token to demonstrate that those addresses are in fact set correctly?

DanBurton (changes requested):

PR #1291, opened 2022-07-25
CORE-1983: View Aliases

chrisnevers:


docs/kToD.ts
```diff
@@ -0,0 +1,345 @@
+/*
```
jeapostrophe, line 1: This file should not be committed. Can you add it to the gitignore so you'll notice?

hs/src/Reach/AST/DLBase.hs
```diff
@@ -1376,7 +1376,7 @@ flattenInterfaceLikeMap = M.fromList . concatMap go . M.toList
     go' :: Maybe SLPart -> (SLVar, a) -> (SLPart, a)
     go' mp (v, x) = (fromMaybe "" (fmap (flip (<>) "_") mp) <> bpack v, x)
 
-type DLViews = InterfaceLikeMap IType
+type DLViews = InterfaceLikeMap (IType, [B.ByteString])
```
jeapostrophe, line 1379: Turn the RHS (the pair) into a type alias, because you use it again in JS.hs

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1671,9 +1671,10 @@ solPLProg PLProg {plp_opts = plo, plp_init = dli, plp_cpprog = CPProg { cpp_at =
               , ext_defn
               ]
     map_defns <- mapM map_defn (M.toList dli_maps)
-    let tgo :: Maybe SLPart -> (SLVar, IType) -> App ((T.Text, Aeson.Value), Doc)
-        tgo v (k, t) = do
-          let vk_ = maybe k (\v' -> bunpack v' <> "_" <> k) v
+    let tgo :: Maybe SLPart -> (SLVar, (IType, [B.ByteString])) -> App ([(T.Text, Aeson.Value)], Doc)
```
jeapostrophe, line 1674: And here

hs/t/n/Err_SolTooManyArgs_view.txt
```diff
@@ -2,7 +2,7 @@ Verifying knowledge assertions
 Verifying for generic connector
   Verifying when ALL participants are honest
   Verifying when NO participants are honest
-Checked 7 theorems; No failures!
+Checked 5 theorems; No failures!
```
jeapostrophe, line 5: I think I committed this earlier

hs/t/n/Err_Alias_Name_Clash.rsh
```diff
@@ -0,0 +1,16 @@
...
+export const main = Reach.App(() => {
+  const A = View('A', {
+    x: UInt
+  }, {
+    x: ['y', 'z']
+  });
+  const B = View('B', {
+    xp: UInt
+  }, {
+    xp: ['y']
```
jeapostrophe, line 12: The documentation had better make clear that the interface is *not* included automatically with aliases

docs/src/rsh/appinit/index.md
```diff
@@ -196,11 +196,16 @@ This section is about defining views during application initialization. Views ar
...
+View('NFT', { owner: Address }, { owner: ["currentHolder"] })
+// or
+View({ owner: Address }, { owner: ["currentHolder"] })
 ```
 
 Views are read-only functions that can be called by other contracts, as well as off-chain.
 
-A view is defined with `{!rsh} View(viewName, viewInterface)` or `{!rsh} View(viewInterface)`, where `{!rsh} viewName` is a string that labels the view and `{!rsh} viewInterface` is an object where each field indicates the type of a function or value provided by the contract associated with the specified DApp.
+A view is defined with `{!rsh} View(viewName, viewInterface, ?viewAlias)` or `{!rsh} View(viewInterface, ?viewAlias)`, where `{!rsh} viewName` is a string that labels the view, `{!rsh} viewInterface` is an object where each field indicates the type of a function or value provided by the contract associated with the specified DApp, and `{!rsh} viewAlias` is an optional object mapping fields from the `{!rsh} viewInterface` to a tuple of aliases.
+An alias, which is a string, is another name that the view will be accessible from.
```
jeapostrophe, line 208: Make clear that the `viewName` is *not* prepended. Give a concrete example.

jeapostrophe (changes requested):

examples/view-alias/index.mjs
```diff
@@ -0,0 +1,54 @@
...
+
+const stdlib = loadStdlib(process.env);
+
+const startingBalance = stdlib.parseCurrency(100);
+
+const [accAlice, accBob] =
+  await stdlib.newTestAccounts(2, startingBalance);
+
+const ctcAlice = accAlice.contract(backend);
+
```
jeapostrophe, line 12: You could add a test that the ABI (on one of the connectors) is what you expect it to be.

hs/t/n/Err_Alias_Name_Clash.rsh
jeapostrophe: This is now inconsistent with your doc change. This file should define the following functions: `A_x, A_y, A_z, B_xp, B_y`. There is no name clash there. This suggests the documentation is wrong or the implementation. I think the implementation is probably the one at fault.

jeapostrophe (changes requested):

hs/t/n/view_alias_shadowed.txt
```diff
@@ -0,0 +1,8 @@
+reachc: error[RE0123]: The name `A_y` has already been used by a Participant, View or API at: ./view_alias_shadowed.rsh:5:16:application
```
jeapostrophe, line 1: Your comment that the name clash handling is already done... I think you should have just a few more examples. This is when a View alias conflicts with an API name. What about when a View alias conflicts with another View alias, or another name?

jeapostrophe (changes requested):

hs/t/n/view_alias_shadowed.txt
chrisnevers: ok, Ill add those

jeapostrophe (dismissed):
These look good to go when the CI is happy

jeapostrophe (approved):

PR #1292, opened 2022-07-25
add Token.accepted

willghatch:
Posting to ask for clarification.

I think the ALGO code I've written for the new `canReceive` function is probably correct, or at least close, but it's not actually being generated.  I think I probably have an issue with `Eval/Core.hs`.  In it, I'm mostly just following a pattern of what I see, but I don't have a very deep understanding of what's going on.  But I'm thinking maybe I'm using `ctxt_lift_expr` wrong or something.  Also, there are some files where I put something that lets me compile and get past a phase, but where I'm really not sure at all whether what I wrote is correct, eg. SMT and `Verify/Knowledge.hs`.

Anyway, I thought this was a good point to ask for any pointers.


examples/can-receive/index.rsh
```diff
@@ -0,0 +1,47 @@
...
+  init()
+
+
+  Alice.publish()
+  const tok = new Token({supply: 10, decimals: 0})
+  commit()
+
+  Bob.publish()
+  // Use canReceive in a consensus step.
+  const x = canReceive(Bob, tok) ? 1 : 0
```
jeapostrophe, line 23: `x` is not used, so it is probably being optimized away. And you should name this function `Token.accepted(acc, tok)` or `Account.tokenAccept(acc, tok)` [or both] not `canReceive`

hs/src/Reach/AST/DLBase.hs
```diff
@@ -781,6 +781,7 @@ data DLExpr
   | DLE_Claim SrcLoc [SLCtxtFrame] ClaimType DLArg (Maybe B.ByteString)
   | DLE_Transfer SrcLoc DLArg DLArg (Maybe DLArg)
   | DLE_TokenInit SrcLoc DLArg
+  | DLE_CanReceive SrcLoc DLArg DLArg
```
jeapostrophe, line 784: I'd call it `DLE_TokenAccepted`

hs/src/Reach/AST/SL.hs
```diff
@@ -771,6 +771,7 @@ data SLPrimitive
   | SLPrim_remote
   | SLPrim_remotef SrcLoc DLArg (Maybe String) SLTypeFun (Maybe SLVal) (Maybe (Either SLVal SLVal)) (Maybe DLRemoteALGO) (Maybe RemoteFunMode)
   | SLPrim_balance
+  | SLPrim_canReceive
```
jeapostrophe, line 774: `SLPrim_TokenAccepted`

hs/src/Reach/Backend/JS.hs
```diff
@@ -456,6 +456,19 @@ jsExpr = \case
...
+        -- TODO - I'm not really sure what to do here, but I imagine I should do something similar to the simulate case.
+        --addr' <- jsArg addr
+        --tok' <- jsArg tok
+        --return $ "await" <+> jsApply (addr' <> ".tokenAccepted") [tok']
+        return "false"
+      JM_View -> impossible "view canReceive"
+      JM_Simulate -> do
+        addr' <- jsArg addr
+        tok' <- jsArg tok
+        return $ "await" <+> jsApply (addr' <> ".tokenAccepted") [tok']
```
jeapostrophe, line 471: A few things...
- Simulation case --- This needs to log the address as one of the addresses that will be included in the txn. That may require a new simulation function in the JS contract interface (which will require an ABI version increment)
- Backend case --- You can't actually call `address.tokenAccepted`, because this code may be evaluated at a different time than the consensus code, after the user has opted-out and can no longer receive it. Thus, you'll need to log the result (but using `DLE_EmitLog` in the code generation in `Eval`)

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2058,6 +2058,12 @@ ce = \case
       addInitTok tok
       void $ checkTxn $ CheckTxn {..}
       void $ makeTxn $ MakeTxn {..}
+  DLE_CanReceive _ addr tok -> do
+    ca addr
```
jeapostrophe, line 2062: You need to note the resource usage for the analyzer

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -679,6 +680,11 @@ solExpr sp = \case
   DLE_Transfer _ who amt mtok ->
     spa $ solTransfer who amt mtok
   DLE_TokenInit {} -> return emptyDoc
+  DLE_CanReceive _ _ _ -> do
+    --addr' <- solArg addr
+    --tok' <- solArg tok
+    -- TODO - this is not really correct, because this is throwing away the arguments, which could potentially be effectful.  Also, this is not being generated anyway, so something is wrong before here, keeping me from getting here.
```
jeapostrophe, line 686: No, they cannot be effectful. Reach's AST is like ANF, so its arguments are always simple and if they were effectful to generate, that would be annotate on their let

hs/src/Reach/Eval/Core.hs
```diff
@@ -2784,6 +2785,13 @@ evalPrim p sargs =
         [] -> getBalanceOf Nothing
         [v] -> getBalanceOf . Just =<< compileCheckType T_Token v
         _ -> illegal_args
+    SLPrim_canReceive -> do
+      at <- withAt id
+      (addr_, tok_) <- two_args
+      addr <- compileCheckType T_Address addr_
+      tok <- compileCheckType T_Token tok_
+      dv <- ctxt_lift_expr (DLVar at Nothing T_Bool) $ DLE_CanReceive at addr tok
```
jeapostrophe, line 2793: Check out how we use `DLE_EmitLog` elsewhere, but this is basically correct

hs/src/Reach/Simulator/Core.hs
```diff
@@ -603,6 +603,7 @@ instance Interp DLExpr where
               transferLedger simContract acc ev n
               return V_Null
     DLE_TokenInit _at _dlarg -> return V_Null
+    DLE_CanReceive _at _dlargA _dlargB -> return $ V_Null
```
jeapostrophe, line 606: This would return true, because in the simulated network, every account can accept every token

jeapostrophe: (i.e. you just introduced a soundness violation)

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1132,6 +1132,9 @@ smt_e at_dv mdv de = do
       mempty
     DLE_TokenInit {} ->
       mempty
+    DLE_CanReceive at _ _ ->
+      -- TODO - I'm really uncertain what to do here, this just lets me get past this phase.
+      unbound at
```
jeapostrophe, line 1137: This is fine. It says that the verifier can't guess whether it is accepted or not.

jeapostrophe (changes requested):

examples/can-receive/index.rsh
willghatch: `x` is used by an interact function.  Can that still be optimized away?  FYI I named it `canReceive` because that's what you wrote in the Jira card, but I agree that the other options are better (they better match the stdlib function).

hs/src/Reach/Backend/JS.hs
willghatch: Can you tell me a little about this bit of code?  My current belief (aka best guess) is that JM_Backend is essentially for code running in local steps, JM_Simulate is for the JS to simulate consensus steps, and JM_View is for simulating views (which on Algorand must be simulated but which on ETH are actually run on a node).  Is that correct?

hs/src/Reach/Connector/ALGO.hs
jeapostrophe: You need
```
incResource R_Account addr
incResource R_Asset tok
```

examples/can-receive/index.rsh
jeapostrophe: It has to be used in a consensus step. For example, you could have `Bob.publish(z)` and then do `check(x == z)` and it will require `x` to be computed and saved

hs/src/Reach/Backend/JS.hs
jeapostrophe: `JM_Backend` is for "client-side code", which is when the client runs the consensus code locally AND for local steps. (Every consensus computation is duplicated locally.)

`JM_Simulate` is for simulating the next consensus step to compute what resources will be used. You'll need to add `addr` to the account list AND `tok` to the token list

You have to add something like this... https://github.com/reach-sh/reach-lang/blob/master/js/stdlib/ts/ALGO.ts#L1821-L1823 .... but call `recordAccount` and `recordToken` (but no fee). You'll be adding a new kind of simulation txn... https://github.com/reach-sh/reach-lang/blob/master/js/stdlib/ts/shared_impl.ts#L570

willghatch: I think I've mostly addressed things here, except I still don't know what to do for the JM_Backend case.  I don't see any other examples of JM_Backend that give me any useful hints.  It needs to return a bool... but it's a bool determined by the state of the consensus network.  So... I'm basically lost for how we can do that besides querying the network.

jeapostrophe: You should have it just return `true`. The `EmitLog` thing will ignore it

examples/token-accepted/index.mjs
```diff
@@ -11,20 +11,27 @@ const accBob = await stdlib.newTestAccount(startingBalance);
 const ctcAlice = accAlice.contract(backend)
 const ctcBob = accBob.contract(backend, ctcAlice.getInfo());
 
+const algoP = stdlib.connector == "ALGO"
```
jeapostrophe, line 14: Make sure to include `;`

examples/token-accepted/index.rsh
```diff
@@ -0,0 +1,53 @@
...
+  const tok = new Token({supply: 10, decimals: 0})
+  commit()
+
+  Bob.only(() => {
+    const bobX = declassify(interact.giveBool())
+  })
+
+  Bob.publish(bobX)
+  const x = Token.accepted(Bob, tok)
+  enforce(bobX == x)
```
jeapostrophe, line 28: I actually meant this...
```
Bob.publish()
const x = ....
commit();

Bob.only(() => { const bobX = x; });
Bob.publish(bobX);
check(bobX == x);
```

hs/src/Reach/Backend/JS.hs
```diff
@@ -456,19 +456,20 @@ jsExpr = \case
...
-        --return $ "await" <+> jsApply (addr' <> ".tokenAccepted") [tok']
-        return "false"
+        -- TODO - This is wrong, but I'm still not sure what to do here.
+        return $ "false"
       JM_View -> impossible "view canReceive"
       JM_Simulate -> do
-        addr' <- jsArg addr
+        acc' <- jsArg acc
         tok' <- jsArg tok
-        return $ "await" <+> jsApply (addr' <> ".tokenAccepted") [tok']
```
jeapostrophe: You should do this part AND add the `tokenAccepted` sim txn

jeapostrophe: The code you have is the wrong type, because you're returning a sim txn where a bool is expected

hs/src/Reach/Eval/Core.hs
```diff
@@ -2785,12 +2786,15 @@ evalPrim p sargs =
         [] -> getBalanceOf Nothing
         [v] -> getBalanceOf . Just =<< compileCheckType T_Token v
         _ -> illegal_args
-    SLPrim_canReceive -> do
+    SLPrim_Token_accepted -> do
       at <- withAt id
       (addr_, tok_) <- two_args
+      -- TODO - The doInternalLog bit makes it so this needs to be in a consensus step.  Can I check the consensus step and generate something different for a local step?  It seems like it would be nice to be able to use this feature in both consensus and local steps.
```
jeapostrophe, line 2792: I think it is okay to not allow it locally, because it relies on consensus (network) state

js/stdlib/ts/shared_impl.ts
```diff
@@ -593,6 +593,10 @@ export type ISimTxn<Token, ContractInfo> = {
 } | {
   kind: 'tokenDestroy',
   tok: Token,
+} | {
+  kind: 'tokenAccepted',
```
jeapostrophe, line 597: You need to increment the JS version number in JS.hs and in the two connectors because of this

jeapostrophe (changes requested):
Changelog and docs

hs/src/Reach/Backend/JS.hs
willghatch: Help me understand JS.hs a little more.  The `JM_Simulate` code I'm writing only runs for Algorand as far as I can tell.  Also it's wrong, because I have the address as a string and I'm not sure how to convert it into an address to do `acc.tokenAccepted(tok)` (and crashing there for ALGO but not for ETH is how I know it's only running for Algorand).  Under what conditions are things simulated, and why do I want to call `.tokenAccepted` in the simulate case but not the backend case?

jeapostrophe: - Yes, it only runs for Algorand
- Look under the covers of `acc.tokenAccepted(tok)` and call a helper function (that maybe you need to make) called `accountTokenAccepted(acc, tok)` that doesn't need an account object
- On Algorand, we have to simulate before EVERY consensus step to learn what resources are used. That's why we have the sim txn to learn that the account and token are used. But, suppose the program is `transfer(1).to(tokenAccepted(a, b) ? c : d)`. We need to decide whether to include `c` or `d`.
- In the backend case, because you did the emit log, the value will be totally ignored anyways... we'll look at the log and find out what the real consensus did when it ran (because the "backend" code runs AFTER the consensus step, whereas the "simulate" code runs BEFORE)

hs/src/Reach/Backend/JS.hs
```diff
@@ -456,20 +456,20 @@ jsExpr = \case
...
-        return $ "false"
+      JM_Backend -> return "true"
       JM_View -> impossible "view canReceive"
       JM_Simulate -> do
-        acc' <- jsArg acc
+        addr' <- jsArg addr
         tok' <- jsArg tok
+        let simTxn = jsSimTxn "tokenAccepted" $ [("tok", tok'), ("addr", addr')]
+        -- TODO - this doesn't work because addr' is a JS string, so it doesn't have a tokenAccepted method.  How do I transform it into an address here?  Or am I going about this the wrong way?
+        let accepted = "await" <+> jsApply (addr' <> ".tokenAccepted") [tok']
```
jeapostrophe, line 468: Something that often works is to add a new function, `simTokenAccepted`, to the interface and have it be the runtime code that calls the appropriate helper. But not necessarily... this is kind of a taste/figure-out-what-is-already there question.

hs/src/Reach/Backend/JS.hs
```diff
@@ -456,20 +456,20 @@ jsExpr = \case
...
         return $
-          jsSimTxn "tokenAccepted" $
-            [ ("tok", tok')
-            , ("acc", acc')
-            ]
+          "await" <+> jsApply
+          (parens $ "async () => " <>
+           (jsBraces $ vsep [simTxn, "return" <+> accepted])) []
```
jeapostrophe, line 472: This is not necessary. You are already in an async context. If you're just doing this because you want to have the effect, my advice is to (a) use the comma operator or (b) have the `simTokenAccepted` function take the effect of `jsSimTxn` as an argument (which is ignored)

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2776,7 +2776,7 @@ ct = \case
 
 -- Reach Constants
 reachAlgoBackendVersion :: Int
-reachAlgoBackendVersion = 10
+reachAlgoBackendVersion = 11
```
jeapostrophe, line 2779: This is not needed. You only do this when you change the connector representation

js/stdlib/ts/ALGO.ts
```diff
@@ -158,8 +158,8 @@ export type NetworkAccount = {
   sk?: SecretKey
 };
 
-const reachBackendVersion = 17;
-const reachAlgoBackendVersion = 10;
+const reachBackendVersion = 18;
+const reachAlgoBackendVersion = 11;
```
jeapostrophe, line 162: Reminder... not needed

jeapostrophe (changes requested):

js/stdlib/ts/shared_backend.ts
```diff
@@ -304,3 +304,9 @@ export const simTokenBurn = (sim_r:any, tok:any, amt:any): void => {
 export const simTokenDestroy = (sim_r:any, tok:any): void => {
   sim_r.txns.push({kind: 'tokenDestroy', tok});
 };
+
+export const simTokenAccepted = (sim_r:any, addr:any, tok:any): boolean => {
+  sim_r.txns.push({kind: 'tokenAccepted', addr, tok});
+  // TODO - I need to query for acceptance here, but that requires things from ALGO.ts...
```
willghatch, line 310: The internals of `acc.tokenAccepted` don't require an account object, but they do require a provider and any way I slice it I need to either get stuff from `ALGO.ts` or re-implement bigger things than I would like to.  `shared_backend.ts` says it has no dependencies on other shared things (although it does import some stuff...).  How do you feel about moving code from `ALGO.ts` into `shared_impl`?

willghatch:
I made a comment on one of the new lines of code, but I'm not sure if you can see it.  (It's a comment on `shared_impl.ts`.)

willghatch:
Oh, now it shows up.  I guess I just wasn't patient enough or needed to refresh or something.

js/stdlib/ts/shared_backend.ts
jeapostrophe: You should add a function to `IContractCompiled` --- https://github.com/reach-sh/reach-lang/blob/master/js/stdlib/ts/shared_impl.ts#L187 --- and generate a call like `ctc.simTokenAccepted` and it will call `tokenAcceptedThingy` and pass it the provider of the account that created the `etc` handle

jeapostrophe (commented):
.

willghatch:
Well, it seems like it's almost there, but while my frontend on algorand accepts the token and then sees the acceptance of the token, the consensus still returns false.  My teal output seems right to me according to the docs.  But either my teal is wrong or accepting the token isn't really working somehow, though.  Any ideas?  Anyway, that's where I'm ending for today.

js/stdlib/ts/ALGO.ts
```diff
@@ -2341,21 +2343,36 @@ const balanceOfM = async (acc: Account, token: Token | null): Promise<BigNumber
...
-      const client = await getAlgodClient();
-      const query = client.accountAssetInformation(addr, tokenId) as ApiCall<AccountAssetInfo>;
-      const accountAssetInfoM = await doQueryM_(dhead, query);
-      if ('val' in accountAssetInfoM) {
-        const assetHolding = accountAssetInfoM['val']['asset-holding'];
-        return assetHolding ? bigNumberify(assetHolding['amount']) : false;
-      } else {
-        return false;
-      }
+      return indexerBalanceOfMNonNet(addr, token, dhead);
```
jeapostrophe, line 2346: This function was labeled "indexer token balance", but it is actually the node balance and the code is identical

js/stdlib/ts/ALGO.ts
```diff
@@ -2341,21 +2343,36 @@ const balanceOfM = async (acc: Account, token: Token | null): Promise<BigNumber
...
+  const query = client.accountAssetInformation(addr, bigNumberToNumber(token)) as unknown as ApiCall<AccountAssetInfo>;
+  const accountAssetInfoM = await doQueryM_(dhead, query);
+  if ('val' in accountAssetInfoM) {
+    return bigNumberify(accountAssetInfoM['val']['asset-holding']?.['amount'] ?? 0);
+  } else {
+    return false;
+  }
+}
+
+const indexerBalanceOfMNonNet = async (addr:string, token:Token, dhead:string) => {
```
jeapostrophe, line 2362: These functions are the same and you should just do one

js/stdlib/ts/ALGO.ts
```diff
@@ -1632,6 +1633,13 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
         return await viewMapRef(i, f);
       };
 
+      const simTokenAccepted = async (sim_r:any, addr:any, tok:any): Promise<boolean> => {
+        simTokenAccepted_(sim_r, addr, tok);
+        // TODO - both of these give the same answer in my test client.  I probably should use them conditionally like in the original function they were ripped from.
+        //return !!(await indexerBalanceOfMNonNet(addr, tok, "balanceOfM"));
+        return !!(await clientBalanceOfMNonNet(addr, tok, "balanceOfM"));
```
jeapostrophe, line 1640: You should do `x !== false`

js/stdlib/ts/ETH_like.ts
```diff
@@ -565,6 +565,7 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account>
         // @ts-ignore
         return res;
       };
+      const simTokenAccepted = async (_sim_r:any, _addr:any, _tok:any): Promise<boolean> => true;
```
jeapostrophe, line 568: You should call the `simTokenAccepted_` function to be thorough

js/stdlib/ts/shared_impl.ts
```diff
@@ -345,14 +346,15 @@ export const stdContract =
   const _initialize = () => {
     const {
       getContractInfo, getContractAddress, getContractCompanion,
-      getBalance, sendrecv, recv, getCurrentStep, getState, apiMapRef
+      getBalance, sendrecv, recv, getCurrentStep, getState, apiMapRef,
+      simTokenAccepted
```
jeapostrophe, line 350: If you wanted to be clever, you could make it so that this returns the `_` function and then you generate the one that does the `sim_txn` setting

js/stdlib/ts/ALGO.ts
jeapostrophe: The string should be `simTokenAccepted`

jeapostrophe (changes requested):

jeapostrophe:
I don't see the debug log in the CI run (because `hs-test` isn't passing). If you want some help, send me a DM of a gist of the debugging log and I'll tell you if I see anything. (See the generated TEAL or .pil would be useful too)

js/stdlib/ts/ALGO.ts
```diff
@@ -2305,49 +2315,54 @@ const balancesOfM = async (acc: Account, tokens: Array<Token|null>): Promise<Arr
...
         return false;
       }
+    }
+  } else {
+    return balanceOfMNonNet(addr, token);
+  }
+}
+
+const balanceOfMNonNet = async (addr:string, token:Token) => {
+  // TODO - I think these can and should be merged.  I think the only meaningful difference is that one case returns 0 where the other returns false, which is probably a bug due to drift.  But I'm not sure which side is correct.
```
willghatch, line 2346: I want to remove the if statement and just use one side.  My guess is that if there is no `asset-holding` field then the return should be false, IE the second case, but I'm not 100% certain.  I could try simplifying to each side and see what tests break both ways, but thought I'd ask first.

willghatch: Oh, I see I missed some of your comments from earlier, including one that talks about this.

js/stdlib/ts/ALGO.ts
willghatch: Yeah, I hadn't looked closely at what they were doing at that point, I was just taking things apart structurally.

docs/src/rsh/consensus/index.md
```diff
@@ -533,6 +533,11 @@ has been called on `{!rsh} tok` yet.
 `{!rsh} Token.track(tok)`, or `{!rsh} tok.track()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to explicitly start tracking the balance of a non-network token in an application.
 This operation is useful for tracking a token that is received from an API call or published within a container, like an `{!rsh} Array` or `{!rsh} Data` object.
 
+---
+
+`{!rsh} Token.accepted(acc, tok)`, where `{!rsh} acc` is an `{!rsh} Account` value and `{!rsh} tok` is a `{!rsh} Token` value, returns true when `{!rsh} acc` has opted in to `{!rsh} tok`, and false otherwise.
```
jeapostrophe, line 538: We never use the term "opt in". Check out existing docs about the same concept... https://docs.reach.sh/frontend/#js_tokenAccept

js/stdlib/ts/ALGO.ts
jeapostrophe: Yea

jeapostrophe (changes requested):

docs/src/rsh/consensus/index.md
```diff
@@ -533,6 +533,11 @@ has been called on `{!rsh} tok` yet.
 `{!rsh} Token.track(tok)`, or `{!rsh} tok.track()`, where `{!rsh} tok` is a `{!rsh} Token` value, may be used to explicitly start tracking the balance of a non-network token in an application.
 This operation is useful for tracking a token that is received from an API call or published within a container, like an `{!rsh} Array` or `{!rsh} Data` object.
 
+---
+
+`{!rsh} Token.accepted(acc, tok)`, where `{!rsh} acc` is an `{!rsh} Account` value and `{!rsh} tok` is a `{!rsh} Token` value, returns true if `{!rsh} acc` is ready to accept non-network tokens specified by `{!rsh} tok`, and false otherwise.
+This always returns true on some consensus networks, but should always be used to ensure your frontend is blockchain agnostic.
```
jeapostrophe, line 539:
```suggestion
This always returns true on some consensus networks, but could be useful on others where accounts need to explicitly agree to receive non-network tokens.
```

js/stdlib/ts/ALGO.ts
```diff
@@ -1632,6 +1633,11 @@ export const connectAccount = async (networkAccount: NetworkAccount): Promise<Ac
         return await viewMapRef(i, f);
       };
 
+      const simTokenAccepted = async (sim_r:any, addr:any, tok:any): Promise<boolean> => {
+        simTokenAccepted_(sim_r, addr, tok);
+        return await balanceOfMNonNet(addr, tok) !== false;
```
jeapostrophe, line 1638:
```suggestion
        return (await balanceOfMNonNet(addr, tok)) !== false;
```

jeapostrophe (changes requested):

jeapostrophe:
Merged!

PR #1296, opened 2022-07-26
CORE-1403 DLExpr for Object.set and Tuple.set

erer1243:
This is my initial implementation of just Object.set for core-1403. Probably has a few errors; putting this up for review while I work on Tuple.set.

### DIFF
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -2025,6 +2025,14 @@ ce = \case
   DLE_TupleRef at ta idx -> do
     ca ta
     cTupleRef at (argTypeOf ta) idx
+  DLE_TupleSet at tup_a index val_a -> do
+    let tup_t = argTypeOf tup_a
+    let tupFields = tupleTypes tup_t
+    let (_, valStart, valLen) = computeExtract tupFields index
+    let tupLen = typeSizeOf tup_t
+    ca tup_a
+    ca val_a
+    csplice at valStart (valStart + valLen) tupLen
   DLE_ObjectRef _at oa f -> do
     let fts = argObjstrTypes oa
     let fidx = objstrFieldIndex (argTypeOf oa) f
@@ -2033,10 +2041,11 @@ ce = \case
     cextract start sz
     cfrombs t
   DLE_ObjectSet at obj_a fieldName val_a -> do
-    let fieldTypes = argObjstrTypes obj_a
+    let obj_t = argTypeOf obj_a
+    let fieldTypes = objstrTypes obj_t
     let fieldIndex = objstrFieldIndex (argTypeOf obj_a) fieldName
     let (_, valStart, valLen) = computeExtract (map snd fieldTypes) fieldIndex
-    let objLen = typeSizeOf $ argTypeOf obj_a
+    let objLen = typeSizeOf obj_t
     ca obj_a
     ca val_a
     csplice at valStart (valStart + valLen) objLen
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -2027,26 +2027,19 @@ ce = \case
     cTupleRef at (argTypeOf ta) idx
   DLE_ObjectRef _at oa f -> do
     let fts = argObjstrTypes oa
-    let fidx = fromIntegral $ fromMaybe (impossible "field") $ List.findIndex ((== f) . fst) fts
+    let fidx = objstrFieldIndex (argTypeOf oa) f
     let (t, start, sz) = computeExtract (map snd fts) fidx
     ca oa
     cextract start sz
     cfrombs t
   DLE_ObjectSet at obj_a fieldName val_a -> do
-    -- This makes the (unsound?) assumption that when adding a new field to an object,
-    -- all copied fields should stay in the order they were already in.
-    let val_t = argTypeOf val_a
-    let oldFieldTypes = argObjstrTypes obj_a
-    let newFieldTypes = M.toAscList $ M.insert fieldName val_t $ M.fromList oldFieldTypes
-    let fieldIdx = fromIntegral $ fromJust $ List.elemIndex fieldName $ map fst newFieldTypes
-    let (_, valStart, oldValLen_) = computeExtract (map snd newFieldTypes) fieldIdx
-    let oldValLen = if length oldFieldTypes == length newFieldTypes then oldValLen_ else 0
+    let fieldTypes = argObjstrTypes obj_a
+    let fieldIndex = objstrFieldIndex (argTypeOf obj_a) fieldName
+    let (_, valStart, valLen) = computeExtract (map snd fieldTypes) fieldIndex
     let objLen = typeSizeOf $ argTypeOf obj_a
-    let newValLen = typeSizeOf val_t
-    when (objLen + newValLen - oldValLen > 4096) $ warn "object too large" -- what to do?
     ca obj_a
     ca val_a
-    csplice at valStart (valStart + oldValLen) objLen
+    csplice at valStart (valStart + valLen) objLen
   DLE_Interact {} -> impossible "consensus interact"
   DLE_Digest _ args -> cdigest $ map go args
     where
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -2026,31 +2026,16 @@ ce = \case
     ca ta
     cTupleRef at (argTypeOf ta) idx
   DLE_TupleSet at tup_a index val_a -> do
-    let tup_t = argTypeOf tup_a
-    let tupFields = tupleTypes tup_t
-    let (_, valStart, valLen) = computeExtract tupFields index
-    let tupLen = typeSizeOf tup_t
     ca tup_a
     ca val_a
-    ctobs $ argTypeOf val_a
-    csplice at valStart (valStart + valLen) tupLen
-  DLE_ObjectRef _at oa f -> do
-    let fts = argObjstrTypes oa
-    let fidx = objstrFieldIndex (argTypeOf oa) f
-    let (t, start, sz) = computeExtract (map snd fts) fidx
-    ca oa
-    cextract start sz
-    cfrombs t
+    cTupleSet at (argTypeOf tup_a) index
+  DLE_ObjectRef at obj_a fieldName -> do
+    ca obj_a
+    uncurry (cTupleRef at) $ objectRefAsTupleRef obj_a fieldName
   DLE_ObjectSet at obj_a fieldName val_a -> do
-    let obj_t = argTypeOf obj_a
-    let fieldTypes = objstrTypes obj_t
-    let fieldIndex = objstrFieldIndex (argTypeOf obj_a) fieldName
-    let (_, valStart, valLen) = computeExtract (map snd fieldTypes) fieldIndex
-    let objLen = typeSizeOf obj_t
     ca obj_a
     ca val_a
-    ctobs $ argTypeOf val_a
-    csplice at valStart (valStart + valLen) objLen
+    uncurry (cTupleSet at) $ objectRefAsTupleRef obj_a fieldName
   DLE_Interact {} -> impossible "consensus interact"
   DLE_Digest _ args -> cdigest $ map go args
     where
@@ -2418,6 +2403,14 @@ ce = \case
       op "itxn_submit"
       code "itxn" ["CreatedApplicationID"]
   where
+    -- On ALGO, objects are represented identically to tuples of their fields in ascending order.
+    -- Consequently, we can pretend objects are tuples and use tuple functions as a shortcut.
+    objectRefAsTupleRef :: DLArg -> String -> (DLType, Integer)
+    objectRefAsTupleRef obj_a fieldName = (objAsTup_t, fieldIndex)
+      where
+        fieldIndex = objstrFieldIndex obj_t fieldName
+        objAsTup_t = T_Tuple $ map snd $ objstrTypes obj_t
+        obj_t = argTypeOf obj_a
     show_stack :: String -> Maybe BS.ByteString -> SrcLoc -> [SLCtxtFrame] -> App ()
     show_stack what msg at fs = do
       let msg' =
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -2032,6 +2032,21 @@ ce = \case
     ca oa
     cextract start sz
     cfrombs t
+  DLE_ObjectSet at obj_a fieldName val_a -> do
+    -- This makes the (unsound?) assumption that when adding a new field to an object,
+    -- all copied fields should stay in the order they were already in.
+    let val_t = argTypeOf val_a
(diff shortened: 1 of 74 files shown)


hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2032,6 +2032,21 @@ ce = \case
     ca oa
     cextract start sz
     cfrombs t
+  DLE_ObjectSet at obj_a fieldName val_a -> do
+    -- This makes the (unsound?) assumption that when adding a new field to an object,
```
jeapostrophe, line 2036: This is fine, as long as the types are the same. That is, we turn a 2d point into a 3d point by doing `DLE_ObjectSet sb pointA "z" 3`

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2032,6 +2032,21 @@ ce = \case
...
+    -- all copied fields should stay in the order they were already in.
+    let val_t = argTypeOf val_a
+    let oldFieldTypes = argObjstrTypes obj_a
+    let newFieldTypes = M.toAscList $ M.insert fieldName val_t $ M.fromList oldFieldTypes
+    let fieldIdx = fromIntegral $ fromJust $ List.elemIndex fieldName $ map fst newFieldTypes
+    let (_, valStart, oldValLen_) = computeExtract (map snd newFieldTypes) fieldIdx
+    let oldValLen = if length oldFieldTypes == length newFieldTypes then oldValLen_ else 0
+    let objLen = typeSizeOf $ argTypeOf obj_a
+    let newValLen = typeSizeOf val_t
+    when (objLen + newValLen - oldValLen > 4096) $ warn "object too large" -- what to do?
```
jeapostrophe, line 2046: You don't need this

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2032,6 +2032,21 @@ ce = \case
     ca oa
     cextract start sz
     cfrombs t
+  DLE_ObjectSet at obj_a fieldName val_a -> do
+    -- This makes the (unsound?) assumption that when adding a new field to an object,
+    -- all copied fields should stay in the order they were already in.
+    let val_t = argTypeOf val_a
+    let oldFieldTypes = argObjstrTypes obj_a
+    let newFieldTypes = M.toAscList $ M.insert fieldName val_t $ M.fromList oldFieldTypes
```
jeapostrophe, line 2040: These two things should be exactly the same

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2032,6 +2032,21 @@ ce = \case
...
+    let fieldIdx = fromIntegral $ fromJust $ List.elemIndex fieldName $ map fst newFieldTypes
+    let (_, valStart, oldValLen_) = computeExtract (map snd newFieldTypes) fieldIdx
+    let oldValLen = if length oldFieldTypes == length newFieldTypes then oldValLen_ else 0
```
jeapostrophe, line 2043: This should just be the size of `val_a`

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2032,6 +2032,21 @@ ce = \case
...
     cextract start sz
     cfrombs t
+  DLE_ObjectSet at obj_a fieldName val_a -> do
+    -- This makes the (unsound?) assumption that when adding a new field to an object,
+    -- all copied fields should stay in the order they were already in.
+    let val_t = argTypeOf val_a
+    let oldFieldTypes = argObjstrTypes obj_a
+    let newFieldTypes = M.toAscList $ M.insert fieldName val_t $ M.fromList oldFieldTypes
+    let fieldIdx = fromIntegral $ fromJust $ List.elemIndex fieldName $ map fst newFieldTypes
+    let (_, valStart, oldValLen_) = computeExtract (map snd newFieldTypes) fieldIdx
```
jeapostrophe, line 2042: I believe that there is code for this elsewhere. It shouldn't be duplicated, so either call that thing or make something new that can be called

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -662,6 +663,16 @@ solExpr sp = \case
           T_Object {} -> objPrefix
           _ -> impossible "objectref"
     return $ oe' <> "." <> p (pretty f) <> sp
+  DLE_ObjectSet _ obj_a fieldName val_a -> do
+    let objFields = M.fromList $ argObjstrTypes obj_a
+    result_t <- solType $ T_Object $ M.insert fieldName (argTypeOf val_a) objFields
```
jeapostrophe, line 668: This should be the same as `obj_a`'s type

hs/src/Reach/Eval/Core.hs
```diff
@@ -3122,6 +3122,20 @@ evalPrim p sargs =
...
+      at <- withAt id
+      (obj, fieldV, valV) <- three_args
+      (objTy, objDLA) <- compileTypeOf obj
+      (valTy, valDLA) <- compileTypeOf valV
+      fieldName <- bunpack <$> mustBeBytes fieldV
+      case objTy of
+        T_Object objFields -> do
+          let objFields' = M.insert fieldName valTy objFields
+          let objTy' = T_Object objFields'
+          let dle = DLE_ObjectSet at objDLA fieldName valDLA
```
jeapostrophe, line 3135: You need to enforce that the type is the same. You'd do this by checking that `objTy` already has `fieldName` with the type `valTy`

hs/src/Reach/Eval/Core.hs
```diff
@@ -1449,9 +1449,9 @@ evalAsEnvM sv@(lvl, obj) = case obj of
   SLV_Prim SLPrim_Object ->
     return $ Just $
       M.fromList
-        [ ("set", retStdLib "Object_set")
+        [ ("set", retV $ public $ SLV_Prim SLPrim_Object_set)
```
jeapostrophe, line 1452: The type enforcement means that this change is not backwards compatible. A way to deal with that is to have a branch in the implementation that does it the "old" way if `Object.set` would change the type

hs/t/y/object_set6.rsh
```diff
@@ -0,0 +1,17 @@
+'reach 0.1';
+
+export const main = Reach.App(() => {
```
jeapostrophe, line 3: You should do an example that changes the type

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1120,6 +1120,23 @@ smt_e at_dv mdv de = do
...
+        val_se <- smt_a at val_a
+        result_se <- smt_v at result_v
+        bound at result_se
+        -- Assert unchanged fields match the old object
+        let copiedFields = M.toList $ M.delete fieldName objFields
+        forM_ copiedFields $ \(key, _) -> do
+          smtAssertCtxt $ smtEq (smtApply (objSort    <> "_" <> key) [obj_se])
+                                (smtApply (resultSort <> "_" <> key) [result_se])
+        -- Assert the changed field matches the given val
+        smtAssertCtxt $ smtEq val_se (smtApply (resultSort <> "_" <> fieldName) [result_se])
```
jeapostrophe, line 1139: This is not a good implementation. It should be `bound at $ smtApply objCtor args` where `args` is `oldValuesBefore <> newValue :: oldValuesAfter`

jeapostrophe (changes requested):

hs/src/Reach/Connector/ALGO.hs
erer1243: That is only true if we're replacing a field with the exact size as val_a

hs/src/Reach/Connector/ALGO.hs
erer1243: Not if we're adding a new field or changing the field's type

hs/src/Reach/Connector/ETH_Solidity.hs
erer1243: Same situation as similar comment on ALGO.hs; not true if we're adding a new field or changing the field's type

hs/src/Reach/Connector/ALGO.hs
jeapostrophe: Types all have fixed sizes

hs/src/Reach/Connector/ALGO.hs
jeapostrophe: We don't want to support that

hs/src/Reach/Connector/ETH_Solidity.hs
jeapostrophe: See prior comments about this

hs/t/y/object_set6.rsh
erer1243: I'm not sure what you mean. Don't 4/6 of these examples change the type?

hs/src/Reach/Eval/Core.hs
```diff
@@ -3124,17 +3124,22 @@ evalPrim p sargs =
...
+          let objFields' = M.insert fieldName val_t objFields
+          let result_t = T_Object objFields'
+          -- DLE_ObjectSet only applies when updating an existing field,
+          -- not when adding a new field or changing a field's type
+          case obj_t == result_t of
+            True -> do
+              let dle = DLE_ObjectSet at obj_a fieldName val_a
+              dlv <- ctxt_lift_expr (DLVar at Nothing result_t) dle
+              return (lvl, SLV_DLVar dlv)
+            False -> undefined -- no idea what parts of this module would pertain to this
```
jeapostrophe, line 3142: We would go back to the "old" way of doing `Object.set` where you re-construct the object from scratch with a new field. You should replace this `undefined` with `error XXX` and make a test case that has that result if you're having a hard time find one

jeapostrophe (changes requested):
Noice!

hs/t/y/object_set6.rsh
jeapostrophe: I'm just dumb then and couldn't tell

hs/src/Reach/Eval/Core.hs
erer1243: well, the old way is calling Object_set, which does {...o, k: v}; then, that syntax is exploded before even reaching this phase. So, I think I need to do the old way right here. Currently looking at explodeTupleLike to figure that out.

erer1243:
I've implemented a pretty similar version of Tuple.set; you mentioned linearize on Wednesday, but I'm not really sure what goes on in linearize. What should I do there?

hs/src/Reach/Eval/Core.hs
jeapostrophe: You could write `explodingObject_set` in the standard library and then generate a call to the standard library. I think that is better than trying to reimplement what it does here

jeapostrophe:
This is a tuple set that is hidden --- https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Linearize.hs#L504-L508

hs/src/Reach/Simulator/Core.hs
```diff
@@ -687,6 +687,7 @@ instance Interp DLExpr where
       -- XXX make an interact point with dr
       return $ V_Contract $ fromIntegral ctcId
     DLE_ObjectSet {} -> undefined -- TODO
+    DLE_TupleSet {} -> undefined -- TODO
```
jeapostrophe, line 690: You should do these

jeapostrophe (changes requested):

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2048,6 +2049,7 @@ ce = \case
     let objLen = typeSizeOf obj_t
     ca obj_a
     ca val_a
+    ctobs $ argTypeOf val_a
     csplice at valStart (valStart + valLen) objLen
```
jeapostrophe, line 2053: I think for code cleanliness, the Tuple and Object cases should be merged together with the Object ones doing some light processing before to get the data that the tuple ones need. I suspect that `DLE_ObjectRef` can also easily use `cTupleRef` (or rather... that there's a function inside `cTupleRef` that it could use)

jeapostrophe (changes requested):

hs/src/Reach/Connector/ALGO.hs
erer1243: I see what you're getting at

erer1243: cTupleSet already existed.. who knew

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2395,6 +2399,14 @@ ce = \case
       op "itxn_submit"
       code "itxn" ["CreatedApplicationID"]
   where
+    -- On ALGO, objects are represented identically to tuples of their fields in ascending order.
+    -- Consequently, we can pretend objects are tuples and use tuple functions as a shortcut.
+    objectRefAsTupleRef :: DLArg -> String -> (DLType, Integer)
```
jeapostrophe, line 2404: Very pretty

hs/src/Reach/Linearize.hs
```diff
@@ -491,21 +491,13 @@ df_com mkk back = \case
...
           [ asn info $ DLE_ArrayRef at infos idx
-          , asn bal $ infoAt 0
-          , asn supply $ infoAt 1
-          , asn destroyed $ infoAt 2
-          , asn info' $ DLE_LArg at $ DLLA_Tuple
-            [ if meta == TM_Balance   then newVal else DLA_Var bal
-            , if meta == TM_Supply    then newVal else DLA_Var supply
-            , if meta == TM_Destroyed then newVal else DLA_Var destroyed
-            ]
+          , asn info' $ DLE_TupleSet at (DLA_Var info) newValIndex newVal
```
jeapostrophe, line 500: My guess is that this is going to change a bunch of golden files for verification errors

hs/t/n/hamza5.txt
```diff
@@ -11,23 +11,23 @@ Verification failed:
...
   // Theorem Formalization
 
-  const tokInfo/266 = tokenInfos/245[0];
+  const tokInfo/233 = tokenInfos/212[0];
   //    ^ would be [8855, 8856, false ]
-  const v267 = tokInfo/266[0];
+  const v234 = tokInfo/233[0];
   //    ^ would be 8855
-  const tokInfos'/282 = Array.set(tokenInfos/245, 0, [(v267 - v267), tokInfo/266[1], tokInfo/266[2] ] );
+  const tokInfos'/246 = Array.set(tokenInfos/212, 0, Tuple.set(tokInfo/233, 0, (v234 - v234) ) );
```
jeapostrophe, line 24: This is very nice

vsce/data/print-keyword-info.json
```diff
@@ -725,6 +725,9 @@
   "exit": {
     "CompletionItemKind": "Function"
   },
+  "explodingObject_set": {
```
jeapostrophe, line 728: The recent change is going to modify this

jeapostrophe (commented):
This looks ready to go, but once you rebase, the `vsce/data` will be wrong.

erer1243:
Rebase didn't seem to cause any issues

jeapostrophe (approved):

jeapostrophe:
Excellent

PR #1306, opened 2022-08-01
DCT-life-guide

TheChronicMonster:
Dan's Lifecycle Guide


DanBurton (changes requested):
Sorry, I didn't notice my review was requested on this earlier. This looks great, BUT, the confluence doc was updated (somewhat reorganized & extended) based on feedback from various people; can you update this to incorporate the changes? (Skip the "see also: <link to google doc> but everything else should be good to include.)

PR #1307, opened 2022-08-01
Core 2090 fix incorrect ide stack trace parsing when the error is raised from rsh stdlib

hirehamir:
### Without this change

image

### With this change

image

### DIFF
--- a/hs/src/Reach/AST/Base.hs
+++ b/hs/src/Reach/AST/Base.hs
@@ -171,24 +171,26 @@ data CompilationError = CompilationError
   , ce_position :: [Int]
   , ce_offendingToken :: Maybe String
   , ce_errorCode :: String
+  , ce_stackTrace :: String
   }
   deriving (Show, Generic, ToJSON, FromJSON)
 
-makeCompilationError :: (ErrorSuggestions a, ErrorMessageForJson a, Show a, HasErrorCode a) => SrcLoc -> a -> CompilationError
-makeCompilationError src err =
+makeCompilationError :: (ErrorSuggestions a, ErrorMessageForJson a, Show a, HasErrorCode a) => SrcLoc -> a -> [String] -> CompilationError
+makeCompilationError src err stackTrace =
   CompilationError
     { ce_suggestions = snd $ errorSuggestions err
     , ce_offendingToken = fst $ errorSuggestions err
     , ce_errorMessage = errorMessageForJson err
     , ce_position = srcloc_line_col src
     , ce_errorCode = makeErrCode (errPrefix err) (errIndex err)
+    , ce_stackTrace = unwords stackTrace
     }
 
 encodeJSONString :: ToJSON a => a -> String
 encodeJSONString = map w2c . LB.unpack . encode
 
-makeErrorJson :: (ErrorSuggestions a, ErrorMessageForJson a, Show a, HasErrorCode a) => SrcLoc -> a -> String
-makeErrorJson src err = encodeJSONString $ makeCompilationError src err
+makeErrorJson :: (ErrorSuggestions a, ErrorMessageForJson a, Show a, HasErrorCode a) => SrcLoc -> a -> [String] -> String
+makeErrorJson src err stackTrace = encodeJSONString $ makeCompilationError src err stackTrace
 
 data CompileErrorException = CompileErrorException
   { cee_error :: CompilationError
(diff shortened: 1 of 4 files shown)


jeapostrophe (approved):

vsce/server/src/server.ts
```diff
@@ -558,15 +563,29 @@ CallStack (from HasCallStack):
...
     //connection.console.log(`Tokens: ` + tokens);
-    const linePos = errorJson.ce_position[0] - 1;
-    const charPos = errorJson.ce_position[1] - 1;
+    let linePos = errorJson.ce_position[0] - 1;
+    let charPos = errorJson.ce_position[1] - 1;
     const suggestions = errorJson.ce_suggestions;
     const actualMessage = errorJson.ce_errorMessage;
     const offendingToken = errorJson.ce_offendingToken;
     const reachCompilerErrorCode = errorJson.ce_errorCode;
+    if (reachCompilerErrorCode === "RE0089") {
```
DanBurton, line 578: This stack trace parsing is likely useful for more than just this particular error code. This is fine for now, but keep it in mind in case a similar issue arises for similar errors.

vsce/server/src/server.ts
```diff
@@ -558,15 +563,29 @@ CallStack (from HasCallStack):
...
+      const { ce_stackTrace } = errorJson;
+      // Match "...at (/app/index.rsh:8:12:application)".
+      const regexp: RegExp = /index.rsh:(\d+):(\d+):.+$/;
```
DanBurton, line 581: Is there a way to ask the extension for "current file name" so that this can work for more than just `index.rsh`?

vsce/server/src/server.ts
hamirmahal: Will do; thank you for your input!

vsce/server/src/server.ts
hamirmahal: So, the `index.rsh` here isn't actually an `index.rsh` on the user's local device: it's a temporary file created in a temporary directory that we compile.

hamirmahal: image

hs/src/Reach/AST/Base.hs
```diff
@@ -171,24 +171,26 @@ data CompilationError = CompilationError
...
-makeCompilationError src err =
+makeCompilationError :: (ErrorSuggestions a, ErrorMessageForJson a, Show a, HasErrorCode a) => SrcLoc -> a -> [String] -> CompilationError
+makeCompilationError src err stackTrace =
   CompilationError
     { ce_suggestions = snd $ errorSuggestions err
     , ce_offendingToken = fst $ errorSuggestions err
     , ce_errorMessage = errorMessageForJson err
     , ce_position = srcloc_line_col src
     , ce_errorCode = makeErrCode (errPrefix err) (errIndex err)
+    , ce_stackTrace = unwords stackTrace
```
chrisnevers, line 186: This might be more useful in the future if the field were `string[]` instead of `string`, which should be valid JSON. If keeping `string`, I think `unlines` makes more sense than `unwords`

hs/src/Reach/Verify/SMT.hs
```diff
@@ -665,10 +665,12 @@ display_fail tat f tk mmsg mrd mdv timeout = do
             []
             (M.toList pm_str_val)
       putLine $ show $ showTrace pm_dv_val smtTrace
+  let stackMsgs = map getStackTraceMessage f
+  let stackTraceTop = if length stackMsgs > 10 then take 10 stackMsgs <> ["  ..."] else stackMsgs
```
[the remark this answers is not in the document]
hamirmahal: Ah, that is a cleaner approach. I'm planning to make that change in a new pull request.

hs/src/Reach/AST/Base.hs
hamirmahal: I do want to keep `ce_stackTrace` a `string`, because then I can just do `String.match` on `ce_stackTrace`, instead of
```ts
ce_stackTrace.forEach(line => {
  const someRegEx = /.+/;
  const result = line.match(someRegEx);
  if (result) {
    // do something
  }
});
```

I think the approach using `unwords` is cleaner, at least for the time being. The Regular Expression on the TypeScript side appears to work more nicely when `ce_stackTrace` isn't joined using `\n`.

I'll keep `unlines` in mind for future reference, though. Thank you for your input!

hs/src/Reach/Verify/SMT.hs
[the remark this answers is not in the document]
hamirmahal: Ah, that is a cleaner approach. I'll make that change now.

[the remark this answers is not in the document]
hamirmahal: #1309

PR #1310, opened 2022-08-03
Data tag

willghatch:
Is this the direction you're looking for, Jay?

Right now the DLE_DataTag has a major flaw in that it doesn't check which `Data` type something is, so this version of `isSome` doesn't either.  I thought “I'll fix this with the `is` function”, but `Maybe` is a type constructor and not a type, so I can't use `is(x, Maybe)` and there is no concrete type that I can pass in there to be correct for `isSome`.


hs/rsh/stdlib.rsh
```diff
@@ -35,8 +35,10 @@ export const fromMaybe = (v, onNull, onSome) => {
   case None: return onNull();
   case Some: return onSome(v); } };
 
-export const isSome = m => fromMaybe(m, () => false, (_) => true);
-export const isNone = m => fromMaybe(m, () => true, (_) => false);
+// TODO - this definitely emits optimized teal code, but it no longer checks that the data is actually a proper Maybe type...
```
chrisnevers, line 38: Maybe a silly idea, but what about this for typechecking?
```
const isSome = m => {
  assert(m.match({ Some: true, None: true });
 return _dataTag(m) == 1;
}

File: hs/rsh/stdlib.rsh [hunk 2]
```diff
@@ -35,8 +35,10 @@ export const fromMaybe = (v, onNull, onSome) => {
...
+// TODO - also optimize isLeft/isRight, etc
+export const isSome = m => _dataTag(m) === 1;
+export const isNone = m => _dataTag(m) === 0;
```
jeapostrophe, line 41: Idea: The new thing is `DLE_DataTag` but the SL feature is `isDataVariant(variantName, otherVariants, instance)` and the type checking happens just in SL but compiles to DL

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2343,6 +2343,10 @@ ce = \case
         -- [ extra ]
         return ()
     label after_lab
+  DLE_DataTag _ d -> do
+    ca d
+    cint 0
+    op "getbyte"
```
jeapostrophe, line 2349: Double check if this is really a number

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -730,6 +731,10 @@ solExpr sp = \case
   DLE_TimeOrder {} -> impossible "timeorder"
   DLE_EmitLog {} -> impossible "emitLog"
   DLE_setApiDetails {} -> impossible "setApiDetails"
+  DLE_DataTag _ d -> do
+    d' <- solArg d
+    -- TODO - what kind of number should this be converted into?
```
jeapostrophe, line 736: `uint256`

hs/src/Reach/Eval/Core.hs
```diff
@@ -624,6 +624,7 @@ base_env_slvals =
   , ("call", SLV_Form $ SLForm_apiCall)
   , (".setApiDetails", SLV_Form $ SLForm_setApiDetails)
   , ("getUntrackedFunds", SLV_Prim $ SLPrim_getUntrackedFunds)
+  , ("_dataTag", SLV_Prim $ SLPrim_DataTag)
```
jeapostrophe, line 627: If it really were this feature, it would be `unsafeDataTag`

hs/src/Reach/Eval/Core.hs
```diff
@@ -3798,6 +3799,19 @@ evalPrim p sargs =
       dv <- ctxt_lift_expr mdv $ DLE_PrimOp at (ADD UI_Word) [DLA_Var untrackedFunds, trackedBal]
       setBalance TM_Balance mtok (DLA_Var dv)
       return (lvl, SLV_DLVar untrackedFunds)
+    SLPrim_DataTag -> do
+      v <- one_arg
+      at <- withAt id
+      case v of
+        SLV_Data _ dmap vn _ -> do
+          let tm = M.fromAscList $ zip (map fst $ M.toAscList dmap) ([0 ..] :: [Integer])
```
jeapostrophe, line 3807: I think this is the function you defined

hs/src/Reach/Verify/SMT.hs
```diff
@@ -1190,6 +1190,17 @@ smt_e at_dv mdv de = do
...
+      d' <- smt_a at d
+      n <- smtTypeSort $ argTypeOf d
+      let variants = map fst $ M.toAscList $ dataTypeMap $ argTypeOf d
+      let variantCtors = map (\v -> n <> "_" <> v) variants
+      let variantVs = map (\vCtor -> Atom $ vCtor <> "_pv") variantCtors
+      let variantPs = map (\(vV, vC) -> List [Atom vC, vV])
+           $ zip variantVs variantCtors
+      let variantCs = map (\(vP, i) -> List [vP, Atom $ show i])
+           $ zip variantPs ([0 ..] :: [Int])
+      bound at $ smtApply "match" [d', List variantCs]
```
jeapostrophe, line 1203: This should probably be a function that gets defined with the data type and then we just call it

jeapostrophe (changes requested):

hs/src/Reach/Connector/ALGO.hs
willghatch: I'm not sure I understand this feedback.  Do you mean “double check that getbyte returns a number”?  If so, the docs say that it returns an integer.

jeapostrophe: Yea. That's what I meant

hs/rsh/stdlib.rsh
willghatch: Thanks!  Good idea, although superseded by Jay's suggestion.

jeapostrophe (changes requested):
Docs, Changelog, tests (static & dynamic)

jeapostrophe:
Excellent

willghatch:
The failing test looks like it gets a slightly different witness on my machine, thus the golden file doesn't match (because I changed it to what I see, and it looks like the CI is getting a different message).  I bet this is just my version of Z3 being out of sync (I see it was just updated).

docs/src/changelog/index.md
```diff
@@ -6,6 +6,7 @@ Versions and changes-within-versions are listed in reverse-chronological order:
 ## 0.1.11: 2022/06 - present
 
 @{verRC("0.1.11")}
++ 2022/08/03: Added support for `{!rsh} isDataVariant`.
```
jeapostrophe, line 9: This should be under rc6 (but actually, you will be beat and it will end up being in rc7)

jeapostrophe: Also, just say "Added ..."

examples/isDataVariant/index.rsh
```diff
@@ -0,0 +1,25 @@
...
+    getDT: Fun([], DT),
+  });
+  init();
+
+  A.only(() => {
+    const x = declassify(interact.getDT());
+  });
+  A.publish(x);
+  const isDV = isDataVariant("A", ["A", "B"], x);
+  enforce(isDV);
```
jeapostrophe, line 16: This is an immoral use of `enforce` :)

docs/src/rsh/compute/index.md
```diff
@@ -1361,6 +1361,13 @@ As a special case, when the type of a variant is `{!rsh} Null`, the `{!rsh} VALU
 
 Data instances are consumed by `{!rsh} switch` statements and `{!rsh} match` expressions.
 
+@{ref("rsh", "isDataVariant")}
+Data instances may be checked against particular variants with the `{!rsh} isDataVariant` function.
+
+```reach
+isDataVariant("Circle", ["Circle", "Square", "Rect"], nice); // true
+```
+
```
jeapostrophe, line 1370: You could say "This is like the following switch... but more efficient"

jeapostrophe (changes requested):

examples/isDataVariant/index.rsh
willghatch: Bleh, it's the easiest way to be sure this kind of test code isn't optimized out of a consensus step :p

jeapostrophe: :hankey:

jeapostrophe (approved):

PR #1315, opened 2022-08-03
stay.py

DanBurton:
Compare & contrast w/ go.py, I guess?

`stay.py` performs per-user analyses to understand the longevity of these user ids.

Sample output from yesterday:

```
$ cat out/stay_day_summary.json  | jq
{
  "user_count": 7264,
  "min_earliest": "2020-08-24",
  "max_latest": "2022-08-02",
  "min_length": 0,
  "median_length": 0,
  "mean_length": 4.135600220264317,
  "max_length": 699,
  "stdev_length": 19.640628643522113,
  "active_cutoff": "2022-07-03",
  "active_count": 531,
  "oldest_active_id": "Numerius Negidius",
  "oldest_active_length": 699
}
```

The mean is 4 days, with a standard deviation of about 20 days.
The oldest active ID is someone who put their own ID in there named "Numerius Negidius". Not naming names but I have 2 guesses about who that might be.

### DIFF
--- a/scripts/compile-log-analyses/stay.py
+++ b/scripts/compile-log-analyses/stay.py
@@ -1,0 +1,126 @@
+import json
+import datetime
+from statistics import mean, median, stdev
+import go
+
+def group_by_id(rows):
+  groups = {}
+  for row in rows:
+    id = go.get_id(row)
+    group = groups.get(id, [])
+    groups[id] = group
+    group.append(row)
+  return groups
+
+def d(s, unit):
+  if unit == 'day':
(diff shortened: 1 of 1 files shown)


jeapostrophe (approved):

jeapostrophe:
Hah! I've been caught, kind of. That name is "John Smith" in Latin and is the default name --- https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Report.hs#L44 --- I think you should ignore it in the analysis

PR #1317, opened 2022-08-04
remove underscored identifiers from --print-keyword-info (CORE-1533)

willghatch:
Looking at the sprint backlog, this was just such low hanging fruit that I thought I would knock it out before doing any real work today.


DanBurton (approved):
sweet 👍

jeapostrophe (approved):

PR #1320, opened 2022-08-05
Fields

willghatch:
I liked the name `Object.fields` more than `Object.keys`, because `Object.keys` exists in JS and returns a list of keys (an array of strings), not an object like we are returning from it.


docs/src/changelog/index.md
```diff
@@ -9,6 +9,7 @@ Versions and changes-within-versions are listed in reverse-chronological order:
 
 + @{rcHead("0.1.11-rc7")}
 + 2022/08/03: Added `{!rsh} isDataVariant`.
++ 2022/08/03: Added `{!rsh} Struct.fields` and `{!rsh} Object.fields`.
```
jeapostrophe, line 12: In the future, they should be in order of commits. No need to change this.

docs/src/rsh/compute/index.md
```diff
@@ -1279,9 +1279,21 @@ these fields may be accompanied by additional fields specified afterwards.
 
 Shorthand for `{!rsh} { x: x, z: 5}`, where `{!rsh} x` is any bound identifier.
 
+#### `Object.fields`
+@{ref("rsh", "Object.fields")}
+
+```reach
+const o1 = Object({a: UInt, b: UInt});
+const fields = Object.fields(o1); // {a: UInt, b: UInt}
+const o2 = Object(...fields, c: UInt));
```
jeapostrophe, line 1288:
```suggestion
const o2 = Object({...fields, c: UInt}));
```

jeapostrophe (changes requested):
Nice, very small change

jeapostrophe (approved):

jeapostrophe:
The keyword file might be wrong now

docs/src/changelog/index.md
willghatch: I also realize now that I didn't change the date after copy/pasting either.  So I'm fixing the order and the date.

jeapostrophe (approved):

PR #1326, opened 2022-08-10
CORE-1861: Support Exponential Notation

chrisnevers:


hs/src/Reach/Eval/Core.hs
```diff
@@ -4440,34 +4441,83 @@ evalId_ lab x = do
 evalId :: String -> SLVar -> App SLSVal
 evalId lab x = sss_sls <$> evalId_ lab x
 
+parseCoefficient :: String -> EvalError -> App (Rational, Maybe Int)
+parseCoefficient b err = do
+  let parseInt i =
+        (fromIntegral iInt, precision)
+        where precision = Nothing
+              iInt = read i :: Integer
```
jeapostrophe, line 4449: I think, but I'm not sure, that these `read`s can be `readMaybe`s and then we can generate errors or impossibles. But I could be wrong

hs/src/Reach/Eval/Core.hs
```diff
@@ -4440,34 +4441,83 @@ evalId_ lab x = do
...
+        return $ (toRational $ (read b :: Double), mPrec)
+    [i] -> return $ parseInt i
+    _ -> expect_ err
+
+parseExponent :: String -> (Integer, Bool)
+parseExponent x = do
+  let x' = case x of
+            '+':x'' -> x''
+            ow -> ow
+  let xInt = abs $ read x' :: Integer
```
jeapostrophe, line 4464: Another readMaybe?

jeapostrophe (changes requested):
"Changes" may be dismissed if I am wrong, but if I am right. If we can do maybe, and it is possible to generate from user input, then there should be tests, but otherwise they are impossibles.

jeapostrophe (approved):

PR #1327, opened 2022-08-11
Integrate verifier

kimmyg:
This PR integrates an automatic verifier into the Algorand connector, conditional on `REACH_EVEREST` being defined.

### DIFF
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE CPP #-}
+
 module Reach.Connector.ALGO (connect_algo, AlgoError (..)) where
 
 import Control.Monad.Extra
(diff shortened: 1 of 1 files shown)


hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE CPP #-}
```
jeapostrophe, line 1: I am anxious about how annoying CPP is. Idea: Create `Reach.Connector.ALGO_Verify` and have it use the macro and have a dummy implementation. This way we are also type-checking the code even when there's no Everest?

jeapostrophe (changes requested):

hs/src/Reach/Connector/ALGO.hs
kimmyg: Do you mean
1. `Reach.Connector.ALGO`  has no `CPP`;
2. `Reach.Connector.ALGO` always runs `Reach.Connector.ALGO_Verify.run`
3. if Everest is around, `Reach.Connector.ALGO_Verify.run = Reach.Closed.TAF.Verify.run`; and
4. otherwise, `Reach.Connector.ALGO_Verify.run _ _ = return ()`?

jeapostrophe: Exactly

jeapostrophe (approved):

PR #1338, opened 2022-08-15
CORE-1004: Consider Safe Ops Impure For Ternary

chrisnevers:
I thought of making `isPure` for `DLE_PrimOp` check the operator's purity instead of always returning `True`, but it caused some issues with how we optimize impure assignments. I think it makes sense to still treat safe operations as pure for the sake of optimization and this at least fixes the ternary problem

### DIFF
--- a/hs/t/y/token_track_container.txt
+++ b/hs/t/y/token_track_container.txt
@@ -3,3 +3,5 @@ Verifying for generic connector
   Verifying when ALL participants are honest
   Verifying when NO participants are honest
 Checked 116 theorems; No failures!
+WARNING: Compiler instructed to emit for Algorand, but the conservative analysis found these potential problems:
+ * Step 2 may use up to 10 transaction references, but the limit is 8. Step 2 starts at ./token_track_container.rsh:24:5:dot.
--- a/hs/t/y/token_track_container.txt
+++ b/hs/t/y/token_track_container.txt
@@ -3,5 +3,3 @@ Verifying for generic connector
   Verifying when ALL participants are honest
   Verifying when NO participants are honest
 Checked 116 theorems; No failures!
-WARNING: Compiler instructed to emit for Algorand, but the conservative analysis found these potential problems:
- * Step 2 may use up to 10 transaction references, but the limit is 8. Step 2 starts at ./token_track_container.rsh:24:5:dot.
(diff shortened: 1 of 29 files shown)


jeapostrophe:
> I thought of making `isPure` for `DLE_PrimOp` check the operator's purity instead of always returning `True`,

That is what I assumed you'd do

> but it caused some issues with how we optimize impure assignments.

Can you elaborate?

>  I think it makes sense to still treat safe operations as pure for the sake of optimization and this at least fixes the ternary problem

I don't understand how this fixes it and is not what I suggested... You've made it so that `DLS_Let` with a safe operation is impure... but the only time that primops show up from the Core evaluator is inside of `DLS_Let`, so basically we ARE marking them as impure, but we're only doing that inside of Core and not inside of anything else (e.g. optimization), is that an explanation of this? If so, you should add this as a comment in the code on the `DLS_Let` case AND the `PrimOp` code, so that future Jay & Nevers know what is going on and why

jeapostrophe:
I realized that this is also going to require a documentation change. And, the way you've done it won't prevent a future optimizer from turning an `if` into an `ite` (which is something that I want to do and partially the motivation for introducing the `veri*` functions in the first place)

chrisnevers:
Ok, I'll go the IsPure route. From what I understand, the issues I encountered were due to impure expressions not being erased when unused because they cause effects. So, variables bound to `DLE_Impossible` in `forall` blocks weren't erased and neither was unused arithmetic in CHandlers which led to unbound variable errors. I thought of a potential workaround for the first part, but perhaps there's a way to address both at once

hs/t/y/token_track_container.txt
```diff
@@ -3,3 +3,5 @@ Verifying for generic connector
   Verifying when ALL participants are honest
   Verifying when NO participants are honest
 Checked 116 theorems; No failures!
+WARNING: Compiler instructed to emit for Algorand, but the conservative analysis found these potential problems:
+ * Step 2 may use up to 10 transaction references, but the limit is 8. Step 2 starts at ./token_track_container.rsh:24:5:dot.
```
jeapostrophe, line 7: Do you know why this appeared now?

chrisnevers: The dynamic token lookup used a safe addition which prevented optimization from happening; I changed it to a verified addition and fixed it. No assertions will be needed for the addition since the verified op is generated after Core

jeapostrophe: This answer seems to say something that is the opposite of the compiler output. You're saying that you enabled an optimization, but this error is essentially saying that the program is too big.

jeapostrophe: Ah, and now I see you're referring to a new commit

jeapostrophe (dismissed):
Assuming CI passes

jeapostrophe (approved):

PR #1340, opened 2022-08-15
CORE-2094 Add option to disable automatic token tracking of published tokens

erer1243:


jeapostrophe (approved):
This looks great. I re-ran the CI

PR #1350, opened 2022-08-19
CORE-2276: Tweaks For Better Compilation

chrisnevers:


hs/src/Reach/Optimize.hs
```diff
@@ -612,6 +631,7 @@ optLet at x e = do
                 meh
   case (extract x, (isPure e && canDupe e), e) of
     (Just dv, True, _) -> doit dv
+    (Just dv, _, DLE_PrimOp {}) -> doit dv
```
jeapostrophe, line 634: It looks to me that this is the most important part and you should make a comment about why we are ignoring purity here

hs/src/Reach/AST/DLBase.hs
```diff
@@ -641,6 +641,7 @@ instance Pretty ApiInfo where
 data PrimVM -- Primitive Verification Mode
   = PV_Safe -- No static assertion, yes dynamic check
   | PV_Veri -- Yes static assertion, no dynamic check
+  | PV_None
```
jeapostrophe, line 644: It doesn't seem like this is significant. Am I wrong?

chrisnevers: One of the issues I noticed with duoswap was that `tokenInfos` couldn't be erased because balance changes were impure. I don't want to make them `PV_Veri` because it would require users to make assertions. So this is useful for being considered pure and not requiring assertions. There are still assertions made about the balance before the operation, but not as part of `evalPrimOp`

jeapostrophe (approved):
Approve assuming CI

hs/src/Reach/AST/DLBase.hs
jeapostrophe: I see

PR #1355, opened 2022-08-19
Verify multiplication zero

willghatch:
This fixes the bug with verification of multiplication overflow.  The problem is that verification that overflow doesn't happen relies on using division, but SMT defines division by zero as zero instead of an error.  This allows a situation where if the second argument of a multiplication can be zero then it triggers division by zero, leading to failure to prove the desired fact about the division.  This adds a check for whether the second argument is zero and does a logical OR of the zero test and the division test.


jeapostrophe (dismissed):
Nice

willghatch:
I tested locally that it solved the problem, but I haven't run any other tests locally, thinking “I'll let the CI do it”.  So... I don't know what's up with the CI, but we should re-run when it's feeling better.

jeapostrophe (approved):

PR #1359, opened 2022-08-22
CORE-2064 Munge view args to fix bug on eth

erer1243:
maybe i also have to increment stdlib version? i don't recall

### DIFF
--- a/hs/src/Reach/Backend/JS.hs
+++ b/hs/src/Reach/Backend/JS.hs
@@ -1136,7 +1136,8 @@ jsViews (cvs, vis) = do
             return $ jsWhen c $ vsep [let', ret']
       let enInfo' :: Maybe SLPart -> (SLVar, DLView) -> App ([(SLVar, Doc)], Doc)
           enInfo' v (k, (vt, aliases)) = do
-            let (_, rng) = itype2arr vt
+            let (dom, rng) = itype2arr vt
+            dom' <- jsArray <$> mapM jsContract dom
             rng' <- jsContract rng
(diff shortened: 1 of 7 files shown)


hs/src/Reach/Backend/JS.hs
```diff
@@ -1145,7 +1146,8 @@ jsViews (cvs, vis) = do
             let view_asn = "const " <> name <> " = " <> decode' <> ";"
             let val = jsObject $
                         M.fromList $
-                          [ ("ty" :: String, rng')
+                          [ ("dom" :: String, dom')
+                          , ("rng", rng')
```
jeapostrophe, line 1150: Yes, this would require incrementing the version number

js/stdlib/ts/ALGO.ts
```diff
@@ -176,7 +176,7 @@ export interface ProviderEnv {
 const defaultALGO_TOKEN_HEADER = 'X-Algo-API-Token';
 const defaultALGO_INDEXER_TOKEN_HEADER = 'X-Indexer-API-Token';
 
-const reachBackendVersion = 21;
+const reachBackendVersion = 22;
```
jeapostrophe, line 179: I see this increment, but not in the compiler?

jeapostrophe (changes requested):

js/stdlib/ts/ALGO.ts
erer1243: forgot to commit it :crying_cat_face:

jeapostrophe (approved):

PR #1361, opened 2022-08-23
Reverse Learning Optimization & Be Sensitive To The Untouchable Vars

### LINKED ISSUE #1183: The Solidity compiler failed with the message: Error: Invalid type for argument in function call. Invalid implicit conversion from struct T32 memory to struct T36 memory.
I have a bunch of `parallelReduce`s (6 in a row). The last two are very simple, it is 

```
  const [needToConfiscate] = parallelReduce([!fullUnlock])
    .invariant(true)
    .while(needToConfiscate)
    .api_(Api.confiscate, () => {
      check(this == metaLabsAddress);

      return [
        (callback) => {
          callback(null);
          transfer([[balance(token), token]]).to(this);
          return [false];
        },
      ];
    });

  const [needToPay] = parallelReduce([!fullUnlock])
    .paySpec([wALGO])
    .invariant(true)
    .while(needToPay)
    .api_(Api.payAndRelease, (amount) => {
      check(this == metaLabsAddress);

      return [
        [0, [amount, wALGO]],
        (callback) => {
          callback(null);
          return [false];
        },
      ];
    })
 ```
 
 It causes
 ```
The Solidity compiler failed with the message:
  
Error: Invalid type for argument in function call. Invalid implicit conversion from struct T32 memory to struct T36 memory requested.
    --> /tmp/reachc-sol-da154d2a4bf5dd20/compiled.sol:1691:10:
     |
1691 |       l6(la);
     |          ^^
     
CallStack (from HasCallStack):
  error, called at src/Reach/Util.hs:65:3 in reach-0.1.11-CbDshw3EQf7JLMU6TWlPrY:Reach.Util
  impossible, called at src/Reach/Connector/ETH_Solidity.hs:1867:13 in reach-0.1.11-CbDshw3EQf7JLMU6TWlPrY:Reach.Connector.ETH_Solidity
```

It is the initialization of `needToPay` with `!fullUnlock` in the second `parallelReduce` which causes the issue (if I change it to `parallelReduce([true])`, it compiles). What is weird that I have the same initialization of `needToConfiscate` in the first `parallelReduce` and it doesn't cause problems.

I hope that this information is enough to debug, if not please tell me and I will try to provide more details.

chrisnevers:
Fixes #1183

### DIFF
--- a/hs/src/Reach/Optimize.hs
+++ b/hs/src/Reach/Optimize.hs
@@ -16,6 +16,7 @@ import Reach.UnrollLoops
 import Reach.Util
 import Safe (atMay)
 import qualified Data.ByteString as B
+import Reach.CollectSvs
 
 type App = ReaderT Env IO
 
@@ -43,6 +44,7 @@ data CommonEnv = CommonEnv
   , cePrev :: M.Map DLExpr RepeatedT
   , ceNots :: M.Map DLVar DLArg
   , ceKnownLargeArgs :: M.Map DLVar DLLargeArg
+  , ceExpr :: M.Map DLVar DLExpr
   }
 
 instance Show CommonEnv where
@@ -55,12 +57,13 @@ instance Semigroup CommonEnv where
       , cePrev = g cePrev
       , ceNots = g ceNots
       , ceKnownLargeArgs = g ceKnownLargeArgs
+      , ceExpr = g ceExpr
       }
     where
       g f = f x <> f y
 
 instance Monoid CommonEnv where
-  mempty = CommonEnv mempty mempty mempty mempty
+  mempty = CommonEnv mempty mempty mempty mempty mempty
 
 data Env = Env
   { eFocus :: Focus
@@ -72,6 +75,7 @@ data Env = Env
   , eMaps :: DLMapInfos
   , eClearMaps :: Bool
   , eSimulate :: Bool
+  , eSvs :: S.Set DLVar
   }
 
 updateClearMaps :: Bool -> Env -> Env
@@ -161,7 +165,7 @@ recordKnownLargeArg dv v =
 
 remember_ :: Bool -> DLVar -> DLExpr -> App ()
 remember_ always v e =
-  updateLookup (\cenv -> cenv {cePrev = up $ cePrev cenv})
+  updateLookup (\cenv -> cenv {cePrev = up $ cePrev cenv, ceExpr = M.insert v e $ ceExpr cenv })
   where
     up prev =
       case always || not (M.member e prev) of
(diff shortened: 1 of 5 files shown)


hs/src/Reach/Optimize.hs
```diff
@@ -520,6 +524,49 @@ instance Extract (Maybe DLVar) where
...
+    DLA_Var dv -> [(dv, result)]
+    _ -> []
+
+instance Learn DLExpr where
+  learn result = \case
+    DLE_PrimOp _ IF_THEN_ELSE [DLA_Var dv, lhs, rhs]
+      -- r = true, x ? true : true  => []
+      -- r = v2  , x ? v2 : y       => [(x, true)]
+      | lhs == result && rhs /= result -> [(dv, true)]
+      | rhs == result && lhs /= result -> [(dv, false)]
```
jeapostrophe, line 542: `==` may be too specific and you should use `equiv` instead (because of source location / metadata)

hs/src/Reach/Optimize.hs
```diff
@@ -629,11 +675,12 @@ optLet at x e = do
...
                     return ()
                 meh
+  svs <- asks eSvs
   case (extract x, (isPure e && canDupe e), e) of
-    (Just dv, True, _) -> doit dv
+    (Just dv, True, _) | not (S.member dv svs) -> doit dv
     -- Optimize arithmetic even if it is impure (PV_Safe).
     -- It may trap as an effect, but we don't need to worry about it happening multiple times
-    (Just dv, _, DLE_PrimOp {}) -> doit dv
+    (Just dv, _, DLE_PrimOp {}) | not (S.member dv svs) -> doit dv
```
jeapostrophe, line 683: Name `not (S.member dv svs)` a function like `allowedToRemove` to make it intuitive

hs/src/Reach/Optimize.hs
```diff
@@ -520,6 +524,49 @@ instance Extract (Maybe DLVar) where
 allTheSame :: (Eq a, Sanitize a) => [a] -> Either (Maybe (a, a)) a
 allTheSame = allEqual . map sani
 
+-- Given that the result of `a` is `DLArg`, learn new bindings
```
jeapostrophe, line 527: This is cool

hs/src/Reach/Optimize.hs
```diff
@@ -529,8 +576,7 @@ optIf mkDo mkIf at c t f =
       -- XXX We could see if c' is something like `DLVar x == DLArg y` and add x -> y to the optimization environment. A general way to do this would be to have something that says "If this variable were True, you'd know _this_. So `z = x && y` would say `z => x` and `z => y`. And `x == y` would say `x = y`.
```
jeapostrophe, line 576: I think you did this, so you can remove the comment

jeapostrophe (changes requested):
Can you put as part of the PR an example of the before & after for the optimization relative to the learning thing?

chrisnevers:
There's not much interesting logic in this example program, but this shows it working:
```JavaScript
const v2145 : Bool! = value/2131 > lastConsensusTime/2137;
const v2146 : Bool! = (fullUnlock/2135 ? false : true);
const v2147 : Bool! = (v2145 ? v2146 : false);
if v2147 then {
  // Before
  (from 13, (continue [..., (fullUnlock/2135 : Bool, fullUnlock/2135), ... ]))
  // After
  (from 13, (continue [..., (fullUnlock/2135 : Bool, false), ... ]))
}
else { ... }
```

Since `v2147` is `true`, we know that `v2145` and `v2146` must be `true` because the other branch is `false`. If `v2146` is `true`, then `fullUnlock` must be `false`

jeapostrophe (approved):

PR #1378, opened 2022-08-29
CORE-2250: Add `Bytes.fromHex`

chrisnevers:


examples/bytesFromHex/index.rsh
```diff
@@ -0,0 +1,17 @@
...
+  const A = Participant('Alice', {
+    b4: Bytes(4)
+  });
+  init();
+
+  A.only(() => {
+    const b4 = declassify(interact.b4);
+  });
+  A.publish(b4).check(() => {
+    check(b4 == Bytes(4).fromHex('0x01ffc9a7'), "is equals");
```
jeapostrophe, line 13: This should be Bytes.fromHex and it would translate into four bytes

examples/bytesFromHex/index.mjs
```diff
@@ -0,0 +1,21 @@
+import {loadStdlib} from '@reach-sh/stdlib';
+import * as backend from './build/index.main.mjs';
+const stdlib = loadStdlib(process.env);
+const { ethers } = stdlib;
+
+const b4 = '0x01ffc9a7';
```
jeapostrophe, line 6: This is not really the value though... why not the actual byte sequence?

hs/src/Reach/AST/SL.hs
```diff
@@ -828,6 +828,7 @@ data SLPrimitive
   | SLPrim_Contract_new
   | SLPrim_Contract_new_ctor DLContractNews
   | SLPrim_toStringDyn
+  | SLPrim_Bytes_fromHex Integer
```
jeapostrophe, line 831: I don't think the Int is necessary because this always results in a static bytes

hs/src/Reach/Backend/JS.hs
```diff
@@ -258,7 +260,11 @@ jsLargeArg = \case
     return $ jsArray [jsString vn, vv']
   DLLA_Struct kvs ->
     jsLargeArg $ DLLA_Obj $ M.fromList kvs
-  DLLA_Bytes b -> return $ jsBytes b
+  DLLA_Bytes b -> do
+    case BS.decodeUtf8' b of
```
jeapostrophe, line 264: Why not fix jsBytes?

hs/src/Reach/Eval/Core.hs
```diff
@@ -1444,7 +1445,8 @@ evalAsEnvM sv@(lvl, obj) = case obj of
   SLV_Type (ST_Bytes len) -> do
     return $ Just $
       M.fromList
-        [("pad", retV $ public $ SLV_Prim $ SLPrim_padTo len)]
+        [ ("pad", retV $ public $ SLV_Prim $ SLPrim_padTo len)
+        , ("fromHex", retV $ public $ SLV_Prim $ SLPrim_Bytes_fromHex len)]
```
jeapostrophe, line 1449: If you added the number because you thought you couldn't add keys... you can add keys to any slprim

hs/src/Reach/Eval/Error.hs
```diff
@@ -168,6 +168,8 @@ data EvalError
   | Err_JSON SLValTy
   | Err_ContractCode ConnectorName String SLValTy
   | Err_Invalid_Exponential_Form String
+  | Err_BytesFromHex_WrongLength Integer String
```
jeapostrophe, line 171: I think this is just a subset of invalid

js/stdlib/ts/ETH_like_compiled.ts
```diff
@@ -140,11 +140,17 @@ const T_Bytes = (len:number): ETH_Ty<CBR_Bytes, ETH_Bytes> => {
   const me = {
     ...CBR.BT_Bytes(len),
     munge: ((bv: CBR_Bytes): ETH_Bytes => {
-      const bs = Array.from(ethers.utils.toUtf8Bytes(bv));
+      const ubs = bv.startsWith('0x')
```
jeapostrophe, line 143: If you do this, you have to update the docs to say the value will be 0x'd if it isn't utf8, but that is backwards incompatible

jeapostrophe (changes requested):

examples/bytesFromHex/index.mjs
chrisnevers: By byte sequence do you mean `[1, 255, 201, 167]`, `\x01\xff\xc9\xa7`, or something else?

jeapostrophe: The second

chrisnevers: javascript turns the hexadecimal characters in the second ones into utf-16 characters and cannot be munged properly, i.e produce 2 bytes instead of 1

jeapostrophe: There has to be a way to represent raw bytes

jeapostrophe: Google says JS defines strings as utf16 so maybe we need the cbr of bytes to always be something else?

jeapostrophe: That's kind of what you did but I think the rule you did is too adhoc

hs/src/Reach/Backend/JS.hs
```diff
@@ -264,7 +266,10 @@ jsLargeArg = \case
   DLLA_StringDyn t -> return $ jsString $ T.unpack t
 
 jsBytes :: B.ByteString -> Doc
-jsBytes = jsString . bunpack
+jsBytes b =
+  case BS.decodeUtf8' b of
+    Left _ -> jsApply "stdlib.bytesFromHex" [dquotes $ "0x" <> pretty (B16.encodeBase16 b)]
```
jeapostrophe, line 271: This requires a version bump

js/stdlib/ts/CBR.ts
```diff
@@ -16,7 +16,7 @@ export const bigNumberToNumber = (x: any) =>
 export type CBR_Null = null;
 export type CBR_Bool = boolean;
 export type CBR_UInt = BigNumber;
-export type CBR_Bytes = string;
+export type CBR_Bytes = Uint8Array | string;
```
jeapostrophe, line 19: This needs to be documented in https://docs.reach.sh/frontend/#ref-frontends-js-types

jeapostrophe: Make sure you say something about when it is one versus the other

jeapostrophe (changes requested):
I did not study the CI failures

jeapostrophe (changes requested):
I accidentally made a comment directly on the commit rather than here

js/stdlib/ts/CBR.ts
```diff
@@ -104,20 +104,24 @@ export const BT_Bytes = (len: number): BackendTy<CBR_Bytes> => ({
...
+      const lenn = bigNumberToNumber(len);
+      const checkLen = (label:string, alen:number, fill:string): string => {
+        const v = val as string;
+        if ( v.length > alen ) {
+          throw Error(`Bytes(${len}) must be a ${label}string less than or equal to ${alen}, but given ${label}string of length ${v.length}`);
+        }
+        return v.padEnd(alen, fill);
+      };
+      if ( val.slice(0,2) === '0x' ) {
+        return checkLen('hex ', lenn*2+2, '0');
```
jeapostrophe, line 117: This rule was always there, but we should add a note to the docs that this is a special case OR we should drop it because there's now the `Uint8Array` option

js/stdlib/ts/ETH_like_compiled.ts
```diff
@@ -140,15 +141,17 @@ const T_Bytes = (len:number): ETH_Ty<CBR_Bytes, ETH_Bytes> => {
...
     }),
     unmunge: ((nvs: ETH_Bytes): CBR_Bytes => {
-      const go = (nv:any) => hexToString(ethers.utils.hexlify(unBigInt(nv)))
-      const nvs_s = (len <= byteChunkSize) ? go(nvs) : nvs.map(go);
-      const nvss = "".concat(...nvs_s);
-      // debug(me.name, nvs, nvss);
-      return me.canonicalize(nvss);
+      const go = (nv:any, i: number) => {
+        const r = ethers.utils.hexlify(unBigInt(nv));
+        return (i > 0 && r.startsWith('0x')) ? r.slice(2) : r;
```
jeapostrophe, line 150: This `0x` thing here makes this function very different than before and it doesn't seem to acknowledge the ui8a vs not thing. Maybe a better way to implement this function is to turn everything into ui8a uniformly and then map or not map?

jeapostrophe (changes requested):

js/stdlib/ts/ETH_like_compiled.ts
chrisnevers: What do you mean by acknowledge the ui8a vs not? It will still produce ui8a's if the bytes are not utf-8 via the `hexToString` call below. I kind of just changed the order of operations from: calling `hexToString` on every element and joining => joining every element and calling `hexToString`. This is to ensure that every element of an array is of the same type, e.g. so we don't get `[string, string, ui8a, string]` for a long `bytes` value that isn't utf-8

jeapostrophe: Okay, I think I understand now

jeapostrophe (changes requested):
Looks good. The vsce keyword file needs to be updated and then it can merge

jeapostrophe (approved):

PR #1379, opened 2022-08-30
makeSeal

willghatch:
This isn't finished, but I thought I might as well start a draft PR.  I'll re-summarize my comments to you from slack:

This feature improves error messages for cases where the sealed type is expected, but not for cases where the sealed type is unexpectedly used.  Changing that would require changes to type checking all over Eval/Core, I think, and so would be difficult.  Also, while this gives a better error message, I think we need more features to make the new sealed error messages feel cohesive.  Eg. if you want to have an array of fixed points or pass fixed points through an interact interface or API, you still need to write the object type, you can't just write FixedPoint.


jeapostrophe:
Yes, we are abandoning this, but hopefully this history will be trackable.

PR #1382, opened 2022-08-31
Raw remote

willghatch:
Boom, as Jay might say.


docs/src/rsh/consensus/index.md
```diff
@@ -615,6 +615,7 @@ In addition, a remote function may be augmented with one of the following operat
     The minimum balance will decrease on a `CloseOut` or `ClearState` transaction.
   + `{!rsh} opts.strictPay` does not optimize away `pay` or `axfer` transactions that are statically zero.
     If this is needed, and not included, then `{!rsh} remote` calls that require a payment transaction to always be present, even if the amount is zero, could fail.
+  + `{!rsh} opts.rawCall` is a bool (default false) that when true omits the method selector from the call.
```
jeapostrophe, line 618:
```suggestion
  + `{!rsh} opts.rawCall` is a boolean (default `{!rsh} false`) that when `{!rsh} true` omits the ABI method selector from the call.
```

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2227,8 +2227,9 @@ ce = \case
           ca ro
           incResource R_App ro
           makeTxn1 "ApplicationID"
-          cbs $ sigStrToBytes sig
-          makeTxn1 "ApplicationArgs"
+          when (not r_rawCall) $ do
```
jeapostrophe, line 2230:
```suggestion
          unless r_rawCall $ do
```

jeapostrophe (changes requested):

willghatch:
Probably because of that annoying VSCE generated file check.

PR #1344, opened 2022-08-17
CORE-2237

jeapostrophe:

PR #1392, opened 2022-09-01
Erc721 testing

willghatch:
This draft PR applies the ERC721 test to our ERC721 contract, but:
• Certain important parts of the contract have been commented out and need to be re-enabled before we can declare it to work.
• Some failing tests have been disabled, and need to be re-enabled and passed.

I'm happy to work on those, of course, but I'm putting this draft PR here because it fulfills the original task of “applying the test suite to the contract”, and fixing the contract to pass the test suite is more scope creep to the card.


jeapostrophe (changes requested):
I was hoping you would extend https://github.com/reach-sh/reach-lang/pull/1344 by moving files around. It makes it harder to see the changes you made this way.

For the checks/code that failed, can you post the error msg?

I also don't understand what `jay-small-test.mjs` is?

willghatch:
Good point, I've pushed it there.  `jay-small-test.mjs` is the test file that you wrote.  I wasn't quite sure what to do with it after applying Oliver's test file, but I wanted to put all the files in one directory.

willghatch:
As for failed checks, there are basically 3 things:

• Code that relies on Bytes.fromHex
• Tests of the ERC721Enumerable interface, that are not actually implemented in the contract
• I have yet to get transferring working with the check that `this` is the owner or authorized, and I haven't figured out what's wrong yet.  But if I uncomment that check function then an assert fails.

The test file doesn't test the transfer function with an extra argument, or burning, or some other cases that we should probably test before we use them “in the wild”.

PR #1393, opened 2022-09-02
CORE-2222 Change some stdlib uses of Account to Address

erer1243:
putting this up to see which examples will fail


js/stdlib/ts/ALGO.ts
```diff
@@ -1217,17 +1217,16 @@ const makeTransferTxn = (
...
 const transfer = async (
   from: Account,
-  to: Account,
+  to: Address,
   value: any,
   token?: Token,
   opts?: TransferOpts,
 ): Promise<RecvTxn> => {
   const sender = from.networkAccount;
-  const receiver = extractAddr(to);
```
jeapostrophe, line 1226: You should still use this function, because we want to allow people to pass `Account`s, since they can be turned into addresses

js/stdlib/ts/ETH_like.ts
```diff
@@ -403,24 +403,23 @@ const doCall = async (
 /** @description Arg order follows "src before dst" convention */
 const transfer = async (
   from: AccountTransferable,
-  to: AccountTransferable,
+  to: Address,
```
jeapostrophe, line 406: Same comment

jeapostrophe (changes requested):

jeapostrophe (changes requested):
This looks great. I think the only thing that needs to change is the docs.

PR #1398, opened 2022-09-02
CORE-2261: ERC-165

chrisnevers:


examples/erc165/index.mjs
```diff
@@ -0,0 +1,40 @@
+import { loadStdlib } from "@reach-sh/stdlib";
```
jeapostrophe, line 1: This is fine for 2261, but 224[789] should use ethers directly like the other tests

jeapostrophe (approved):

PR #1404, opened 2022-09-06
Contract from address

willghatch:
Working on `Contract.fromAddress`, I'm somewhat stuck writing the ETH connector code, and I think this will be much easier for @jeapostrophe to answer than for me to figure out.  In writing the `solExpr` implementation, I need to conditionally do some side effects to create a Data large arg, then return that large arg.  But this will be in an expression context, so I can't just use these side effect statements there.  I want to lift these statements into a function (and lift the function only once), but I'm not sure how to do that in the context of `solExpr`.  So do you have any suggestions, or am I missing something?

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -633,6 +633,31 @@ const [ netRecv, [gilRecv, zmdRecv], randomValue ] =
 might be the way to communicate with a random oracle that receives a conservative approximation of its actual cost and returns what it does not use, along with some amount of network tokens, `GIL`, and `ZMD`.
 This operation may not be used with `{!rsh} REMOTE_FUN.bill`.
 
+### Contract.fromAddress
+@{ref("rsh", "Contract.fromAddress")}
+
+Takes an `{!rsh} Address}` and returns a `{!rsh} Maybe(Contract)`.
+
+On Algorand this always returns `{!rsh} None`.
+On Ethereum and Conflux, it returns `{!rsh} None` for addresses of externally owned accounts, and it returns `{!rsh} Some` for addresses of contracts under most circumstances.
+However, it also returns `{!rsh} None` for addresses of contracts that are under construction, for addresses of contracts that have been destroyed, and for addresses of contracts that have not yet been created.
+
+```reach
+bob.publish(addr);
+const ctcMaybe = Contract.fromAddress(addr);
+const result = ctcMaybe.match({
+  Some: (ctc) => {
+    const r = remote( ctc, {
+      f: Fun([UInt], UInt),
+    });
+    return r.f(5);
+  },
+  None: () => {
+    return 0;
+  },
+});
+```
+
 ### Mappings: creation and modification
 
 @{ref("rsh", "Map")}
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -636,27 +636,16 @@ This operation may not be used with `{!rsh} REMOTE_FUN.bill`.
 ### Contract.fromAddress
 @{ref("rsh", "Contract.fromAddress")}
 
+```reach
+const ctcMaybe = Contract.fromAddress(address);
+```
+
 Takes an `{!rsh} Address}` and returns a `{!rsh} Maybe(Contract)`.
 
 On Algorand this always returns `{!rsh} None`.
 On Ethereum and Conflux, it returns `{!rsh} None` for addresses of externally owned accounts, and it returns `{!rsh} Some` for addresses of contracts under most circumstances.
 However, it also returns `{!rsh} None` for addresses of contracts that are under construction, for addresses of contracts that have been destroyed, and for addresses of contracts that have not yet been created.
 
-```reach
-bob.publish(addr);
-const ctcMaybe = Contract.fromAddress(addr);
-const result = ctcMaybe.match({
-  Some: (ctc) => {
-    const r = remote( ctc, {
-      f: Fun([UInt], UInt),
-    });
-    return r.f(5);
-  },
-  None: () => {
-    return 0;
-  },
-});
-```
 
 ### Mappings: creation and modification
 
--- a/hs/src/Reach/Backend/JS.hs
+++ b/hs/src/Reach/Backend/JS.hs
@@ -627,6 +627,8 @@ jsExpr = \case
     mo' <- jsArg mo
     da' <- jsArg da
     return $ jsApply "stdlib.fromSome" [mo', da']
+  DLE_ContractFromAddress _at _addr -> do
+    impossible "TODO - implement ContractFromAddress"
   DLE_ContractNew at cns dr -> do
     (ctxt_mode <$> ask) >>= \case
       JM_Backend -> return "undefined /* ContractNew */"
--- a/hs/src/Reach/Backend/JS.hs
+++ b/hs/src/Reach/Backend/JS.hs
@@ -628,7 +628,7 @@ jsExpr = \case
     da' <- jsArg da
     return $ jsApply "stdlib.fromSome" [mo', da']
   DLE_ContractFromAddress _at _addr -> do
-    impossible "TODO - implement ContractFromAddress"
+    return $ jsArray [jsString "None", "null"]
   DLE_ContractNew at cns dr -> do
     (ctxt_mode <$> ask) >>= \case
       JM_Backend -> return "undefined /* ContractNew */"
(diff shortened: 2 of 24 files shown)


hs/src/Reach/Backend/JS.hs
```diff
@@ -627,6 +627,8 @@ jsExpr = \case
     mo' <- jsArg mo
     da' <- jsArg da
     return $ jsApply "stdlib.fromSome" [mo', da']
+  DLE_ContractFromAddress _at _addr -> do
+    return $ jsArray [jsString "None", "null"]
```
jeapostrophe, line 631: This can't be the implementation. Either it needs to do the real thing or it needs to log the answer and then look it up

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2413,6 +2413,9 @@ ce = \case
     -- [ Default, Object, Tag ]
     -- [ False, True, Cond ]
     op "select"
+  DLE_ContractFromAddress _at _addr -> do
+    size <- typeSizeOf $ T_Data (M.fromList [("None", T_Null), ("Some", T_Address)])
```
jeapostrophe, line 2417: First, you should can cla. Second, there's a function for constructing a Maybe DLType in DLBase

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -795,6 +796,17 @@ solExpr sp = \case
     return $ parens $ c <+> "?" <+> (mo' <> "._" <> pretty vn) <+> ":" <+> da'
   DLE_GetUntrackedFunds {} -> impossible "getUntrackedFunds"
   DLE_ContractNew {} -> impossible "contractNew"
+  DLE_ContractFromAddress at addr -> do
```
jeapostrophe, line 799: You should make this a call to a function in the sol stdlib. Then make Maybe Address a built in type so it can be returned directly OR catch this call at the DLLet level and make it so you can assign the fields directly

hs/src/Reach/Simulator/Core.hs
```diff
@@ -703,6 +703,8 @@ instance Interp DLExpr where
       setGlobal $ g {e_nctc = ctcId}
       -- XXX make an interact point with dr
       return $ V_Contract $ fromIntegral ctcId
+    DLE_ContractFromAddress _at _addr -> do
+      return $ V_Data "None" V_Null
```
jeapostrophe, line 707: You should make a comment here that a better impl is to make it a choice OR consult the table of apps

jeapostrophe (changes requested):
Don't forget docs and change log

willghatch:
I'm having an issue with the solidity backend generating two copies of the maybe address type, and I'm not sure why.  IE I'm getting this error:

```
--> build/index.main.sol

Error: Type enum _enum_T2 is not implicitly convertible to expected type enum _enum_T3.
   --> build/index.main.sol:772:22:
    |
772 |       _f.v91.which = _enum_T2.Some;
    |                      ^^^^^^^^^^^^^

Error: Type enum _enum_T2 is not implicitly convertible to expected type enum _enum_T3.
   --> build/index.main.sol:776:22:
    |
776 |       _f.v91.which = _enum_T2.None;
    |                      ^^^^^^^^^^^^^

Error: Type enum _enum_T2 is not implicitly convertible to expected type enum _enum_T3.
   --> build/index.main.sol:823:23:
    |
823 |       _f.v106.which = _enum_T2.Some;
    |                       ^^^^^^^^^^^^^

Error: Type enum _enum_T2 is not implicitly convertible to expected type enum _enum_T3.
   --> build/index.main.sol:827:23:
    |
827 |       _f.v106.which = _enum_T2.None;
    |                       ^^^^^^^^^^^^^
```

Where T2 and T3 (and their _enum_TX counterparts) are basically duplicates.

hs/src/Reach/Connector/ALGO.hs
```diff
@@ -2413,6 +2413,12 @@ ce = \case
     -- [ Default, Object, Tag ]
     -- [ False, True, Cond ]
     op "select"
+  DLE_ContractFromAddress _at _addr -> do
+    case maybeT T_Address of
+      T_Data maybeMap -> do
+        let la = DLLA_Data maybeMap "None" $ DLA_Literal DLL_Null
+        cla la
+      _ -> impossible "maybe not maybe"
```
jeapostrophe, line 2421: This is `cla $ mdaToMaybeLA T_Address Nothing`

https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/AST/DLBase.hs#L467-L468

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1060,6 +1062,20 @@ solCom = \case
       [ solSet ("uint256" <+> actBalV) bal
       , solSet (solMemVar dv) ite
       ]
+  DL_Let _ (DLV_Let _ dv) (DLE_ContractFromAddress _at addr) -> do
+    maybeMap <- case maybeT T_Address of
+      T_Data mm -> return mm
+      _ -> impossible "maybe not maybe"
```
jeapostrophe, line 1068: `dataTypeMap`

jeapostrophe: But I think I'd use `mdaToMaybeLA` twice

hs/src/Reach/Eval/Core.hs
```diff
@@ -4098,6 +4099,19 @@ evalPrim p sargs =
...
+      x <- one_arg
+      (xt, xa) <- compileTypeOf x
+      case xt of
+        T_Address -> do
+          let mkv = DLVar at Nothing $ maybeT T_Contract
+          let e = DLE_ContractFromAddress at xa
+          fsv <- ctxt_lift_expr mkv e
+          fsv' <- doInternalLog_ Nothing fsv
+          return (lvl, SLV_DLVar fsv')
+        _ -> expect_t x $ Err_Expected "Address"
```
jeapostrophe, line 4114: This is a use of `compileCheckType`

hs/src/Reach/Connector/ETH_Solidity.hs
```diff
@@ -1060,6 +1062,20 @@ solCom = \case
       [ solSet ("uint256" <+> actBalV) bal
       , solSet (solMemVar dv) ite
       ]
+  DL_Let _ (DLV_Let _ dv) (DLE_ContractFromAddress _at addr) -> do
+    maybeMap <- case maybeT T_Address of
```
jeapostrophe, line 1066: This should be `T_Contract`, that's the source of your struct mismatch error

jeapostrophe (changes requested):

hs/src/Reach/Connector/ETH_Solidity.hs
willghatch: Oof, yes, that's pretty obvious in hindsight, thanks.

hs/src/Reach/Connector/ALGO.hs
willghatch: What does `c` stand for in all of these functions?

docs/src/rsh/consensus/index.md
```diff
@@ -633,6 +633,31 @@ const [ netRecv, [gilRecv, zmdRecv], randomValue ] =
...
+### Contract.fromAddress
+@{ref("rsh", "Contract.fromAddress")}
+
+Takes an `{!rsh} Address}` and returns a `{!rsh} Maybe(Contract)`.
+
+On Algorand this always returns `{!rsh} None`.
+On Ethereum and Conflux, it returns `{!rsh} None` for addresses of externally owned accounts, and it returns `{!rsh} Some` for addresses of contracts under most circumstances.
+However, it also returns `{!rsh} None` for addresses of contracts that are under construction, for addresses of contracts that have been destroyed, and for addresses of contracts that have not yet been created.
+
+```reach
```
jeapostrophe, line 645: We try to not have samples directly in the documentation, but instead, use sampling

docs/src/rsh/consensus/index.md
```diff
@@ -633,6 +633,31 @@ const [ netRecv, [gilRecv, zmdRecv], randomValue ] =
 might be the way to communicate with a random oracle that receives a conservative approximation of its actual cost and returns what it does not use, along with some amount of network tokens, `GIL`, and `ZMD`.
 This operation may not be used with `{!rsh} REMOTE_FUN.bill`.
 
+### Contract.fromAddress
+@{ref("rsh", "Contract.fromAddress")}
+
```
jeapostrophe, line 638: The style is to start with a small sample to just show the syntax

jeapostrophe (changes requested):

docs/src/rsh/consensus/index.md
willghatch: I think the most reasonable sample might be the ERC721 contract.  What say we sample that once it's merged?

PR #1407, opened 2022-09-08
CORE-2307 shorten algorand truncation check

erer1243:
todo: example proving it works

### DIFF
--- a/hs/src/Reach/Connector/ALGO.hs
+++ b/hs/src/Reach/Connector/ALGO.hs
@@ -1429,7 +1429,7 @@ cprim = \case
           -- [ v ]
           let ext i = cint (8 * i) >> op "extract_uint64"
           unless (trunc || pv == PV_Veri) $ do
-            op "// Truncation check"
+            comment "Truncation check"
             op "dup"
             op "bitlen"
             cint 64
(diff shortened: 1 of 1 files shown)


hs/src/Reach/Connector/ALGO.hs
```diff
@@ -1429,12 +1429,12 @@ cprim = \case
           -- [ v ]
           let ext i = cint (8 * i) >> op "extract_uint64"
           unless (trunc || pv == PV_Veri) $ do
-            dupn 3
-            -- [ v, v, v, v ]
-            let go i = ext i >> cint 0 >> asserteq
-            go 2
-            go 1
-            go 0
+            op "// Truncation check"
```
jeapostrophe, line 1432: There's a function called `comment` for this

jeapostrophe (changes requested):
Re: an example --- I think this already does it: https://github.com/reach-sh/reach-lang/blob/master/examples/uint256/index.mjs#L73

erer1243:
you're right, the uint256 example does it. so, no new example

jeapostrophe (approved):

erer1243:
this fails some of the closed code. do you want me to take a stab at that or just leave it to someone else?

jeapostrophe:
Fix it, I messaged how in Slack

PR #1408, opened 2022-09-09
CORE-2304 allow examples to be marked and run by themselves

erer1243:
empty pr so i can play with circleci; actual changes coming later


.circleci/generate-ci.py
```diff
@@ -0,0 +1,84 @@
...
+def connector(executor, connector, parallelism):
+  examples_name = f"examples.{connector}"
+  browser_tests_name = f"browser-tests.{connector}"
+  devnet_name = f"devnet-{connector.lower()}"
+
+  build_image(executor, devnet_name, [])
+  add_build_job("examples", {
+    "name": examples_name,
+    "connector": connector,
+    "size": parallelism,
```
jeapostrophe, line 50: This is not the way the other one worked. The other one did `total / 16` and you are doing `16`. This is very different.

jeapostrophe (changes requested):
This is a scandalous PR and if you do it, you should totally remove `make.sh` so it isn't confusing.

FWIW, the way to actually do this task is to change this line:

https://github.com/reach-sh/reach-lang/blob/master/.circleci/example.sh#L71

L71 basically says "Tell me all of the test jobs and split them up by `size`" (from L50), so if there were 100 tasks and size was 5 then they will be bunched into groups of 20 (5 x 20 = 100).

What you're to do is to have a way to indicate that some examples need to run completely alone. So suppose that there

```
total = 100 // ls -1 examples | wc -l
solo = 10 // wc -l examples/solos
para = total - solo = 90
per = 16 // hardcoded
```

and then you're going to make

```
size = solo + para / per
```

And you're going to use change L71 so that the solo are excluded and you can't use `circleci tests split` anymore because I don't know how to change what it thinks the level is. So you'll basically need to implement it yourself by looking at CIRCLE_NODE_INDEX and CIRCLE_NODE_TOTAL and testing if index <= solo (so it should run the index'd solo test) or if it is more and it should run the `index - solo`'d `para/per` group of para tasks

https://circleci.com/docs/parallelism-faster-jobs
https://circleci.com/docs/env-vars#built-in-environment-variables

erer1243:
Hey, pr's not in review column!! Stop looking!! It'll be good at the end

.circleci/config.pre.yml
```diff
@@ -257,7 +257,7 @@ jobs:
         mkdir -p /tmp/workspace/artifacts /tmp/test_results /tmp/workspace/record /tmp/artifacts
     - run:
         command: |
-          cd .circleci && ./example.sh <<parameters.connector>> "${CIRCLE_NODE_TOTAL}" "${CIRCLE_NODE_INDEX}"
+          cd .circleci && ./run-examples.py <<parameters.connector>>
```
jeapostrophe, line 260: The reason I originally had it the other way was so I could test it without setting the environment variables

.circleci/config.pre.yml
```diff
@@ -277,14 +277,14 @@ jobs:
...
     # Regular test failures go through to the sink.
-    - slack/notify:
-        event: fail
-        custom: |
-          { "blocks": [
-              { "type": "section",
-                "text": {
-                  "type": "mrkdwn",
-                  "text": "*CATASTROPHIC* ${CIRCLE_USERNAME}/<https://github.com/reach-sh/reach-lang/tree/${CIRCLE_BRANCH}|${CIRCLE_BRANCH}>${CIRCLE_PULL_REQUEST:+/<$CIRCLE_PULL_REQUEST|PR>} > examples.<<parameters.connector>> <${CIRCLE_BUILD_URL}|more...>" }}]}
+    # - slack/notify:
```
jeapostrophe, line 280: Reminder to revert before merging

.circleci/special-examples.txt
```diff
@@ -0,0 +1,5 @@
+# Special examples are run by themselves on their own parallel runner by CircleCI
```
jeapostrophe, line 1: The goal is to take all of the examples that have long timeouts and make them have short timeouts and run by themselves.

Look at a run like this --- https://app.circleci.com/pipelines/github/reach-sh/reach-lang/8703/workflows/258a6e3e-2452-4f99-a61a-1fa33c3b18c2 --- and then look at the ALGO job --- https://app.circleci.com/pipelines/github/reach-sh/reach-lang/8703/workflows/258a6e3e-2452-4f99-a61a-1fa33c3b18c2/jobs/155010 --- and see that there are some that take 7m and some that take 3.5m. My hypothesis is that the 7m ones are those long timeouts AND the tests are flaky.

.circleci/run-examples.py
```diff
@@ -0,0 +1,91 @@
...
+
+os.chdir("../examples")
+cmd("mkdir -p /tmp/artifacts /tmp/test_results /tmp/workspace/record")
+cmd("../reach devnet --await-background")
+
+# Generate what to pass to stdin for `circleci tests split`
+# The split command takes some lines, then magically splits the lines
+# between runners and prints which lines the current runner is assigned.
+# We pass it exactly the same number of lines as there are parallel runners,
+# so we get exactly one line back per runner.
```
jeapostrophe, line 69: It really isn't "magical" and basically what you're doing is...
```
(special, regular) = examples.split_examples_into_jobs(connector)
comb = special <> regular
comb[CIRCE_NODE_INDEX]
```
But I don't think your `special` is a list of lists, like regular is. My advice... change it so `split_examples_into_jobs` returns `[ [ Job ] ]` where the length is `parallel_jobs_for(conn)` and then you always `run_multiple_examples`, but sometimes the list just has one thing

jeapostrophe (changes requested):

erer1243:
it'll be ready for review when I unmark it as a draft.

.circleci/examples.py
```diff
@@ -52,9 +55,35 @@ def divide_list(lst, n):
...
+# This passes every special example on its own line,
+# and "job-group-<n>" for every set of regular examples.
+# Then, this decodes the line printed by the split command into a list
+# of examples to run, and returns that list.
+def circleci_split_examples(connector):
+  (special_examples, regular_examples) = split_examples_into_jobs(connector)
+  split_args = special_examples + [f"job-group-{n}" for n in range(len(regular_examples))]
+  assert len(split_args) == int(os.environ["CIRCLE_NODE_TOTAL"])
+  split_args_bytes = "\n".join(split_args).encode('utf-8')
+  circleci_split = subprocess.run(["circleci", "tests", "split"], input=split_args_bytes,
```
jeapostrophe, line 77: How come you're doing it the hard way with this extra process and text processing? I think you should do it via just selecting the right element of the return of `split_examples_into_jobs` by looking at `CIRCLE_NODE_INDEX`

.circleci/generate-ci.py
```diff
@@ -10,6 +10,14 @@
 config = yaml.load(Path("config.pre.yml"))
 build_sink_deps = []
 
+# Insert an entry into config.gen.yml in the build workflow:
```
jeapostrophe, line 13: You should update the Makefile to use this file rather than `make.sh` for doing local checking of the format

jeapostrophe (changes requested):

.circleci/examples.py
erer1243: I didn't think of doing it that way

.circleci/generate-ci.py
erer1243: I don't think make.sh is used anywhere

.circleci/examples.py
jeapostrophe: I mentioned it in an earlier comment, I guess it got delete.

.circleci/generate-ci.py
jeapostrophe: I meant this https://github.com/reach-sh/reach-lang/blob/master/.circleci/Makefile#L8-L9 but you fixed it. Sorry!

jeapostrophe (dismissed):

PR #1413, opened 2022-09-12
DEVX-291

TheChronicMonster:
NFT Tutorial Series 1 
stdlib minter that works with ALGO and ETH connectors


examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,39 @@
...
+const receiverAddrFormat = await stdlib.formatAddress(receiver);
+console.log(`The receiver's formatted address is ${receiverAddrFormat}`);
+
+const fmt = (x) => stdlib.formatCurrency(x, 4);
+const getBal = async (who) => fmt(await stdlib.balanceOf(who));
+const bal = await getBal(accMinter);
+console.log(`Minter starting balance: ${bal}.`);
+
+console.log(`Creating the NFT`);
+const theNFT = await stdlib.launchToken(accMinter, "nftName", "SYM", { supply: 1});
```
jeapostrophe, line 29: You should make sure that your example follows the appropriate ARC specs completely and would get shown on the standard galleries. Ideally your example/tutorial will make it easy to talk about things like traits and randomness

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,39 @@
...
+const nftId = theNFT.id;
+    
+await receiver.tokenAccept(nftId);
+if (stdlib.connector == 'ALGO') {console.log(`Receiver opted-in to NFT`)};
+receiver.tokenAccepted(nftId);
```
chrisnevers, line 34: I think this either isn't needed or you should use the result, e.g:
```javascript
if (stdlib.connector == 'ALGO' && receiver.tokenAccepted(nftId)) {console.log(`Token accepted`)};
```

PR #1415, opened 2022-09-12
CORE-2205: Remove Empty Switch In Dekont

### LINKED ISSUE #1325: Problem with the Reach compiler
## Problem Definition

I have encountered this error:

_reachc: The compiler has encountered an internal error:

  df_com switch (v113 : Data({"None": Null, "Some": Bytes(128)})) {
  case None as v114 : Null/True: {
    const eff = protect<Null>("Creator".interact.reportPosition(did/92, v113 ));
    exit(); }
  case Some as v115 : Bytes(128)/True: {
    const eff = protect<Null>("Creator".interact.reportPosition(did/92, v113 ));
    exit(); } }

This error indicates a problem with the Reach compiler, not your program. Please report this error, along with the pertinent program, to the Reach team as soon as possible so we can fix it.

Open an issue at: https://github.com/reach-sh/reach-lang/issues_

## Code

This is the code the generate the error.
In particular, the error appears if I add the **else**.  

```
const keepGoing = 
  parallelReduce(true) 
    .invariant(balance() == balance())
    .while(keepGoing)
    .api(attacherAPI.insert_position,
      (x, z, y) => { 
        y(z);
        if(isSome(easy_map[z])){ 
          return false;
        }else{
       map_my[z] = fromSome(map_my[z],x);
        return true;
        }

      }
    )
```

Thank you.

chrisnevers:
Fixes #1325


jeapostrophe (approved):

hs/src/Reach/Linearize.hs
```diff
@@ -113,7 +114,9 @@ dk1 k s =
           DKBM_Con -> return $ (con, k)
           DKBM_Do -> return $ (loc, mt)
       let cm1 (dv', b, l) = (,,) dv' b <$> dk_ k' l
-      mk <$> mapM cm1 csm
+      mapM cm1 csm >>= \case
```
jeapostrophe, line 117: This pattern feels weird to me. I think I'd just name it then do a test.

jeapostrophe:
It looks like tests failed?

jeapostrophe (dismissed):
Style is fine, but it also seems like nothing else changed from the failed CI?

chrisnevers:
The examples that failed on CI pass locally, I think it was a fluke

PR #1424, opened 2022-09-15
CORE-2289 add stdlib.contract

erer1243:


docs/src/frontend/index.md
```diff
@@ -833,13 +833,16 @@ The third thing you should do in a frontend is create a contract handle, so you
...
 ```js
-acc.contract(bin, ?info) => ctc
+acc.contract(bin: Backend, info?: Promise<ContractInfo>) => Contract
+stdlib.contract(bin: Backend, info?: Promise<ContractInfo>) => Promise<Contract>
 ```
 
-Returns a Reach contract handle based on the `{!js} bin` argument provided with access to the account `{!js} acc`.
+Returns a Reach contract handle based on the `{!js} bin` argument, and optional `ContractInfo` provided.
+The `{!js} acc.contract` form will use the account `{!js} acc` to interact with the returned contract.
+The `{!js} stdlib.contract` form is equivalent to `{!js} (await stdlib.createAccount()).contract`, generating an account on demand.
```
jeapostrophe, line 842:
```suggestion
The `{!js} stdlib.contract` form is equivalent to `{!js} (await stdlib.createAccount()).contract`, generating an account on demand.
(This means that the contract handle may not be used for non-read-only interaction.)
```

docs/src/frontend/index.md
```diff
@@ -853,16 +856,16 @@ This deployment can only happen one time, so subsequent attempts will fail with
 This function may emit warnings if there is any danger, risk, or subtlety to using this contract on your chosen consensus network.
 You can omit this warning by setting `{!cmd} REACH_NO_WARN`, but we recommend that you do not.
 
-This function does not block.
+The `acc.contract` form of the function does not block.
```
jeapostrophe, line 859:
```suggestion
The `{!js} acc.contract` form of the function does not block.
```

js/stdlib/ts/ALGO.ts
```diff
@@ -2707,6 +2707,9 @@ const launchToken = async (accCreator: Account, name: string, sym: string, opts:
   return { name, sym, id, mint, optOut };
 }
 
+const contract = (bin: Backend, ctcInfo?: Promise<ContractInfo>): Promise<Contract> =>
+  createAccount().then(acc => acc.contract(bin, ctcInfo));
```
jeapostrophe, line 2711: I think there should be something in `shared_impl` to make it easy to do this and keep them in all in sync

jeapostrophe (changes requested):

js/stdlib/ts/ALGO.ts
erer1243: i did this, but i don't think it's better

js/stdlib/ts/ALGO.ts
```diff
@@ -2730,6 +2731,6 @@ const launchToken = async (accCreator: Account, name: string, sym: string, opts:
     parseCurrency, minimumBalance, formatCurrency,
     reachStdlib, algosdk,
     connector, standardUnit, atomicUnit,
-    tokensAccepted,
-  };
+    tokensAccepted
```
jeapostrophe, line 2734: Heresy!

jeapostrophe (approved):

js/stdlib/ts/ALGO.ts
erer1243: hearsay

PR #1425, opened 2022-09-15
DEVX-291c

TheChronicMonster:


examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,43 @@
...
+    
+if (stdlib.connector == 'ALGO' && await receiver.tokenAccept(nftId)) {console.log(`Receiver opted-in to NFT`)};
+if (stdlib.connector == 'ALGO' && await receiver.tokenAccepted(nftId)) {console.log(`Token accepted`)};
+await stdlib.transfer(minter, receiver, 1, nftId);
+console.log(`NFT transfer made from minter to receiver`);
+
+const postBal = await getBal(minter);
+console.log(`Minter balance after transfer: ${postBal}.`);
+
+const [postAmtNFT] = await stdlib.balancesOf(receiver, [nftId]);
```
TheChronicMonster, line 42: @chrisnevers I feel like there should be a better way to call and display the NFT balance.

chrisnevers: You can do 
```javascript
const postAmtNFT = await receiver.balanceOf(nftId);
```

chrisnevers (dismissed):

examples/simple-nft-minter/index.rsh
```diff
@@ -0,0 +1 @@
+'reach 0.1'
```
jeapostrophe, line 1: This is going to need a `txt` file to pass CI

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,43 @@
...
+const receiverAddrFormat = await stdlib.formatAddress(receiver);
+console.log(`The receiver's formatted address is ${receiverAddrFormat}`);
+
+const fmt = (x) => stdlib.formatCurrency(x, 4);
+const getBal = async (who) => fmt(await stdlib.balanceOf(who));
+const bal = await getBal(minter);
+console.log(`Minter starting balance: ${bal}.`);
+
+console.log(`Creating the NFT`);
+const theNFT = await stdlib.launchToken(minter, "nftName", "SYM", { supply: 1, url: "https://reach.sh", clawback: null, freeze: null, defaultFrozen: false, reserve: null, note: Uint8Array[1]});
```
jeapostrophe, line 28: This is formatted ugly and it does not match the ARC specs for the metadata. You should confirm that whatever you do will show up in the testnet galleries.

jeapostrophe: You should make this have cute names that "feel" like a real NFT.

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,43 @@
...
+if (stdlib.connector == 'ALGO' && await receiver.tokenAccept(nftId)) {console.log(`Receiver opted-in to NFT`)};
+if (stdlib.connector == 'ALGO' && await receiver.tokenAccepted(nftId)) {console.log(`Token accepted`)};
+await stdlib.transfer(minter, receiver, 1, nftId);
+console.log(`NFT transfer made from minter to receiver`);
+
+const postBal = await getBal(minter);
+console.log(`Minter balance after transfer: ${postBal}.`);
+
+const postAmtNFT = await receiver.balanceOf(nftId);
+console.log(`Receiver has ${postAmtNFT} of the NFT`);
```
jeapostrophe, line 43: I think it would be better to structure this like...
- A library (i.e. a single function) that can launch an NFT
- A test case (i.e. a use of the function with the testing code... basically what you have)
- An interactive program that does it live
- A demo "real" program that launches N NFTs with different traits/etc

jeapostrophe (changes requested):

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,63 @@
+import { Reach, test } from '@reach-sh/stdlib';
+const stdlib = new Reach(process.env, { REACH_NO_WARN: 'Y' });
+
+const startingBalance = stdlib.parseCurrency(100);
+
+const [minter, receiver] = await stdlib.newTestAccounts(2, startingBalance);
+
+const gasLimit = 500000;
+if (stdlib.connector == 'ETH') {minter.setGasLimit(gasLimit)};
```
chrisnevers, line 9: This should also be applied to `CFX`. You could just do `minter.setGasLimit(gasLimit)` without the `if` condition because it is a no-op when using `ALGO`.

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,63 @@
...
+console.log(`The minter's formatted address is ${minterAddrFormat}`);
+const receiverAddrFormat = await stdlib.formatAddress(receiver);
+console.log(`The receiver's formatted address is ${receiverAddrFormat}`);
+
+const fmt = (x) => stdlib.formatCurrency(x, 4);
+const getBal = async (who) => fmt(await stdlib.balanceOf(who));
+const bal = await getBal(minter);
+console.log(`Minter starting balance: ${bal}.`);
+
+const mintNFT = async () => {
```
chrisnevers, line 27: I think `mintNFT` and `transferNFT` should be abstracted more. I'd make any value that someone would want to customize a parameter. For example, I would make the minter account a parameter of this function as well as an object of token properties. 

```javascript
const mintNFT = async (minter, opts) => { ... }
// where opts : { name, symbol, supply, au, c, ... }

await mintNFT(minter, {});
```
I think default assignments would work well here, so if the caller of `mintNFT` doesn't care about a particular field, like `reserve`, the function will use `null` by default.

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,63 @@
...
+        f: null, // freeze address
+        defaultFrozen: false, 
+        reserve: null, 
+        note: Uint8Array[1],
+    });
+    const nftId = await theNFT.id;
+    test.chk('NFT ID', nftId, theNFT.id);
+
+    await transferNFT(nftId);
+    return await nftId;
```
chrisnevers, line 42: a couple of unnecessary `await`s in this function as well as the `test.chk`

chrisnevers (changes requested):

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,70 @@
...
+    supply: 1, 
+    au: "ipfs://bafybeigdyrzt5...", //asset url
+    c: null, // clawback
+    f: null, // freeze address
+    defaultFrozen: false, 
+    reserve: null, 
+    note: Uint8Array[1],
+};
+
+const mintNFT = async (minter, name, symbol, opts = {supply: 1, au: "ipfs://bafybeigdyrzt5...", c: null, f: null, defaultFrozen: false, reserve: null, note: Uint8Array[1]}) => {
```
chrisnevers, line 40: I don't think this `supply` or `url` should be the default. The reach stdlib will set defaults for all these fields except for `note`. Given that the default `note` is empty, I would actually just remove these default parameters.

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,70 @@
...
+const getBal = async (who) => fmt(await stdlib.balanceOf(who));
+const bal = await getBal(minter);
+console.log(`Minter starting balance: ${bal}.`);
+
+const name = "JPAlgos";
+const symbol = "JPA";
+
+const opts = { 
+    supply: 1, 
+    au: "ipfs://bafybeigdyrzt5...", //asset url
```
chrisnevers, line 32: This should be `url`. `au` is algorand specific and we don't acknowledge this field when launching the token.

chrisnevers (changes requested):

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,70 @@
...
+
+const minterAddrFormat = await stdlib.formatAddress(minter);
+console.log(`The minter's formatted address is ${minterAddrFormat}`);
+const receiverAddrFormat = await stdlib.formatAddress(receiver);
+console.log(`The receiver's formatted address is ${receiverAddrFormat}`);
+
+const fmt = (x) => stdlib.formatCurrency(x, 4);
+const getBal = async (who) => fmt(await stdlib.balanceOf(who));
+const bal = await getBal(minter);
+console.log(`Minter starting balance: ${bal}.`);
```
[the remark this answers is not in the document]
chrisnevers: You'll probably need an additional param to `logBalance` like `accStr` so it'll log `Minter has ...` instead of printing the actual acc object

examples/simple-nft-minter/index.mjs
TheChronicMonster: How would I learn this information?

examples/simple-nft-minter/index.mjs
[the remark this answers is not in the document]
TheChronicMonster: exp +💯

examples/simple-nft-minter/index.mjs
TheChronicMonster: re: the library:
Could I take the `mintNFT` function and wrap it in a `launchNFT` method and add that to the Reach stdlib? 
That'd be neat!
I think the `logBalance` and `transferNFT` functions are too specific to be useful as library methods.

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,75 @@
...
+const receiverAddrFormat = await stdlib.formatAddress(receiver);
+console.log(`The receiver's formatted address is ${receiverAddrFormat}`);
+
+const fmt = (x) => stdlib.formatCurrency(x, 4);
+const getBal = async (who, tok) => fmt(await stdlib.balanceOf(who, tok));
+
+const logBalance = async (acc, tok, accStr) => {
+    const bal = await getBal(acc, tok, accStr);
+    const trueBal = await tok ? opts.supply : bal;
+    const unit = await tok ? 'of the NFT' : stdlib.connector;
```
chrisnevers, line 27: TIL of `stdlib.standardUnit`. It should be used instead of `stdlib.connector`. Right now, `standardUnit == connector` for all connectors, but that might not be the case for future connectors and standard unit is semantically correct. Also, this doesn't need to be awaited

chrisnevers (changes requested):

examples/simple-nft-minter/index.mjs
chrisnevers: That we don't support `au`? It's not listed under https://docs.reach.sh/frontend/#js_stdlib.launchToken & we only support what we say in the docs 😉

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,75 @@
...
+    accStr = acc.networkAccount.addr.toString();
```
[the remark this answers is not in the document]
TheChronicMonster: I've updated the `logBalance()` call as 
`await logBalance(minter, null, "Minter");`
but the address continues to print. 

Removing `null` from the arguments results in an `invalid BigNumber string` error.

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,75 @@
...
+console.log(`The minter's formatted address is ${minterAddrFormat}`);
+const receiverAddrFormat = await stdlib.formatAddress(receiver);
+console.log(`The receiver's formatted address is ${receiverAddrFormat}`);
+
+const fmt = (x) => stdlib.formatCurrency(x, 4);
+const getBal = async (who, tok) => fmt(await stdlib.balanceOf(who, tok));
+
+const logBalance = async (acc, tok, accStr) => {
+    const bal = await getBal(acc, tok, accStr);
+    const trueBal = await tok ? opts.supply : bal;
```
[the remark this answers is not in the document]
TheChronicMonster: formatting the non-network token was causing the amount to show as `0`. I created a ternary to check if `tok` is used.
```
const getBal = async (who, tok) => tok ? (await stdlib.balanceOf(who, tok)) : fmt(await stdlib.balanceOf(who));
```

examples/simple-nft-minter/index.mjs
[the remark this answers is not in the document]
TheChronicMonster: I was reassigning `accStr` inside the function and that's why I couldn't see the string in console logs.

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+const receiverAddrFormat = await stdlib.formatAddress(receiver);
+console.log(`The receiver's formatted address is ${receiverAddrFormat}`);
+
+const fmt = (x) => stdlib.formatCurrency(x, 4);
+const getBal = async (who, tok) => tok ? (await stdlib.balanceOf(who, tok)) : fmt(await stdlib.balanceOf(who));
+
+const logBalance = async (acc, tok, accStr = acc.getAddress()) => {
+    const bal = await getBal(acc, tok, accStr);
+    const unit = tok ? 'of the NFT' : stdlib.standardUnit;
+    console.log(`${accStr} has ${bal} ${unit}.`);
```
chrisnevers, line 26: I just made an addition to the stdlib that should be released tomorrow: `acc.getDebugLabel`.

So, when you create the accounts you can set the debug label
```JavaScript
minter.setDebugLabel('minter');
```

and this function can be cleaned up to:
```JavaScript
const logBalance = async (acc, tok) => {
    const bal = await getBal(acc, tok, accStr);
    const unit = tok ? 'of the NFT' : stdlib.standardUnit;
    console.log(`${acc.getDebugLabel()} has ${bal} ${unit}.`);
    return bal;
}
```

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+
+const transferNFT = async (minter, receiver, nftId, supply) => {
+    const preAmtNFT = await logBalance(minter, nftId, "Minter");
+
+    if (stdlib.connector == 'ALGO' && await receiver.tokenAccept(nftId)) {
+        console.log(`Receiver opted-in to NFT`);
+    };
+    if (stdlib.connector == 'ALGO' && await receiver.tokenAccepted(nftId)) {
+        console.log(`Token accepted`);
+    };
```
chrisnevers, line 61:
```suggestion
    if (stdlib.connector == 'ALGO') {
        await receiver.tokenAccept(nftId);
        console.log(`Receiver opted-in to NFT`);
    };
```
I believe if there is an issue accepting the token, an error will be thrown. So, it is redundant to later check `tokenAccepted`.

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+    await stdlib.transfer(minter, receiver, supply, nftId);
+    console.log(`${supply} ${symbol} transferred from ${minter.getAddress()} to ${receiver.getAddress()}`);
```
chrisnevers, line 63: These will be able to be `minter.getDebugLabel()` and `receiver.getDebugLabel()`

chrisnevers (commented):
Besides these comments, it looks good to me

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+    supply: 1, 
+    url: "ipfs://bafybeigdyrzt5...", //asset url
+    c: null, // clawback
+    f: null, // freeze address
+    defaultFrozen: false, 
+    reserve: null, 
+    note: Uint8Array[1],
+};
+
+const mintNFT = async (minter, name, symbol, opts = {supply, url, c, f, defaultFrozen, reserve, note}) => {
```
chrisnevers, line 49:
```suggestion
const mintNFT = async (minter, name, symbol, opts) => {
```
default params not needed

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+    const bal = await getBal(acc, tok);
+    const unit = tok ? 'of the NFT' : stdlib.standardUnit;
+    console.log(`${acc.getDebugLabel()} has ${bal} ${unit}.`);
+    return bal;
+}
+
+await logBalance(minter, null);
+
+const name = (stdlib.connector == 'ALGO') ? "JPAlgos" : "JPals";
+const symbol = (stdlib.connector == 'ALGO') ? "JPA" : "JPAL";
```
chrisnevers, line 37: I think for the purposes of this tutorial, there's no need to be so detailed regarding names. Maybe just pick a generic name and symbol?

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+const getBal = async (who, tok) => tok ? (await stdlib.balanceOf(who, tok)) : fmt(await stdlib.balanceOf(who));
+
+const logBalance = async (acc, tok) => {
+    const bal = await getBal(acc, tok);
+    const unit = tok ? 'of the NFT' : stdlib.standardUnit;
+    console.log(`${acc.getDebugLabel()} has ${bal} ${unit}.`);
+    return bal;
+}
+
+await logBalance(minter, null);
```
chrisnevers, line 34:
```suggestion
await logBalance(minter);
```

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+    await stdlib.transfer(minter, receiver, supply, nftId);
+    console.log(`${supply} ${symbol} transferred from ${minter.getDebugLabel()} to ${receiver.getDebugLabel()}`);
+    
+    const postAmtNFT = await logBalance(receiver, nftId);
+    test.chk('NFT AMT', preAmtNFT, postAmtNFT);
+}
+
+const nftId = await mintNFT(minter, name, symbol, opts);
+await transferNFT(minter, receiver, nftId, opts.supply);
+await logBalance(minter, null);
```
chrisnevers, line 72:
```suggestion
await logBalance(minter);
```

chrisnevers (commented):
LGTM after these comments

chrisnevers:
I think gen.log should be gitignored

examples/simple-nft-minter/index.mjs
TheChronicMonster: In an earlier review Jay asked for a cute NFT name. 
I wanted to show that the contract works on Algo or ETH.
Seems like most of the NFTs on Algo fit "algo" into the name, but that doesn't make sense for ETH deployments.

chrisnevers (commented):
LGTM, I think Jay needs to sign off though

examples/simple-nft-minter/index.mjs
jeapostrophe: I agree with Nevers. I think any time you write a tutorial, you want to pursue simplicity at all costs. My advice for cuteness is for the name to be "JPals" and the symbol to be "JPs" ;)

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+const symbol = (stdlib.connector == 'ALGO') ? "JPA" : "JPAL";
+
+const opts = { 
+    supply: 1, 
+    url: "ipfs://bafybeigdyrzt5...", //asset url
+    c: null, // clawback
+    f: null, // freeze address
+    defaultFrozen: false, 
+    reserve: null, 
+    note: Uint8Array[1],
```
jeapostrophe, line 46: I think this is just `undefined` and I'm surprised it is not an error

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+    console.log(`Creating the NFT`);
+    const theNFT = await stdlib.launchToken(minter, name, symbol, opts);
+    console.log(theNFT);
+    return theNFT.id;
+}
+
+const transferNFT = async (minter, receiver, nftId, supply) => {
+    const preAmtNFT = await logBalance(minter, nftId);
+
+    if (stdlib.connector == 'ALGO') {
```
jeapostrophe, line 59: I think you should always do it and note that it does nothing on ETH

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+
+const minterAddrFormat = await stdlib.formatAddress(minter);
+console.log(`The minter's formatted address is ${minterAddrFormat}`);
+const receiverAddrFormat = await stdlib.formatAddress(receiver);
+console.log(`The receiver's formatted address is ${receiverAddrFormat}`);
+
+const fmt = (x) => stdlib.formatCurrency(x, 4);
+const getBal = async (who, tok) => tok ? (await stdlib.balanceOf(who, tok)) : fmt(await stdlib.balanceOf(who));
+
+const logBalance = async (acc, tok) => {
```
jeapostrophe, line 27: In these two functions, you have an "acceptable" path on `tok` being `undefined`. I think it is kinder to use `(who, tok = false)`, to better communicate to the reader that often you expect it to not be provided

examples/simple-nft-minter/index.mjs
jeapostrophe, line 46: I think that you should use the ARC spec and have the note actually contain what it is supposed to

examples/simple-nft-minter/index.mjs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+await logBalance(minter);
+
+const name = (stdlib.connector == 'ALGO') ? "JPAlgos" : "JPals";
+const symbol = (stdlib.connector == 'ALGO') ? "JPA" : "JPAL";
+
+const opts = { 
+    supply: 1, 
+    url: "ipfs://bafybeigdyrzt5...", //asset url
```
jeapostrophe, line 41: I think it is more compelling if you demonstrate actually creating it too. The same information that you're going to put in the Note field, you want to upload to IPFS.

This page has a little tutorial:

https://js.ipfs.tech/

under "Adding data to IPFS"

If you are nervous about including it in the code always or having our CI do it, etc, then at least make a little program on the side (or a way to run this program) and comment on it in your guide

jeapostrophe (changes requested):

PR #1427, opened 2022-09-16
CORE-1846 Normalize mnemonic and secret key formats

erer1243:


docs/src/frontend/index.md
```diff
@@ -497,11 +498,12 @@ The `acc` object `{!rsh} await`s the user to input the secret and then sets the
 ---
 @{ref("js", "newAccountFromMnemonic")}
 ```js
-stdlib.newAccountFromMnemonic(phrase: string) => Promise<acc>
+stdlib.newAccountFromMnemonic(phrase: string) => Promise<Account>
 ```
 
 Returns a Promise for a Reach account abstraction for an account on the consensus network specified by the given mnemonic phrase.
-The details of the mnemonic phrase encoding are specified uniquely to the consensus network.
+The mnemonic phrase must be a string of whitespace-separated words from the BIP-39 English wordlist.
```
jeapostrophe, line 505: It is best to have a link to this. @TheChronicMonster

jeapostrophe (approved):

docs/src/frontend/index.md
erer1243: We could link to the wordlist but I don't think it would be helpful to lay users.

PR #1428, opened 2022-09-16
CORE-2275 Reduce try/catch surface in getView1

erer1243:
A couple weeks ago we fixed a bug that was hard to find because it was caught in this try block and ignored. Though it has been fixed, I believe it is smarter to have less code surface that blindly ignores errors. So this strips the getView1 try to a the critical read action only, and shortens the rest a little.

### DIFF
--- a/js/stdlib/ts/ALGO.ts
+++ b/js/stdlib/ts/ALGO.ts
@@ -2187,20 +2187,21 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account> 
           debug('getView1', v, k, args);
           const { decode } = vim;
           const ch = await getC();
-          const step = await getCurrentStep_(ch);
-          const vi = bigNumberToNumber(step);
-          const vTys = vs[vi];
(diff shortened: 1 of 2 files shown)


js/stdlib/ts/ALGO.ts
```diff
@@ -2191,18 +2191,15 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account>
             const step = await getCurrentStep_(ch);
             const vi = bigNumberToNumber(step);
             const vtys = vs[vi];
-            if ( ! vtys ) { throw Error(`no views for state ${step}`); }
-            const [ _, vvs ] = await getState_(getC, _ => vtys);
+            if (!vtys) { throw Error(`no views for state ${step}`); }
+            const vvs = (await getState_(getC, _ => vtys))[1];
```
jeapostrophe, line 2195: I really prefer destructuring rather than `[1]`

jeapostrophe (approved):

PR #1432, opened 2022-09-19
Remote algo accounts

willghatch:
Add accounts field to remote.ALGO.


jeapostrophe (approved):

jeapostrophe (approved):
You should use the gen-test-output script

willghatch:
> You should use the gen-test-output script

Aha, thanks, I will in the future.  I should have looked for how it was officially done long ago.

PR #1433, opened 2022-09-20
Account app opt in

willghatch:


docs/src/frontend/index.md
```diff
@@ -696,6 +696,26 @@ acc.tokensAccepted() => Promise<Array<Token>>
 Returns a Promise for an array of tokens that are accepted by `acc`.
 On networks which do not keep track of this information (e.g. Ethereum), this returns an empty array.
 
+---
+@{ref("js", "appOptIn")}@{ref("js", "stdlib.accountAppOptIn")}
+```js
+acc.appOptIn(contract) => Promise<void>
+stdlib.accountAppOptIn(account, contract) => Promise<void>
```
jeapostrophe, line 703: This one shouldn't be added because it is a write function

docs/src/frontend/index.md
```diff
@@ -696,6 +696,26 @@ acc.tokensAccepted() => Promise<Array<Token>>
 Returns a Promise for an array of tokens that are accepted by `acc`.
 On networks which do not keep track of this information (e.g. Ethereum), this returns an empty array.
 
+---
+@{ref("js", "appOptIn")}@{ref("js", "stdlib.accountAppOptIn")}
+```js
+acc.appOptIn(contract) => Promise<void>
```
jeapostrophe, line 702: I believe the type should be Contract and explicitly say it is the "info". I partially think it should be part of the ctc object actually

docs/src/frontend/index.md
```diff
@@ -696,6 +696,26 @@ acc.tokensAccepted() => Promise<Array<Token>>
...
+stdlib.accountAppOptIn(account, contract) => Promise<void>
+```
+
+This does nothing on some networks.
+On others, it opts into the given `contract`.
+
+---
+@{ref("js", "appOptedIn")}@{ref("js", "stdlib.accountAppOptedIn")}
+```js
+acc.appOptedIn(contract) => Promise<boolean>
```
jeapostrophe, line 712: Clarify type as Contract / info

js/stdlib/ts/ALGO.ts
```diff
@@ -2423,6 +2421,31 @@ const balanceOf = async (acc: Account | Address, token?: Token): Promise<BigNumb
...
 };
 
+const doAccountAppOptedIn = async (nacc: NetworkAccount, ctcId: BigNumber): Promise<boolean> => {
+  const ls = await getLocalState_(nacc.addr, ctcId);
+  return ls !== undefined;
+}
+const accountAppOptedIn = async (acc: Account, ctc: Contract): Promise<boolean> => {
+  return await doAccountAppOptedIn(acc.networkAccount, await ctc.getInfo());
+}
+const doAccountAppOptIn = async (nacc: NetworkAccount, ctcId: BigNumber): Promise<void> => {
```
jeapostrophe, line 2431: This function will not work for non Reach code and it is dangerous because without contract verification, the user has no idea how much it will cost

js/stdlib/ts/shared_impl.ts
```diff
@@ -533,6 +533,8 @@ export type IAccount<NetworkAccount, Backend, Contract, ContractInfo, Token> = {
   stdlib: Object,
   getAddress: () => string,
   setDebugLabel: (lab: string) => IAccount<NetworkAccount, Backend, Contract, ContractInfo, Token>,
+  appOptedIn: (ctc: Contract) => Promise<boolean>,
+  appOptIn: (ctc: Contract) => Promise<void>,
```
jeapostrophe, line 537: These types are wrong. It should be ContractInfo I think

jeapostrophe (changes requested):

willghatch:
I don't think that failure is my fault.

docs/src/frontend/index.md
willghatch: So we only want to add `accountAppOptedIn` and not `accountAppOptIn`?

docs/src/frontend/index.md
willghatch: I'm not sure I understand what you mean.  Do you mean you want the docs to say `appOptIn(contract: Contract)`?  Or you mean you would prefer this as `ctc.appOptIn(account)`?

docs/src/frontend/index.md
jeapostrophe: I am saying that `stdlib.accountAppOptIn` should not exist because we generally don't have write functions in the `stdlib` namespace. Furthermore, I think that it shouldn't be in `acc` either. It should be in `ctc`, so... `acc.contract(info).optIn()`, because we need to do contract verification

docs/src/frontend/index.md
jeapostrophe: I had two ideas. The second sentence of mine and the third sentence of yours line up, but there's no need for the `account` argument, because a `ctc` knows its account. If you do that, then there will be no `contract` argument, so the first sentence isn't relevant here

docs/src/frontend/index.md
jeapostrophe: This one would need some clarification for what type `contract` is. We need to ensure that the user understands it is the "info" for a contract (like the Reach type `Contract`) and not a contract handle in JS

docs/src/frontend/index.md
```diff
@@ -696,6 +696,25 @@ acc.tokensAccepted() => Promise<Array<Token>>
...
+ctc.appOptIn() => Promise<void>
+```
+
+This does nothing on some networks.
+On others, it opts into the contract `ctc`.
+
+---
+@{ref("js", "appOptedIn")}@{ref("js", "stdlib.accountAppOptedIn")}
+```js
+acc.appOptedIn(contract) => Promise<boolean>
```
jeapostrophe, line 711:
```suggestion
acc.appOptedIn(info:Contract) => Promise<boolean>
```

docs/src/frontend/index.md
```diff
@@ -696,6 +696,25 @@ acc.tokensAccepted() => Promise<Array<Token>>
...
+stdlib.accountAppOptedIn(account, contract) => Promise<boolean>
```
jeapostrophe, line 712:
```suggestion
stdlib.accountAppOptedIn(acc: Account | Address, info:Contract) => Promise<boolean>
```

docs/src/frontend/index.md
```diff
@@ -696,6 +696,25 @@ acc.tokensAccepted() => Promise<Array<Token>>
...
+```
+
+On some networks this always returns `true`.
+On others, it checks whether the account has opted into `contract`.
```
jeapostrophe, line 716:
```suggestion
On others, it checks whether the account has opted into the contract specified by `info`.
```

docs/src/frontend/index.md
jeapostrophe, line 712: The naming convention for things like this is to use the same name with no prefix, so `stdlib.appOptedIn` would be the name

examples/account-appOptIn/index.mjs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const ctcB = accB.contract(backend, await ctc.getInfo());
+await ctcB.appOptIn();
+// A second opt-in should be a no-op, not a crash.
+await ctcB.appOptIn();
+
+assert(await stdlib.accountAppOptedIn(accB, await ctc.getInfo()));
+assert(await accB.appOptedIn(await ctc.getInfo()));
+
```
jeapostrophe, line 34: Please include an example of `stdlib.appOptedIn` with an address rather than an account handle

js/stdlib/ts/ALGO.ts
```diff
@@ -2426,6 +2426,30 @@ const balanceOf = async (acc: Account | Address, token?: Token): Promise<BigNumb
   return (await balanceOfM(acc, token || null)) || bigNumberify(0);
 };
 
+const doAccountAppOptedIn = async (nacc: NetworkAccount, ctcId: ContractInfo): Promise<boolean> => {
+  const ls = await getLocalState_(nacc.addr, ctcId);
+  return ls !== undefined;
+}
+const accountAppOptedIn = async (acc: Account, ctc: ContractInfo): Promise<boolean> => {
```
jeapostrophe, line 2433: This should take `Account | Address`, because `Account` has a signer and the whole point of this is to be able to inspect other accounts that you don't control, not just to have a different calling pattern

js/stdlib/ts/ETH_like.ts
```diff
@@ -1176,6 +1182,13 @@ function setCustomHttpEventHandler() {
   console.warn(`setCustomHttpEventHandler is not supported on this connector`);
 }
 
+const doAppOptIn = async (_ctc: ContractInfo): Promise<void> => {
+  return;
+};
+const accountAppOptedIn = async (_acc: Account, _ctc: ContractInfo): Promise<boolean> => {
```
jeapostrophe, line 1188: This function should be included in `interfaces.ts` as part of the stdlib (and it needs to take `Account | Address`

jeapostrophe (changes requested):

examples/account-appOptIn/index.mjs
jeapostrophe: .

js/stdlib/ts/ALGO.ts
```diff
@@ -2426,15 +2426,16 @@ const balanceOf = async (acc: Account | Address, token?: Token): Promise<BigNumb
   return (await balanceOfM(acc, token || null)) || bigNumberify(0);
 };
 
-const doAccountAppOptedIn = async (nacc: NetworkAccount, ctcId: ContractInfo): Promise<boolean> => {
-  const ls = await getLocalState_(nacc.addr, ctcId);
+const doAccountAppOptedIn = async (nacc: string, ctcId: ContractInfo): Promise<boolean> => {
```
jeapostrophe, line 2429: This should be `Address` or something like that

jeapostrophe (changes requested):

examples/account-appOptIn/index.mjs
willghatch: I did.  Do you want me to do it both ways every time I call it?

jeapostrophe: I'm sorry, I didn't notice it, maybe it was because of rebasing.

jeapostrophe (approved):

PR #1434, opened 2022-09-21
add script to measure ETH bytecode size

willghatch:


willghatch:
Oh, now that I look at the Jira ticket, this doesn't quite match the description, but it gives the data we wanted.

jeapostrophe (approved):

PR #1436, opened 2022-09-23
improve size measurement script

willghatch:
Here is an improved version of the measurement script.  It reports ETH as well as ALGO and the ratio of ETH to ALGO, it can be run on a single directory or on the full examples directory, and it can optionally compile before measuring.


scripts/measure-size.mjs
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env node
+
+import * as fs from 'fs/promises';
+import * as process from 'process';
+import * as cp from "child_process";
+
+// Probably I should use an arg parsing package, but I don't want to bother dealing with dependencies right now...
```
jeapostrophe, line 7: Check out zx for scripts like this

jeapostrophe (approved):

scripts/measure-size.mjs
willghatch: Where is the configuration of what is in the circle environment?  I didn't want to deal with any dependencies because I don't know what is in the environment or how to add things.

jeapostrophe: I assumed we weren't going to run this in Circle... it is just for us?

jeapostrophe: But, to answer your question, you can look in `/.circleci/config.pre.yml` and there will be an image that is our base

willghatch: Oh, the ticket about measuring things says to add it to the CI so we are measuring every build.  Anyway, I saw a reference to an image, but I didn't see where it was defined.  But I can poke around later.

jeapostrophe: Oh, I didn't recall the measuring every build. I'm not sure how we're going to monitor that in practice. FWIW, I would love to always have zx available. It's great.

These are the images: https://github.com/reach-sh/reach-lang/blob/master/.circleci/config.pre.yml#L16-L25

PR #1437, opened 2022-09-23
CORE-2211: Upgrade CFX SDK

chrisnevers:


jeapostrophe (approved):

jeapostrophe (changes requested):
Actually, I think you need to change `js/js-deps/package.json` and then run `make update-lock` in `js/js-deps` to really see a difference. You also have to update `js/react-runner`

PR #1438, opened 2022-09-23
WIP - add untested edits to add size measurement to circleci

willghatch:
I haven't ever touched the CI scripts, so I have no idea initially if this will work, but I'm making a draft PR so I can see what the CI does with what I've written.


jeapostrophe (changes requested):
As I mentioned in the other one, I don't think this is actually needed, because we don't know how we're going to be monitoring it. If we want to start making some sort of "stats dashboard" maybe this can go on it. I'd abandon this and declare victory on the main card.

willghatch:
Ok, sounds good.

PR #1447, opened 2022-09-29
APIs that exist twice don't need multiple fields in ApiRng.

willghatch:
Frankly I have no idea how this was working before, because the generated code would set one field of the ApiRng struct but read from another, and somehow get the value set in the other field!  Maybe this is a solidity bug.

### DIFF
--- a/hs/src/Reach/Connector/ETH_Solidity.hs
+++ b/hs/src/Reach/Connector/ETH_Solidity.hs
@@ -1568,7 +1568,8 @@ apiArgs tyMsg (ApiInfo {..}) = do
 
 apiDef :: SLPart -> Bool -> ApiInfo -> App Doc
 apiDef who qualify ApiInfo {..} = do
-  let who_s = adjustApiName (bunpack who) ai_which qualify
+  let who_orig = (bunpack who)
+  let who_s = adjustApiName who_orig ai_which qualify
   let mf = solMsg_fun ai_which
   (argDefns, tyLifts, args, m_arg_ty) <- apiArgs "_t.msg" $ ApiInfo {..}
   when (length args > apiMaxArgs) $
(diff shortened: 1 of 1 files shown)


jeapostrophe (commented):
I want to push a little bit on this. If this is really a problem, I feel like there has to be an example that you could cook up that shows an error.

On Slack, you asked if aliases could have different types and the answer is definitely no in one way and yes in the other. If we have
```
f :: X -> Y
g :: A - >B
```
Then `f` and `g` are unique functions with unique ranges.
We can then alias both of them to `h`:
```
alias f h
alias g h
```
Now it is as if `h` has two types:
```
h :: X -> Y
h :: A -> B
```
So in this case "Do aliases have different types?"
1. Yes, `h` has two different types because of aliases.
2. No, `f` and its `h` alias have the same type; as do `g` and its `h` alias.

When you asked before, I thought you were talking about 1, but maybe you were talking about 2?

willghatch:
My question was silly.  The example was not about aliases, it was about the same non-aliased API being used in two places.  I was confounding the two when I asked the question.  The API has the same return type everywhere (and the same domain everywhere), and the `ApiRng` struct just had extra fields added for seemingly no reason.  These extra fields were accessed only for reading, while another field was also generated for the API as a whole which was used for all writes.  Again I don't know how it ever worked, and I think it's a solidity bug.  But this patch removes the extra fields and moves all reads to the field that is written, which is unique to the API, but not unique to the different calls to the API in the Reach program.

willghatch:
To maybe clarify a bit, the `ApiRng` struct was generated with a field for the API, and also a field for the API for each message/state it was used in.  It was writing to the field for the API overall (IE `Writer_f`), but depending on the `current_step`, was returning either `_Writer_f3` or `_Writer_f4` (for step 3 or 4, respectively).  But the return type of the API is the same whether it is step 3 or 4, it was already writing to the same field for both, and it should have been reading from the same field for both.

jeapostrophe (approved):

jeapostrophe:
Ship it!

PR #1448, opened 2022-09-30
js: getEventSigs

DanBurton:
```
$ REACH_CONNECTOR_MODE=ALGO reach run
static assertions passed
Warning: your program uses stdlib.fundFromFaucet. That means it only works on Reach devnets!
logs [
  <Buffer 82 c4 61 fe 00 00 00 00 00 00 00 07>,
  <Buffer 5c 4c 54 5c 00 00 00 00 00 00 00 07>,
  <Buffer d2 a5 9f 6c 01 00 68 69 68 6f 23 41 9a 66 49 b0 3f e7 06 dd 77 fe 1d 9e 39 bc 02 ea 1a b3 c4 ef a9 42 87 88 67 58 9d a1 7e ad 00 00 00 00 00 00 00 30 ... 64 more bytes>
]
5c4c545c size 12 from pub(uint64)
d2a59f6c size 114 from big(((byte,byte[1]),byte[2][2]),(address,uint64,uint256,digest))
log observation assertions passed
```

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -903,6 +903,22 @@ When the `{!js} showFull` argument is not `{!js} true`, internal implementation
 details are omitted.
 
 ---
+@{ref("js", "ctc.getEventTys")}
+```js
+ctc.getEventTys() => Record<string, ReachType[]>
+```
+
+Returns a mapping from event name to the reach types of the data emitted for that event.
+This can be used to determine the connector-specific event signature.
+For example:
+
+```js
+const tys = ctc.getEventTys();
+const e = 'Swap';
+const sig = `${e}${tys[e].toString()}`;
+```
+
+---
 @{ref("js", "getInternalState")}
 ```js
 ctc.getInternalState() => Promise<{ [key: string], ReachType }>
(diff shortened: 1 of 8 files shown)


jeapostrophe (changes requested):
This needs documentation and changelog entry.

With the addition of https://docs.reach.sh/frontend/#js_toString, I think it is acceptable to say that this function actually just is `cosnt getEventSigs = () => bin._getEvents({ reachStdlib: stdlib })`, because we can then tell people, "Yea, it is a list of all the events and their types" and people can call it themselves.

docs/src/frontend/index.md
```diff
@@ -902,6 +902,22 @@ Returns the ABI to the contract in a connector-specific format.
 When the `{!js} showFull` argument is not `{!js} true`, internal implementation
 details are omitted.
 
+---
+@{ref("js", "ctc.getEventTys")}
+```js
+ctc.getEventTys() => Record<string, ReachType[]>
+```
+
+Returns a mapping from event name to the reach types of the data emitted for that event.
```
jeapostrophe, line 911:
```suggestion
Returns a mapping from event name to the Reach types of the data emitted for that event.
```

DanBurton:
* changed the impl to just boil down to _getEvents
* added docs & changelog
* expanded the example to demo how to get and use the ARC-28 sig from this info

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #1450, opened 2022-10-03
fix measure-size script compile option

willghatch:
At some point I broke the `--compile` option when I was rearranging code, and this fixes it...


willghatch:
And then I'll just force push another one line improvement into the same PR since I just barely posted it...

jeapostrophe (approved):

PR #1453, opened 2022-10-06
CORE-1009: Float If Optimization

chrisnevers:
I added a `Maybe DLVar` annotation to `if`/`do` `DLStmt`s; it represents the variable that is being set within the block. It is used for efficiently determining what `if`s to float. The opt is essentially:

- Process `DL_Var dv`
  - Process Tail
  - If `{ dv : 0 }` is in the env
    - Drop `DL_Var` because we generated a `const` for it
  - Otherwise 
    - we couldn't float the definition so keep `DL_Var dv`
- Sees `DL_LocalIf (Just answer)`
  - Increment the value associated with `answer` in the env.
  - Process DLStmt and Tail
  - If both branches of the `if` are pure
    - Lift the branch stmts. 
    - Decrement value associated with `answer` in env, signaling we inlined def. 
    - Generate a `const` or `set` for `answer` depending on whether we're dealing with nested `if`s that assign to the same var.
  - Otherwise 
    - do nothing (this will leave `{ dv: n } where n > 0` in the env, signaling that we failed to inline the def

### DIFF
--- a/hs/src/Reach/Linearize.hs
+++ b/hs/src/Reach/Linearize.hs
@@ -562,8 +562,8 @@ df_t = \case
 
 df_con :: DKTail -> DFApp LLConsensus
 df_con = \case
-  DK_If a mans c t f ->
-    LLC_If a mans c <$> df_con t <*> df_con f
+  DK_If a _ c t f ->
+    LLC_If a c <$> df_con t <*> df_con f
   DK_Switch a v csm ->
     LLC_Switch a v <$> mapM cm1 csm
     where
--- a/hs/src/Reach/Linearize.hs
+++ b/hs/src/Reach/Linearize.hs
@@ -88,6 +88,13 @@ d_inv :: (Applicative m) => (a -> m b) -> DLInvariant a -> m (DLInvariant b)
 d_inv block (DLInvariant inv_b minv_lab) =
   DLInvariant <$> block inv_b <*> pure minv_lab
 
+getRet :: Maybe DLVar -> DKApp (Maybe DLVar)
+getRet = \case
+  Just a  -> return $ Just a
+  Nothing -> do
+    ret <- asks eRet
+    return $ fst3 . snd <$> ret
+
 dk1 :: DKTail -> DLSStmt -> DKApp DKTail
 dk1 k s =
   case s of
@@ -96,9 +103,10 @@ dk1 k s =
       com' $ DKC_ArrayMap at ans xs as i <$> dk_block at f
     DLS_ArrayReduce at ans xs z b as i f ->
       com' $ DKC_ArrayReduce at ans xs z b as i <$> dk_block at f
-    DLS_If at c _ t f -> do
-      let con = DK_If at c
-      let loc t' f' = DK_Com (DKC_LocalIf at c t' f') k
+    DLS_If at mans c _ t f -> do
+      ans <- getRet mans
+      let con = DK_If at ans c
+      let loc t' f' = DK_Com (DKC_LocalIf at ans c t' f') k
       let mt = DK_Stop at
       (mk, k') <-
         getDKBM s >>= \case
@@ -135,7 +143,7 @@ dk1 k s =
       case isLocal s of
         True -> do
           ss' <- withReturn ret (dv, False, DK_Stop at) $ dk_top at ss
-          return $ DK_Com (DKC_Var at dv) $ DK_Com (DKC_LocalDo at ss') k
+          return $ DK_Com (DKC_Var at dv) $ DK_Com (DKC_LocalDo at (Just dv) ss') k
         False ->
           withReturn ret (dv, True, k) $
             case turnVarIntoLet of
@@ -255,8 +263,8 @@ instance CanLift DKCommon where
     DKC_ArrayReduce _ _ _ _ _ _ _ f -> canLift f
     DKC_Var {} -> True
     DKC_Set {} -> True
-    DKC_LocalDo _ t -> canLift t
-    DKC_LocalIf _ _ t f -> canLift t && canLift f
+    DKC_LocalDo _ _ t -> canLift t
+    DKC_LocalIf _ _ _ t f -> canLift t && canLift f
     DKC_LocalSwitch _ _ csm -> canLift csm
     DKC_MapReduce _ _ _ _ _ _ _ f -> canLift f
     DKC_FluidSet {} -> True
@@ -319,7 +327,7 @@ instance LiftCon DKTail where
     DK_Stop at -> return $ DK_Stop at
     DK_ToConsensus at send recv mtime -> do
       DK_ToConsensus at send <$> (noLifts $ lc recv) <*> lc mtime
-    DK_If at c t f -> DK_If at c <$> lc t <*> lc f
+    DK_If at mans c t f -> DK_If at mans c <$> lc t <*> lc f
     DK_Switch at v csm -> DK_Switch at v <$> lc csm
     DK_FromConsensus at1 at2 fs k ->
       captureLifts at1 $ DK_FromConsensus at1 at2 fs <$> lc k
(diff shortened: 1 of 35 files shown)


hs/src/Reach/Optimize.hs
```diff
@@ -246,6 +248,99 @@ opt_v2v v = fst <$> opt_v2p v
...
+
+instance IsCom LLConsensus where
+  isCom = \case
+    LLC_Com m k -> Just (m, k)
+    _ -> Nothing
+
+class LiftPure a where
+  -- Returns what `DLVar` is `DL_Set` to and the pure `DLStmt`s along the way
+  -- Returns `Nothing` if there are impure stmts in `a` or there is no `DL_Set`
+  liftPure :: DLVar -> a -> (Maybe DLArg, [DLStmt])
```
jeapostrophe, line 282: This is a bit of bike-shedding and a demand for purity... I think `DLVar -> a -> Maybe (DLArg, [DLStmt])` is a more accurate type.

hs/src/Reach/Optimize.hs
```diff
@@ -246,6 +248,99 @@ opt_v2v v = fst <$> opt_v2p v
...
+    -- There were zero branches that could not be lifted, i.e.
+    -- we generated a `const` assign, now drop this `var`.
+    Just 0 -> return $ k
+    _ -> return $ mkk (DL_Var at dv) k
+
+mkVar :: SrcLoc -> DLType -> App DLVar
+mkVar at t = do
+  ctr <- asks eCounter
+  idx <- liftIO $ incCounter ctr
+  return $ DLVar at Nothing t idx
```
jeapostrophe, line 309: Just for cleanliness, you should move this up near the `Env` definition of utilities for this file

hs/src/Reach/Optimize.hs
```diff
@@ -246,6 +248,99 @@ opt_v2v v = fst <$> opt_v2p v
...
+floatIf mkk at ansDv c tt ft k = do
+  case isPure tt && isPure ft of
+    True ->
+      case (liftPure ansDv tt, liftPure ansDv ft) of
+        ((Just ta, t_lifts), (Just fa, f_lifts)) -> do
+          eVarsSetR <- asks eVarsSet
+          eVarsSet <- (liftIO . readIORef) eVarsSetR
+          let mVal = M.lookup ansDv eVarsSet
+          -- Track that we successfully lifted this assignment
+          liftIO $ modifyIORef' eVarsSetR $ M.insert ansDv $ mSub1 mVal
```
jeapostrophe, line 321: There is the `M.alter` function

hs/src/Reach/Optimize.hs
```diff
@@ -246,6 +248,99 @@ opt_v2v v = fst <$> opt_v2p v
...
+          let asn v = DL_Let at (DLV_Let DVC_Many v) $ DLE_PrimOp at IF_THEN_ELSE [c, ta, fa]
+          t <- case mVal of
+                Just 1 -> return $ mkk (asn ansDv) k
```
jeapostrophe, line 324: This is matching on 1 because it is GOING to be 0 after this? I feel like that's confusing and I want you to match on `mSub1 mVal`

hs/src/Reach/Optimize.hs
```diff
@@ -246,6 +248,99 @@ opt_v2v v = fst <$> opt_v2p v
...
+                Just _ -> do
+                  tmp <- mkVar at (varType ansDv)
+                  let s2 = DL_Set at ansDv (DLA_Var tmp)
+                  return $ mkk (asn tmp) (mkk s2 k)
```
jeapostrophe, line 328: This is the situation where there are two ifs that defining the same variable and we've made one of them pure but the other (yet)?

hs/src/Reach/Optimize.hs
```diff
@@ -246,6 +248,99 @@ opt_v2v v = fst <$> opt_v2p v
...
+          let (ma, dt') = liftPure dv dt
+          (ma, ds : dt')
+        False -> (Nothing, [])
+
+floatVar :: (DLStmt -> t -> t) -> SrcLoc -> DLVar -> t -> App t
+floatVar mkk at dv k = do
+  eVarsSet <- (liftIO . readIORef) =<< asks eVarsSet
+  case M.lookup dv eVarsSet of
+    -- There were zero branches that could not be lifted, i.e.
+    -- we generated a `const` assign, now drop this `var`.
```
jeapostrophe, line 301: It is possible for this to happen because even `Optimize` is generally "top down", `float` is "locally bottom up" and we float after optimizing the tail

hs/src/Reach/Optimize.hs
```diff
@@ -933,10 +1044,12 @@ instance {-# OVERLAPPING #-} (Optimize a, Optimize b, Optimize c, Optimize d, Op
 
 instance Optimize ETail where
   opt = \case
-    ET_Com m k -> mkCom ET_Com <$> opt m <*> opt k
+    ET_Com m k -> do
+      preFloat m
+      float ET_Com =<< mkCom ET_Com <$> opt m <*> opt k
```
jeapostrophe, line 1049: You duplicate this pattern above on line 990 and below on line 1074. You should make a function and a comment that this is "locally bottom-up"

jeapostrophe (changes requested):
I took a long to review this because I needed to understand it well. It makes sense to me, but I want more evidence that it applies. In the comments, I request seeing for this specific program you made, but I would also like to know how many other program it applies to.

My suggestion for how to figure that out is to add a simple print out when it applies, run hs-test-accept and see how many files get updated and try to get some sense of their size change. I am tempted to suggest that we always run the tests with REACH_DEBUG=Y so we get the cost output on Algorand always and can use that as a way to observe when compiler changes drastically change cost and size. [Maybe we can make it so the ETH connector also gives size information.]

hs/src/Reach/Linearize.hs
```diff
@@ -96,9 +103,10 @@ dk1 k s =
       com' $ DKC_ArrayMap at ans xs as i <$> dk_block at f
     DLS_ArrayReduce at ans xs z b as i f ->
       com' $ DKC_ArrayReduce at ans xs z b as i <$> dk_block at f
-    DLS_If at c _ t f -> do
-      let con = DK_If at c
-      let loc t' f' = DK_Com (DKC_LocalIf at c t' f') k
+    DLS_If at mans c _ t f -> do
+      ans <- getRet mans
```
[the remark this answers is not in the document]
chrisnevers: If `mans = Just dv`, we already know that this `if` will set `dv`. If we do not have that metadata, try to lookup in the env what DLVar the return value will be assigned to. 

This is necessary for the `dk1 DLS_Prompt` case:

```Haskell
withReturn ret (dv, True, k) $
    case turnVarIntoLet of
      True -> dk_ k ss
      False -> DK_Com (DKC_Var at dv) <$> dk_ k ss
```
so we don't forgot that `dv` is being set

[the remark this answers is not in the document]
jeapostrophe: Cool. Can you add a comment?

hs/src/Reach/Optimize.hs
```diff
@@ -786,18 +881,30 @@ instance Optimize DLStmt where
...
+    DL_LocalDo _ _ t -> gcs t
+
+-- Track that we are attempting to lift this assignment
+preFloat :: DLStmt -> App ()
+preFloat = \case
+  DL_LocalIf _ (Just dv) _ _ _ -> do
+    eVarsSetR <- asks eVarsSet
+    eVarsSet <- (liftIO . readIORef) eVarsSetR
+    liftIO $ modifyIORef' eVarsSetR $
+      M.insert dv $ maybe (1) succ $ M.lookup dv eVarsSet
```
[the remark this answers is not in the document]
chrisnevers: It should always increment even if it is already set because we're tracking how many `if`s we ought to float in order to drop `DL_Var`. If there's any impure `if`s, `dv` will be associated with a non-zero number and we'll know to keep `DL_Var`

hs/src/Reach/Optimize.hs
chrisnevers: Yes, there's multiple `if`s defining the same variable and we are in the process of floating an inner `if`; we've yet to know if the outer layers are pure

chrisnevers:
The float opt applies to the following programs:

| Name | Step Budgets (Before) | Step Budgets (After) | Program Bytes (Before)| Program Bytes (After) |
|-|-|-|-|-|
|no-tail-if.rsh|151<br/>184|151<br/>185|485|471|
|ttt/index.rsh|152<br/>143<br/>469<br/>443<br/>445|152<br/>143<br/>440<br/>413<br/>415|1428|1288|

As well as `localifret.rsh`, `lazy-ifs.rsh`, `Fix_Return_MustBeTail.rsh`. However, these 3 programs did not have any size changes.

jeapostrophe (changes requested):
Looks good, just that one comment left

reach-everest
```diff
@@ -0,0 +1 @@
+Subproject commit de8f349881cd51330e1e6cbb0001bbc3073cfedd
```
jeapostrophe, line 1: Oops

jeapostrophe (approved):

jeapostrophe (approved):

PR #1456, opened 2022-10-07
CORE-2528 Add simulationValues

erer1243:
Since I said I would make a pr tonight, I wanted to push what I had. Mostly™ done, not 100%, but feedback welcome. todo: testing & docs

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -625,7 +625,7 @@ In addition, a remote function may be augmented with one of the following operat
     The field represents how many of each non-network token Reach will assume your contract received from the remote call, for the purposes of transaction simulation.
     Amounts must be specified in the same order as in the call to `{!rsh} bill` or `{!rsh} withBill`.
     This is useful when using `{!rsh} withBill` and `{!rsh} enforce` to enforce that your contract received a certain number non-network tokens from the remote call.
-  + `{!rsh} opts.simReturnVal` is a field whose type must match the returned type of the remote function (default unspecified).
+  + `{!rsh} opts.simReturnVal` is a field whose type must match the returned type of the remote function (default depends on return type).
     The field represents the return value of the remote function, for the purposes of transaction simulation.
     This is useful when using `{!rsh} enforce` on the return value of the remote function.
 
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -618,6 +618,16 @@ In addition, a remote function may be augmented with one of the following operat
   + `{!rsh} opts.strictPay` does not optimize away `pay` or `axfer` transactions that are statically zero.
     If this is needed, and not included, then `{!rsh} remote` calls that require a payment transaction to always be present, even if the amount is zero, could fail.
   + `{!rsh} opts.rawCall` is a boolean (default `{!rsh} false`) that when `{!rsh} true` omits the ABI method selector from the call.
+  + `{!rsh} opts.simNetRecv` is an integer field (default `{!rsh} 0`).
+    The field represents how many network tokens Reach will assume your contract received from the remote call, for the purposes of transaction simulation.
+    This is useful when using `{!rsh} withBill` and `{!rsh} enforce` to enforce that your contract received a certain number network tokens from the remote call.
+  + `{!rsh} opts.simTokensRecv` is a tuple-of-integers field (default all zeros).
+    The field represents how many of each non-network token Reach will assume your contract received from the remote call, for the purposes of transaction simulation.
+    Amounts must be specified in the same order as in the call to `{!rsh} bill` or `{!rsh} withBill`.
+    This is useful when using `{!rsh} withBill` and `{!rsh} enforce` to enforce that your contract received a certain number non-network tokens from the remote call.
+  + `{!rsh} opts.simReturnVal` is a field whose type must match the returned type of the remote function (default unspecified).
+    The field represents the return value of the remote function, for the purposes of transaction simulation.
+    This is useful when using `{!rsh} enforce` on the return value of the remote function.
 
 If the remote contract is not expected to return non-network tokens then a pair is returned, where the amount of network tokens received is the first element, and the original result is the second element.
 
--- a/hs/src/Reach/AST/DLBase.hs
+++ b/hs/src/Reach/AST/DLBase.hs
@@ -359,6 +359,9 @@ instance PrettySubst DLArg where
 argLitZero :: DLArg
 argLitZero = DLA_Literal $ DLL_Int sb UI_Word 0
 
+argLitNull :: DLArg
+argLitNull = DLA_Literal DLL_Null
+
 staticZero :: DLArg -> Bool
 staticZero = \case
   DLA_Literal (DLL_Int _ _ 0) -> True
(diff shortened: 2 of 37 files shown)


hs/src/Reach/AST/DLBase.hs
```diff
@@ -755,11 +758,14 @@ data DLRemoteALGO = DLRemoteALGO
   , ralgo_onCompletion :: DLRemoteALGOOC
   , ralgo_strictPay :: Bool
   , ralgo_rawCall :: Bool
+  , ralgo_simValues :: (DLArg, -- netRecv
+                        Either (Maybe [DLArg]) DLArg, -- nonNetRecv (Left = UInt bill amounts given by user, Right = compiled into Tuple of UInts)
+                        DLArg) -- returnVal
```
jeapostrophe, line 763: Small issue... I think you should make a record for this. I find it very tedious to work with tuples in all the different passes. Furthermore, I think it would be nice if we don't enforce them to be present and not all of them. So I really think it should be...

```suggestion
  , ralgo_simValues :: (DLArg, -- netRecv
                        Either (Maybe [DLArg]) DLArg, -- nonNetRecv (Left = UInt bill amounts given by user, Right = compiled into Tuple of UInts)
                        DLArg) -- returnVal
  , ralgo_sim_netRecv :: Maybe DLArg
  , ralgo_sim_bill :: Maybe (Either [DLArg] DLArg) -- not sure about this
  , ralgo_sim_ret :: Maybe DLArg
```

Maybe `netRecv` and `bill` don't need to be `Maybe` and the `0` can be put in by the code generator, but there's no way to represent the `undefined` for `ret`

hs/src/Reach/Eval/Core.hs
```diff
@@ -3680,11 +3680,13 @@ evalPrim p sargs =
       ensure_modes [SLM_ConsensusStep, SLM_ConsensusPure] "remote ALGO"
       metam <- mustBeObject =<< one_arg
       metam' <- mapM (ensure_public . sss_sls) metam
-      let actual = M.keysSet metam'
-      let valid = S.fromList $ [ "fees", "accounts", "assets", "addressToAccount", "apps", "onCompletion", "strictPay", "rawCall" ]
-      unless (actual `S.isSubsetOf` valid) $ do
+      let keys = M.keysSet metam'
+      let validKeys = S.fromList $ [ "fees", "accounts", "assets", "addressToAccount", "apps",
+                                     "onCompletion", "strictPay", "rawCall",
+                                     "simulationValues" ]
```
jeapostrophe, line 3686: I think "simulationValues" is a confusing interface and it should be `sim_netRecv` ... etc

hs/src/Reach/Eval/Core.hs
```diff
@@ -3721,6 +3723,30 @@ evalPrim p sargs =
         Just (SLV_Bytes _ "UpdateApplication") -> return $ RA_UpdateApplication
         Just (SLV_Bytes _ "DeleteApplication") -> return $ RA_DeleteApplication
         Just _ -> expect_ $ Err_Remote_ALGO_extra $ [ "illegal value for onCompletion" ]
+      -- nonNetRecv default depends on call to bill(). bill() might be called later, so we generate
+      -- the default when the remote fn is called (in later branch, where RemoteFunMode = Nothing)
```
jeapostrophe, line 3727: I think this should not work that way and instead it should be `Nothing` until specified. I think that you can't call `bill` after `.ALGO` and when you do, it checks that what you're calling it with is consistent with `bill`

hs/src/Reach/Eval/Core.hs
```diff
@@ -3721,6 +3723,30 @@ evalPrim p sargs =
...
+      let nonNetRecvDef = Left Nothing
+      let netRecvDef = DLA_Literal $ DLL_Int at UI_Word 0
+      let returnValDef = DLA_Literal $ DLL_Null
```
jeapostrophe, line 3730: This is not a good default

jeapostrophe (changes requested):

docs/src/rsh/consensus/index.md
```diff
@@ -618,6 +618,16 @@ In addition, a remote function may be augmented with one of the following operat
...
   + `{!rsh} opts.rawCall` is a boolean (default `{!rsh} false`) that when `{!rsh} true` omits the ABI method selector from the call.
+  + `{!rsh} opts.simNetRecv` is an integer field (default `{!rsh} 0`).
+    The field represents how many network tokens Reach will assume your contract received from the remote call, for the purposes of transaction simulation.
+    This is useful when using `{!rsh} withBill` and `{!rsh} enforce` to enforce that your contract received a certain number network tokens from the remote call.
+  + `{!rsh} opts.simTokensRecv` is a tuple-of-integers field (default all zeros).
+    The field represents how many of each non-network token Reach will assume your contract received from the remote call, for the purposes of transaction simulation.
+    Amounts must be specified in the same order as in the call to `{!rsh} bill` or `{!rsh} withBill`.
+    This is useful when using `{!rsh} withBill` and `{!rsh} enforce` to enforce that your contract received a certain number non-network tokens from the remote call.
+  + `{!rsh} opts.simReturnVal` is a field whose type must match the returned type of the remote function (default unspecified).
```
jeapostrophe, line 628: Isn't the default the "default value" of each type? I think your JS change needs to use that if it is undefined

hs/src/Reach/Backend/JS.hs
```diff
@@ -551,16 +552,20 @@ jsExpr = \case
...
+        let DLRemoteALGO { ralgo_simNetRecv, ralgo_simTokensRecv, ralgo_simReturnVal } = dr_ralgo dr
+        netRecv' <- jsArg ralgo_simNetRecv
+        tokensRecv' <- jsArg $ case ralgo_simTokensRecv of
+          RA_Tuple t -> t
+          _ -> impossible "expected RA_Tuple"
+        returnVal' <- maybe (pure "undefined") jsArg ralgo_simReturnVal
+        let arr = jsArray [netRecv'    <> " /* simNetRecv */",
+                           tokensRecv' <> " /* simTokensRecv */",
+                           returnVal'  <> " /* simReturnVal */"]
+        return $ jsNewScope $ simTxn <> hardline <> jsReturn arr
```
jeapostrophe, line 568: Why new scope?

jeapostrophe: Oh. Cause the comma is dumb?

js/stdlib/ts/ALGO.ts
```diff
@@ -1763,8 +1763,8 @@ const connectAccount = async (networkAccount: NetworkAccount): Promise<Account>
           getOutput: (async <X extends CBR_Val>(o_mode:string, o_lab:string, o_ctc:ALGO_Ty<X>, o_val:X): Promise<X> => {
             void(o_mode);
             void(o_lab);
-            void(o_val);
-            return o_ctc.defaultValue;
```
jeapostrophe, line 1767: This is what I meant. You are now letting an undefined get in the runtime

jeapostrophe: Ideally you'd make a test to show the problem, such as a remote returning a tuple

jeapostrophe (changes requested):

hs/src/Reach/Backend/JS.hs
erer1243: comma was harder to read

js/stdlib/ts/ALGO.ts
erer1243: oh, i was confused about that as well, but forgot that I had changed this. thanks for pointing it out.

hs/src/Reach/Backend/JS.hs
jeapostrophe: Yea, this is much nicer

js/stdlib/ts/ALGO.ts
jeapostrophe: Noice

jeapostrophe (approved):

jeapostrophe:
This actually isn't passing CI yet, because the log-attack needs new numbers

jeapostrophe (approved):

PR #1466, opened 2022-10-14
CORE-2531: Switch Fusion

chrisnevers:
The comments should outline the process but let me know if I need to explain it more. 

One augmentation that could be handy in the future is to add metadata to `DL_LocalSwitch` to track if it sets a DLVar, like we do for floating `if`. Then, if we inline a switch that sets a DLVar, float that `DL_Var` too. Right now, it won't lift a switch if it sets a var in its branches.

### DIFF
--- a/hs/src/Reach/Optimize.hs
+++ b/hs/src/Reach/Optimize.hs
@@ -978,17 +978,19 @@ fuseStmt fuseEnv mk t =
       -- Now, look in the env to see if there are any
       -- subsequent switches to fuse into this stmt.
       k' <- fuseSwitch mk at dv scs k
+      fusionCasesR <- asks eFusionCases
       case isCom k' of
         Just (DL_LocalSwitch _ _ scs1, k1) | M.member dv fuseEnv -> do
           -- If there are previous switches looking to fuse,
           -- reset the fusionCase env to how it was
           -- before we processed this branch, add the info
           -- for this switch to the env, and drop this stmt
-            fusionCasesR <- asks eFusionCases
             liftIO $ writeIORef fusionCasesR $ M.insert dv scs1 fuseEnv
             return $ mk (DL_Nop at) k1
         -- If there are no previous switches looking to fuse, keep this stmt and return
-        _ -> return k'
+        _ -> do
+          liftIO $ writeIORef fusionCasesR fuseEnv
+          return k'
     _ -> return t
 
 -- Optimize is generally top-down, but this is locally bottom-up. We float/fuse after optimizing the tail
--- a/hs/src/Reach/Optimize.hs
+++ b/hs/src/Reach/Optimize.hs
@@ -20,6 +20,7 @@ import Safe (atMay)
 import qualified Data.ByteString as B
 import Reach.CollectSvs
 import Reach.AST.SL
+import Reach.Subst
 
 type App = ReaderT Env IO
 
@@ -80,6 +81,7 @@ data Env = Env
   , eSimulate :: Bool
   , eSvs :: S.Set DLVar
   , eVarsSet ::IORef (M.Map DLVar Int)
+  , eFusionCases :: IORef (M.Map DLVar (SwitchCases DLTail))
   }
 
 mkVar :: SrcLoc -> DLType -> App DLVar
@@ -134,7 +136,10 @@ newScope m = do
   Env {..} <- ask
   eEnvsR' <- liftIO $ dupeIORef eEnvsR
   eVarsSet' <- liftIO $ dupeIORef eVarsSet
-  local (\e -> e {eEnvsR = eEnvsR', eVarsSet = eVarsSet'}) m
+  eFusionCases' <- liftIO $ dupeIORef eFusionCases
+  local (\e -> e { eEnvsR = eEnvsR'
+                  , eVarsSet = eVarsSet'
+                  , eFusionCases = eFusionCases' }) m
 
 lookupCommon :: Ord a => (CommonEnv -> M.Map a b) -> a -> App (Maybe b)
 lookupCommon dict obj = do
@@ -236,6 +241,7 @@ mkEnv0 eCounter eDroppedAsserts eConst eParts eMaps eSimulate eSvs = do
           map (\x -> (x, mempty)) $ F_Ctor : F_All : F_Consensus : map F_One eParts
   eEnvsR <- liftIO $ newIORef eEnvs
   eVarsSet <- liftIO $ newIORef mempty
+  eFusionCases <- liftIO $ newIORef mempty
   return $ Env {..}
 
 maybeClearMaps :: App a -> App a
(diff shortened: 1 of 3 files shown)


hs/src/Reach/Optimize.hs
```diff
@@ -920,11 +926,72 @@ preFloat = \case
...
+    Just (DL_LocalSwitch at dv scs, k) -> do
+      -- We've processed everything in the tail.
+      -- Now, look in the env to see if there are any
+      -- subsequent switches to fuse into this stmt.
+      k' <- fuseSwitch mk at dv scs k
+      case isCom k' of
+        Just (DL_LocalSwitch _ _ scs1, k1) | M.member dv fuseEnv -> do
+          -- If there are previous switches looking to fuse,
+          -- reset the fusionCase env to how it was
+          -- before we processed this branch, add the info
```
jeapostrophe, line 980: How does it get reset? It looks like it only ever has an insert?
Or do you mean because we insert into `fuseEnv` rather than using `modifyIORef`?

hs/src/Reach/Optimize.hs
```diff
@@ -920,11 +926,72 @@ preFloat = \case
...
+          -- for this switch to the env, and drop this stmt
+            fusionCasesR <- asks eFusionCases
+            liftIO $ writeIORef fusionCasesR $ M.insert dv scs1 fuseEnv
+            return $ mk (DL_Nop at) k1
+        -- If there are no previous switches looking to fuse, keep this stmt and return
+        _ -> return k'
```
jeapostrophe, line 986: Shouldn't this always reset to `fuseEnv`?

jeapostrophe: If so, I feel like I want to think of a test case that fails with this mistake?

hs/src/Reach/Optimize.hs
```diff
@@ -920,11 +926,72 @@ preFloat = \case
...
+  fusionCases <- liftIO . readIORef =<< asks eFusionCases
+  return $ M.lookup dv fusionCases
+
+preFuse :: DLStmt -> App (M.Map DLVar (SwitchCases DLTail))
+preFuse s = do
+  oldEnv <- liftIO . readIORef =<< asks eFusionCases
+  case s of
+    DL_LocalSwitch _ dv _ -> updateFusion dv mempty
+    DL_Let {} -> cantFuse
+    DL_Var {} -> cantFuse
```
jeapostrophe, line 945: You should make a comment that we could if they weren't bound, but that we're not trying to track that yet

hs/src/Reach/Optimize.hs
```diff
@@ -920,11 +926,72 @@ preFloat = \case
...
+updateFusion dv scs = do
+  fusionCasesR <- asks eFusionCases
+  liftIO $ modifyIORef' fusionCasesR $ M.insert dv scs
+
+readFusionCases :: DLVar -> App (Maybe (SwitchCases DLTail))
+readFusionCases dv = do
+  fusionCases <- liftIO . readIORef =<< asks eFusionCases
+  return $ M.lookup dv fusionCases
+
+preFuse :: DLStmt -> App (M.Map DLVar (SwitchCases DLTail))
```
jeapostrophe, line 939: I think you should define the `FuseEnv` type

hs/src/Reach/Optimize.hs
```diff
@@ -920,11 +926,72 @@ preFloat = \case
...
+  where
+    ret cases = return $ mk (DL_LocalSwitch at dv cases) k
+    fuse cases c (v1, b1, t1) =
+      case M.lookup c cases of
+        Just (v2, _, t2) -> do
+          -- The branch we're inling binds `v2`.
+          -- The branch we're inling *into* binds `v1`.
+          -- Substitute occurrences of `v2` in `t2` with `v1`.
+          let t2' = subst_ (M.fromList [(v2, v1)]) t2
+          (v1, b1, dtReplace DT_Com t2' t1)
```
jeapostrophe, line 965: I think it is simpler to introduce `v2 := v1` and let other optimizations take care of it. Maybe I'm wrong and that is expensive.

jeapostrophe (changes requested):
- Does it apply to any programs in the corpus?
- How come it doesn't apply to C or ET Switches?
- Can you make a new card for that var tracking idea if it is too hard to just do?

chrisnevers:
It applies to 21 tests, I will try to do some benchmarks later. 

By C & ET Switch, do you mean `CT_Switch` and `ET_Switch`? If so, I'm not sure how fusion would work for them because the tails of the program are inside the branches

hs/src/Reach/Optimize.hs
chrisnevers: The only way the current env would differ from `fuseEnv` is the value associated with `dv`. However, if `dv` is not in the map (which is this case), it's not really necessary to reset the env, because nothing will try to look at it further up the stack

chrisnevers:
JIRA ticket for the enhancement mentioned https://reachsh.atlassian.net/browse/CORE-2664

hs/src/Reach/Optimize.hs
```diff
@@ -134,7 +137,10 @@ newScope m = do
   Env {..} <- ask
   eEnvsR' <- liftIO $ dupeIORef eEnvsR
   eVarsSet' <- liftIO $ dupeIORef eVarsSet
-  local (\e -> e {eEnvsR = eEnvsR', eVarsSet = eVarsSet'}) m
+  eFusionCases' <- liftIO $ dupeIORef eFusionCases
+  local (\e -> e { eEnvsR = eEnvsR'
+                  , eVarsSet = eVarsSet'
+                  , eFusionCases = eFusionCases' }) m
```
jeapostrophe, line 143:
```suggestion
                 , eVarsSet = eVarsSet'
                 , eFusionCases = eFusionCases' }) m
```

hs/src/Reach/Optimize.hs
jeapostrophe: I am too nervous. I think you should do
```
liftIO $ writeIORef fusionCasesR fuseEnv
return k'
```

jeapostrophe: If not, then I don't see why you don't juse do `modifyIORef` on line 988

jeapostrophe (changes requested):
I want to see those benchmarks.

Regarding the C/E switches. You are correct. I don't know how often it would apply, but it would be possible to turn something like
```
localSwitch (x) {
...
}
switch(x) {
...
}
```
into a single `switch`.

This code moves the bottom switch UP but a transformation that did the above would move the top switch DOWN. It would duplicate code in between the two switches or move them:
```
switch (x) { case0: { c0_0... }, ... }; switch (x) { case0: { c1_0... }, ... };
=>
switch (x) { case0: { c0_0... c1_0... }, ... }

switch (x) { case0: { c0_0... }, ... }; y; z
=>
switch (x) { case0: { c0_0... y }, ... }; z
```

Maybe the time to do this is when you KNOW that there is a later non-local switch that dispatches on the variable. Given that we know how Reach does code generation, that's going to be an API dispatch, so this is basically a way to force that we immediately dispatch on the API and then optimize away all the other switches.

jeapostrophe:
(BTW, I don't expect you to implement my long comment about a different way to do it.)

jeapostrophe (approved):

PR #1471, opened 2022-10-17
js:ETH - ignore "after last accepted block"

DanBurton:
Not tested yet. Not sure if there's a good way to add anything to our examples or test suite for this.

### DIFF
--- a/js/stdlib/ts/ETH_like.ts
+++ b/js/stdlib/ts/ETH_like.ts
@@ -259,7 +259,7 @@ const newEventQueue = (): EventQueue => {
     } catch (e) {
       const es = `${e}`;
       debug(dhead, `err`, e, es);
-      if ( es.includes('Unable to find block hash') ) {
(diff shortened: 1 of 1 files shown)


jeapostrophe (dismissed):

jeapostrophe:
It is HORRIFYING that there is a space on the LEFT but not on the RIGHT

js/stdlib/ts/ETH_like.ts
```diff
@@ -259,7 +259,7 @@ const newEventQueue = (): EventQueue => {
     } catch (e) {
       const es = `${e}`;
       debug(dhead, `err`, e, es);
-      if ( es.includes('Unable to find block hash') ) {
+      if ( es.includes('Unable to find block hash') || es.includes('after last accepted block')) {
```
DanBurton, line 262:
```suggestion
      if ( es.includes('Unable to find block hash') || es.includes('after last accepted block') ) {
```

PR #1472, opened 2022-10-18
WIP - working on BytesDyn casting

willghatch:
I wanted to push this branch to get some comments.

(1) I think I can't call the casting function `BytesDyn` because in the default environment `BytesDyn` returns a type, not an `SLPrim`.  I could probably call it `BytesDyn.cast()` by adding the the type to `evalAsEnvM`.  I guess that's probably what I'll ultimately do, as there is precident with eg. `Token` type being there, although I decided for now to work on it as `BytesDynCast`.  Do I understand everything correctly here?

(2) I currently have an implementation of interpreting this primitive that I think should make sense, but when I use it on the ERC721 contract (replacing the `EmptyBytesDyn` variable with `BytesDynCast("")`) and compile it I get an error because it generates (in Solidity) two copies of the `ERC721TokenReceiver` interface, one with `bytes memory` and one with `bytes0`.  It seems like something is seeing through the layer where I'm setting the type as `BytesDyn` to see the type `Bytes(0)` instead.  My plan for implementation was to use `ctxt_lift_expr` with `DLE_Arg` to just make a new DLVar that references the old one but has the new type.  But it looks like that's probably not the right approach.  Do you have a recommendation?

### DIFF
--- a/docs/src/changelog/index.md
+++ b/docs/src/changelog/index.md
@@ -6,6 +6,7 @@ Versions and changes-within-versions are listed in reverse-chronological order: 
 ## 0.1.12: 2022/09 - present
 
 @{verRC("0.1.12")}
++ 2022/10/25: Added `{!rsh} BytesDyn` casting from fixed-length bytes.
 
 @{rcHead("0.1.12-rc.5")}
 
--- a/docs/src/rsh/compute/index.md
+++ b/docs/src/rsh/compute/index.md
@@ -2976,7 +2976,7 @@ The `IDs` component is intended to hold byte strings that represent ERC-165 inte
 For example, we define all of the interfaces that are part of the [ERC-721](https://eips.ethereum.org/EIPS/eip-721) NFT specification using `{!rsh} mixin`:
 ```reach
 load: /examples/ERC721/index.rsh
-md5: fc2bac5972751cb3280a6e974d864096
+md5: c31c20d4e9add1a3ef3038596d965089
 range: 8 - 59
 ```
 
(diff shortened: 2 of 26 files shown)


jeapostrophe:
1. We already support calling types as functions: https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Eval/Core.hs#L4420-L4421

2. I don't see an example program and so it is hard for me to track things. You're going to get into trouble because you haven't changed `dlvToDL` (https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/Eval/Core.hs#L693)

willghatch:
I haven't rearranged things to address (1) yet, but I've now addressed (2) and I'm stuck with the same issue.  But I have a little more context for it now.  I see now that the generated solidity for ERC721 even before the change included two Solidity interface definitions for a single Reach interface definition, specifically for `onERC721Received`.  I guess this is happening due to inlining/linearization, but it surprises me that it doesn't get a cached single definition.  Anyway, with the change (that I pushed this time) of the ERC721 contract to write an empty BytesDyn instead of using one passed in, one of those solidity definitions changes from using `bytes` as its type to `bytes0`.  I can see the code in `ETH_Solidity` for `DLE_Remote` is using the arguments to generate the type (rather than using the interface definition), so it makes sense why it's generating `bytes0` if that's the type it sees there.  But... how is it passing the type check that the argument is a `BytesDyn` but then showing up as `Bytes[0]` at the level of solidity generation?  I haven't figured that out yet.  So my implementation so far seems to be able to give that expression the type `BytesDyn` in some but not all places...

jeapostrophe (commented):
If you look at the intermediate files, you'll be able to see type annotations on all of the variables at every step

willghatch:
So you said you only cared about the static case, but by then I had figured out how to do the dynamic case, and I had assumed a dynamic case initially and had already done much of the work for it.  So I ultimately implemented it for both cases, although I got stuck for a while on the SMT part for the dynamic case, which I hadn't thought about previously...

jeapostrophe (changes requested):
All that is left is the changelog. Good work.

willghatch:
Ah, sorry, I always forget about the changelog.

jeapostrophe (approved):

PR #1473, opened 2022-10-18
Add `acc.getDebugLabel`

chrisnevers:
Makes it easier for frontends to do custom logging when given just an `Account` value

### DIFF
--- a/docs/src/changelog/index.md
+++ b/docs/src/changelog/index.md
@@ -6,7 +6,8 @@ Versions and changes-within-versions are listed in reverse-chronological order: 
 ## 0.1.12: 2022/09 - present
 
 @{verRC("0.1.12")}
-+ 2022/09/19: Added `{!js} ctc.appOptIn` and `{!js} account.appOptedIn`
++ 2022/10/18: Added `${!js} acc.getDebugLabel`.
++ 2022/10/19: Added `{!js} ctc.appOptIn` and `{!js} account.appOptedIn`
 
 @{rcHead("0.1.12-rc.5")}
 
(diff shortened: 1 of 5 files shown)


docs/src/changelog/index.md
```diff
@@ -6,7 +6,8 @@ Versions and changes-within-versions are listed in reverse-chronological order:
 ## 0.1.12: 2022/09 - present
 
 @{verRC("0.1.12")}
-+ 2022/09/19: Added `{!js} ctc.appOptIn` and `{!js} account.appOptedIn`
++ 2022/10/18: Added `${!js} acc.getDebugLabel`.
```
jeapostrophe, line 9:
```suggestion
+ 2022/10/18: Added `{!js} acc.getDebugLabel`.
```

docs/src/frontend/index.md
```diff
@@ -639,6 +639,14 @@ In this snippet, `{!js} getAddress` obtains the addresses of `Alice` and `Bob`,
 Next, each address is logged so that participants can see both addresses.
 This can be useful to verify that the address receiving the payment is the correct address.
 
+---
+@{ref("js", "getDebugLabel")}
+```js
+acc.getDebugLabel() => string
+```
+
+Returns the label used to distinguish an account in debug logs. If no label is provided, then the first four digits of the account address will be used.
```
jeapostrophe, line 648:
```suggestion
Returns the label used to distinguish an account in debug logs.
If no label was previously provided with `{!js} acc.setDebugLabel`, then the first four digits of the account address will be used.
```

jeapostrophe (changes requested):

PR #1474, opened 2022-10-18
Type mismatch location

willghatch:
For CORE-2482.  When there is a type mismatch error, this patch makes it display location information for the source of the expected and actual types, at least for some instances of type mismatches.

### DIFF
--- a/hs/src/Reach/Eval/Core.hs
+++ b/hs/src/Reach/Eval/Core.hs
@@ -974,8 +974,8 @@ compileTypeOf v = do
 compileTypeOfs :: [SLVal] -> App ([DLType], [DLArg])
 compileTypeOfs vs = unzip <$> mapM compileTypeOf vs
 
-compileCheckType :: DLType -> SLVal -> App DLArg
-compileCheckType et v = compileArgExpr_ =<< typeCheck_d Nothing et v
+compileCheckType :: Maybe SrcLoc -> DLType -> SLVal -> App DLArg
+compileCheckType el et v = compileArgExpr_ =<< typeCheck_d el et v
 
 compileToVar :: SLVal -> App DLVar
 compileToVar v = do
@@ -1699,11 +1699,11 @@ compileTimeArg :: SLVal -> App DLTimeArg
 compileTimeArg = \case
   SLV_Data _ dm "Left" v
     | correctData dm ->
-      Left <$> compileCheckType t v
+      Left <$> compileCheckType Nothing t v
   SLV_Data at dm "Right" v
     | correctData dm -> do
       liftIO $ emitWarning (Just at) $ W_NetworkSeconds
-      Right <$> compileCheckType t v
+      Right <$> compileCheckType Nothing t v
   SLV_DLVar (DLVar _ _ (T_Data dm) _)
     | correctData dm ->
       expect_ Err_TimeArg_NotStatic
@@ -2165,9 +2165,9 @@ evalDistinctTokens old = \case
 evalITE :: SecurityLevel -> SLVal -> SLVal -> SLVal -> App SLSVal
 evalITE lvl c t f = do
   at <- withAt id
-  ca <- compileCheckType T_Bool c
+  ca <- compileCheckType (Just at) T_Bool c
   (tt, ta) <- compileTypeOf t
-  fa <- compileCheckType tt f
+  fa <- compileCheckType (Just $ srclocOf t) tt f
   dv <- ctxt_lift_expr (DLVar at Nothing tt) (DLE_PrimOp at IF_THEN_ELSE [ca, ta, fa])
   return $ (lvl, SLV_DLVar dv)
 
@@ -2470,7 +2470,7 @@ doFluidSet_ fv da = do
 doFluidSet :: FluidVar -> SLSVal -> App ()
 doFluidSet fv ssv = do
   sv <- ensure_public ssv
-  da <- compileCheckType (fluidVarType fv) sv
+  da <- compileCheckType Nothing (fluidVarType fv) sv
   doFluidSet_ fv da
 
 doBaseWaitUpdate :: DLTimeArg -> App ()
@@ -2564,7 +2564,7 @@ tokenMetaAssert tm tok lhs op msg = do
 assumeOp :: B.ByteString -> SPrimOp -> DLVar -> SLVal -> App ()
 assumeOp lab op v bound = do
   cmp <- evalPrimOp op [public $ SLV_DLVar v, public bound]
-  cmp_v <- compileCheckType T_Bool $ snd cmp
+  cmp_v <- compileCheckType Nothing T_Bool $ snd cmp
   doClaim CT_Assume cmp_v $ Just ("assume " <> lab)
 
 assumeGtZero :: DLVar -> App ()
@@ -2581,8 +2581,8 @@ assumeBalanceUpdate :: SLVal -> SLVal -> (PrimOp, DLVar -> App b) -> App b
 assumeBalanceUpdate balV amtV (op, assumeF) = do
   at <- withAt id
   let tInt = T_UInt UI_Word
-  bal_dv <- compileCheckType tInt balV
-  amt_dv <- compileCheckType tInt amtV
+  bal_dv <- compileCheckType Nothing tInt balV
+  amt_dv <- compileCheckType Nothing tInt amtV
   -- Avoid using `evalPrimOp` to avoid generating claims for verify arithmetic
   bal' <- ctxt_lift_expr (DLVar at Nothing tInt) $ DLE_PrimOp at op [bal_dv, amt_dv]
   assumeF bal'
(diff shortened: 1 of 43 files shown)


hs/src/Reach/Eval/Core.hs
```diff
@@ -675,16 +675,16 @@ jsClo at name js env_ = SLV_Clo at Nothing $ SLClo (Just name) args body cloenv
...
 
-typeEq :: DLType -> DLType -> App ()
-typeEq e a = do
+typeEq :: DLType -> DLType -> Maybe SrcLoc -> Maybe SrcLoc -> App ()
+typeEq e a mel mal = do
   unless (typeEqb e a) $
-    expect_ $ Err_Type_Mismatch e a
+    expect_ $ Err_Type_Mismatch e a mel mal
 
 typeEqs :: [DLType] -> App DLType
```
jeapostrophe, line 683: If you're gonna do it... do it right!
```suggestion
typeEqs :: [(Maybe SrcLoc, DLType)] -> App DLType
```

willghatch:
I'll note that the ticket wants it to be reported as “nth arg to f”, but that seems a lot harder and also less general than showing source locations.

hs/src/Reach/Eval/Core.hs
```diff
@@ -891,7 +891,7 @@ applyType ct v = \case
 typeCheck_s :: ClaimType -> SLType -> SLVal -> App DLArgExpr
 typeCheck_s ct st val = do
   dt <- st2dte st
-  res <- typeCheck_d dt val
+  res <- typeCheck_d Nothing dt val
```
jeapostrophe, line 894: You could make this take an argument. I think you should define `type MSrcLoc = Maybe SrcLoc`, because it is so brutal to type out

hs/src/Reach/Eval/Core.hs
```diff
@@ -1699,11 +1699,11 @@ compileTimeArg :: SLVal -> App DLTimeArg
 compileTimeArg = \case
   SLV_Data _ dm "Left" v
     | correctData dm ->
-      Left <$> compileCheckType t v
+      Left <$> compileCheckType Nothing t v
```
jeapostrophe, line 1702: These `t`'s location is "by definition"

hs/src/Reach/Eval/Core.hs
```diff
@@ -2338,7 +2339,7 @@ evalPrimOp sp sargs = do
     make_var dom rng args' = do
       at <- withAt id
       args'e <-
-        mapM (uncurry typeCheck_d)
+        mapM (uncurry (typeCheck_d Nothing))
```
jeapostrophe, line 2342: More "by definition"

hs/src/Reach/Eval/Core.hs
```diff
@@ -2563,7 +2564,7 @@ tokenMetaAssert tm tok lhs op msg = do
 assumeOp :: B.ByteString -> SPrimOp -> DLVar -> SLVal -> App ()
 assumeOp lab op v bound = do
   cmp <- evalPrimOp op [public $ SLV_DLVar v, public bound]
-  cmp_v <- compileCheckType T_Bool $ snd cmp
+  cmp_v <- compileCheckType Nothing T_Bool $ snd cmp
```
jeapostrophe, line 2567: All ditto

hs/src/Reach/Eval/Core.hs
```diff
@@ -2580,8 +2581,8 @@ assumeBalanceUpdate :: SLVal -> SLVal -> (PrimOp, DLVar -> App b) -> App b
 assumeBalanceUpdate balV amtV (op, assumeF) = do
   at <- withAt id
   let tInt = T_UInt UI_Word
-  bal_dv <- compileCheckType tInt balV
-  amt_dv <- compileCheckType tInt amtV
+  bal_dv <- compileCheckType Nothing tInt balV
+  amt_dv <- compileCheckType Nothing tInt amtV
```
jeapostrophe, line 2585: Ditto

hs/src/Reach/Eval/Core.hs
```diff
@@ -2663,7 +2664,7 @@ mustBeObjectTy err = \case
 mustBeUIntTy :: DLType -> App DLType
 mustBeUIntTy ty = case ty of
   T_UInt {} -> return ty
-  _ -> expect_ $ Err_Type_Mismatch (T_UInt UI_Word) ty
+  _ -> expect_ $ Err_Type_Mismatch (T_UInt UI_Word) ty Nothing Nothing
```
jeapostrophe, line 2667: Push the args back onto `mustBeUIntTy`?

hs/src/Reach/Eval/Core.hs
```diff
@@ -2771,13 +2772,13 @@ evalPrim p sargs =
           [x] -> return (x, Nothing)
           [x, y] -> return (x, Just y)
           _ -> illegal_args
-      toka <- compileCheckType T_Token tokv
       at <- withAt id
+      toka <- compileCheckType (Just at) T_Token tokv
```
jeapostrophe, line 2776: This is actually a "by definition", unless you want to use the srcloc of the call as a the source for "by definition"s in general

jeapostrophe (changes requested):
Cool. A few comments.

hs/src/Reach/Eval/Core.hs
willghatch: Yeah, I used the call as the source for things that were “by definition” generally.  I didn't hit all of them just because I wasn't immediately sure what each type check was for, and wanted to get to a PR to get feedback before spending more time figuring them out.  I think I'll switch the ones where I used the call site to have a “definition” source location.

hs/src/Reach/Eval/Core.hs
willghatch: Ah, I dropped some branching paths in my quest to push args up to where they needed to be used.

willghatch:
There are a couple of places where `typeCheck_s` is used where I'm not sure what expected source to give:  (1) for Data variants and (2) for refined args.  So I'm leaving them as `Nothing` for now until you tell me otherwise.

hs/src/Reach/Eval/Core.hs
```diff
@@ -3410,7 +3410,7 @@ evalPrim p sargs =
       vv <- case (vt, args) of
         (ST_Null, []) -> return $ SLV_Null at "variant"
         _ -> one_arg
-      void $ typeCheck_s CT_Assert vt vv
+      void $ typeCheck_s CT_Assert Nothing vt vv
```
jeapostrophe, line 3413: This should be the srcloc of the definition of the variant. I don't think that is actually tracked, but could be

hs/src/Reach/Eval/Core.hs
```diff
@@ -4295,7 +4295,7 @@ assertRefinedArgs :: ClaimType -> [SLSVal] -> SrcLoc -> SLTypeFun -> App (SLVal,
 assertRefinedArgs ct sargs iat (SLTypeFun {..}) = do
   let argvs = map snd sargs
   arges <-
-    mapM (uncurry $ typeCheck_s ct)
+    mapM (uncurry $ typeCheck_s ct Nothing)
```
jeapostrophe, line 4298: This would be `stf_at` (I think that exists). It is because of the type of refinement. It might be `iat`

hs/src/Reach/Eval/Core.hs
```diff
@@ -5209,7 +5209,7 @@ doToConsensus ks (ToConsensusRec {..}) = locAt slptc_at $ do
   let tc_send = fmap snd tc_send'
   let msg_dass = fmap ds_msg tc_send
   let msg_dass_t = transpose $ M.elems $ msg_dass
-  let get_msg_t = typeEqs . map argTypeOf
+  let get_msg_t = typeEqs . map (\x -> (Nothing, argTypeOf x))
   msg_ts <- mapM get_msg_t msg_dass_t
```
jeapostrophe, line 5213: `tc_send` is the ToConsensus send structure, which is an instance of a `DLSend` --- https://github.com/reach-sh/reach-lang/blob/master/hs/src/Reach/AST/DLBase.hs#L1390 --- that structure represents their piece of the `publish` (constructed on line 5204)

The correct source location for this `x`'s type is in the result of the `evalId` on line 5201.

Basically, we want to change
```
                ds_msg <- mapM (\v -> snd <$> (compileTypeOf =<< ensure_public =<< evalId "publish msg" v)) msg
```
into
```
                (ds_msg_at, ds_msg) <- unzip <$> mapM (\v -> do
                  sv <- evalId "publish msg" v
                  va <- snd <$> (compileTypeOf =<< ensure_public =<< sv)
                  return (srcLocOf sv, va)) msg
```
(Or make it so that the type of `ds_msg` is `[(SrcLoc, DLArg)]` rather than `[DLArg]`.

jeapostrophe (changes requested):

willghatch:
I've added all of those, but they don't actually affect any of our `t/n` golden files as it happens.

hs/src/Reach/Eval/Core.hs
```diff
@@ -5191,27 +5192,34 @@ doToConsensus ks (ToConsensusRec {..}) = locAt slptc_at $ do
...
-        (only_lifts, res) <-
+        (only_lifts, (res, res_ats)) <-
           locWho who $
             locSt st_lpure $
               captureLifts $ do
                 evalChecks
                 let repeat_dv = M.lookup who pdvs
                 ds_isClass <- is_class who
-                ds_msg <- mapM (\v -> snd <$> (compileTypeOf =<< ensure_public =<< evalId "publish msg" v)) msg
+                --ds_msg <- mapM (\v -> snd <$> (compileTypeOf =<< ensure_public =<< evalId "publish msg" v)) msg
```
jeapostrophe, line 5202:
```suggestion
```

PR #1475, opened 2022-10-18
fix docs about Token.burn

willghatch:
I noticed a discrepancy between the docs and the implementation of Token.burn.  I've updated the docs according to my understanding from seeing the implementation.

### DIFF
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -517,7 +517,7 @@ Reach assumes that network tokens and non-network tokens behave identically, but
 
 ---
 
-`{!rsh} Token.burn(tok, amt)`, or `{!rsh} tok.burn(amt)`, where `{!rsh} tok` is a `{!rsh} Token` value and `{!rsh} amt` is a `{!rsh} UInt` value, may be used to @{defn("burn")} tokens in the contract account, meaning that they are utterly destroyed and can never be recovered.
+`{!rsh} Token.burn(tok, amt)`, or `{!rsh} tok.burn(tok)`, where `{!rsh} tok` is a `{!rsh} Token` value and `{!rsh} amt` is a `{!rsh} UInt` value, may be used to @{defn("burn")} tokens in the contract account, meaning that they are utterly destroyed and can never be recovered.  If `{!rsh} amt` is not given, the current balance of the token will be used.
 
 ---
 
--- a/docs/src/rsh/consensus/index.md
+++ b/docs/src/rsh/consensus/index.md
@@ -517,7 +517,8 @@ Reach assumes that network tokens and non-network tokens behave identically, but
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/src/rsh/consensus/index.md
```diff
@@ -517,7 +517,7 @@ Reach assumes that network tokens and non-network tokens behave identically, but
 
 ---
 
-`{!rsh} Token.burn(tok, amt)`, or `{!rsh} tok.burn(amt)`, where `{!rsh} tok` is a `{!rsh} Token` value and `{!rsh} amt` is a `{!rsh} UInt` value, may be used to @{defn("burn")} tokens in the contract account, meaning that they are utterly destroyed and can never be recovered.
+`{!rsh} Token.burn(tok, amt)`, or `{!rsh} tok.burn(tok)`, where `{!rsh} tok` is a `{!rsh} Token` value and `{!rsh} amt` is a `{!rsh} UInt` value, may be used to @{defn("burn")} tokens in the contract account, meaning that they are utterly destroyed and can never be recovered.  If `{!rsh} amt` is not given, the current balance of the token will be used.
```
jeapostrophe, line 520: The first change from `tok` to `amt` seems wrong. Does the code really do that?

willghatch: It takes 1 or 2 args, it always takes a token and optionally takes an amount.  If the amount is not given it uses `getBalanceOf'` on the token.

https://github.com/reach-sh/reach-lang/blob/c2ca1fbc3820cd3641969a7402f9018438b64b69/hs/src/Reach/Eval/Core.hs#L2768

willghatch: Ah, let me try to make a range instead of one line...

https://github.com/reach-sh/reach-lang/blob/c2ca1fbc3820cd3641969a7402f9018438b64b69/hs/src/Reach/Eval/Core.hs#L2768-L2781

willghatch: Note those permalink URLs based on the revision hash *wink* *wink*.

willghatch: Oh, now that I look again, I think you are right -- I didn't notice that it's used as a method call on `tok` instead of as a method call.  So the original was incomplete while my change is wrong.

jeapostrophe: The kicker is this line
https://github.com/reach-sh/reach-lang/blob/c2ca1fbc3820cd3641969a7402f9018438b64b69/hs/src/Reach/Eval/Core.hs#L1397
which makes it so that the first argument is implicitly passed as the variable itself.

jeapostrophe, line 520:
```suggestion
`{!rsh} Token.burn(tok, amt?)`, or `{!rsh} tok.burn(amt?)`, where `{!rsh} tok` is a `{!rsh} Token` value and `{!rsh} amt` is a `{!rsh} UInt` value, may be used to @{defn("burn")} tokens in the contract account, meaning that they are utterly destroyed and can never be recovered. 
If `{!rsh} amt` is not given, the current balance of the token will be used.
```

jeapostrophe (changes requested):

PR #1477, opened 2022-10-19
DEVX-540-transfer-&-ethers-ex

TheChronicMonster:
Update `transfer` to say that `to` can be an `account` or an `address`

Moved ethers example directly under ethereum provider specific explanation.

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -367,6 +367,14 @@ This function's API is considered unstable.
 On Ethereum, `{!js} provider` is an instance of `{!js} ethers.provider`.
 See: [https://docs.ethers.io/v5/api/providers/provider/](https://docs.ethers.io/v5/api/providers/provider/)
 
+Example:
+
+```js
+load: /examples/ganache/index.mjs
+md5: cca336d85844697eb44884226f9d2ce0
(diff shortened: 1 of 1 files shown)


docs/src/frontend/index.md
```diff
@@ -791,8 +791,8 @@ On networks that do not, this will always return zero.
 stdlib.transfer(from: Account, to: Account | Address, amount, token?: Token, opts?: TransferOpts) => Promise<void>
 ```
 
-Performs a transfer of `{!js} amount` from `{!js} from` to `{!js} to`,
-which are accounts, such as those returned by `{!js} connectAccount`.
+Performs a transfer of `{!js} amount` from `{!js} from`, which is an account, such as those returned by `{!js} connectAccount`;
+to `{!js} to`, which can be an account or an address.
```
chrisnevers, line 795: I think this a bit verbose. How about something like: 

Performs a transfer of `amount` from `from` to `to`, where `from` is an account and `to` is either an account or address.

chrisnevers (approved):

TheChronicMonster:
All the tests pass, now!

jeapostrophe (approved):

PR #1479, opened 2022-10-19
change isDataVariant test to use different length keys

willghatch:
Proving to myself and everybody that I was wrong and temporarily crazy when I wrote that I made a mistake in my implementation of this feature the first time.  I don't remember what I did that made me think there was a bug here.  But now I can rest assured in the knowledge of my own infallibility when writing code, and can view with skepticism any claims from myself or others that there is ever any problem with my code at any time for any reason.


jeapostrophe:
:)

PR #1480, opened 2022-10-19
erc20: make connector-generic tests

willghatch:
The tests that we had are specific to the ETH connector for the purpose of comparing to OpenZeppelin.  But I thought we should have connector-generic tests because they will (A) show that the contract works on Algorand once it adds box storage, and (B) I think it will be easier for tutorials based on the ERC20 with “normal” generic tests instead of tests that rely on the ethers library.


willghatch:
To be clear: I basically duplicated the existing tests, but in a way that is generic for Reach connectors.  I did this after I was talking to Nick about how a tutorial based on the ERC20 might be weird in terms of testing with its reliance on the ethers library.

PR #1485, opened 2022-10-25
DEVX-605-rsvp-gas

TheChronicMonster:
Added gas limits to RSVP-6-vevt
Changed `ETH` `err` value from `transaction may fail` to `transaction failed`

### DIFF
--- a/docs/src/tut/rsvp/index.md
+++ b/docs/src/tut/rsvp/index.md
@@ -795,7 +795,10 @@ range: 59-70
 ```
 
 We've added a gas limit using `{!js} stdlib.setGasLimit` for testing on Ethereum.
-This prevents `{!rsh} API`s from expending all of their tokens on gas fees.
+Normally the gas limit is computed by looking at the contract code, but it is not always accurate. 
+There is always a gas limit, even if we don't specify anything.
+We've set the "limit" to the largest possible amount of gas that could ever be spent. 
+Thus, we're removing the limit by setting it to be "infinity".
 
 Next, we modify the `willGo` function so that the Guest, before they register, inspects the event details and looks at the reservation price.
 This is representative of a user interface that informs the Guest of how much they'll be expected to pay to make a reservation.
--- a/docs/src/tut/rsvp/index.md
+++ b/docs/src/tut/rsvp/index.md
@@ -776,8 +776,8 @@ Instead, we'll just make a few simple changes to the test framework to demonstra
 
 ```
 load: /examples/rsvp-6-vevt/index.mjs
-md5: 89fe2faa0411e3d6af06b4ed081ba2aa
-range: 49-59
+md5: 65e69ecdbcea0eab0d04838833ddea55
+range: 47-57
 ```
 
 First, we'll have the Host monitor all registrations for this event.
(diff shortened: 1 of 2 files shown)


examples/rsvp-6-vevt/index.mjs
```diff
@@ -14,6 +15,9 @@ const makeRSVP = async ({ hostLabel, name, reservation, timeLimit }) => {
   const sbal = stdlib.parseCurrency(100);
   const accHost = await stdlib.newTestAccount(sbal);
   accHost.setDebugLabel(hostLabel);
+  if (stdlib.connector != 'ALGO') {
```
jeapostrophe, line 18: If I recall, there's no danger to setting it anyways, it just gets ignored

examples/rsvp-6-vevt/index.mjs
```diff
@@ -57,6 +61,9 @@ const makeRSVP = async ({ hostLabel, name, reservation, timeLimit }) => {
   const makeGuest = async (label) => {
     const acc = await stdlib.newTestAccount(sbal);
     acc.setDebugLabel(label);
+    if (stdlib.connector != 'ALGO') {
```
jeapostrophe, line 64: Ditto

examples/rsvp-6-vevt/index.mjs
```diff
@@ -2,9 +2,10 @@ import { loadStdlib, test } from '@reach-sh/stdlib';
 import * as backend from './build/index.main.mjs';
 
 // Basics
+const GAS_LIMIT = 5000000;
 const stdlib = loadStdlib({ REACH_NO_WARN: 'Y' });
 const err = {
-  'ETH': 'transaction may fail',
+  'ETH': 'transaction failed',
   'ALGO': 'assert failed',
   'CFX': 'transaction is reverted',
```
jeapostrophe, line 10: There's another error here:  https://output.circle-artifacts.com/output/job/97218047-c06c-4bcb-9a82-4d9ac415609f/artifacts/1/tmp/artifacts/CFX.rsvp-6-vevt.gz

jeapostrophe: Receipt outcomeStatus is nonzero

examples/rsvp-6-vevt/index.mjs
TheChronicMonster: I use the pattern to avoid a warning when connected to ALGO. But if cleaner codebase is more desirable, I can remove the condition.

docs/src/tut/rsvp/index.md
```diff
@@ -790,10 +790,13 @@ This is representative of a user interface that shows the Host each registration
...
 ```
 load: /examples/rsvp-6-vevt/index.mjs
-md5: 6c9ca7f7e8a9cb738005b61cabc1e8b0
-range: 57-67
+md5: 65e69ecdbcea0eab0d04838833ddea55
+range: 59-70
 ```
 
+We've added a gas limit using `{!js} stdlib.setGasLimit` for testing on Ethereum.
+This prevents `{!rsh} API`s from expending all of their tokens on gas fees.
```
jeapostrophe, line 798: This is not accurate. There is always a gas limit, even if we don't specify anything.

Normally the gas limit is computed by looking at the contract code, but it is not always accurate. What we're doing here is setting the "limit" to the largest possible amount of gas that could ever be spent. Thus, we're really _removing_ the limit by setting it to be "infinity".

TheChronicMonster: Got it!

jeapostrophe (approved):
This is fine once CI passes

PR #1486, opened 2022-10-25
Tuple.includes

willghatch:
This almost works, but in the new `tupleIncludes` example, the dynamic case tests fail with an error about the generated SMT code:

```
reachc: The compiler has encountered an internal error:

  Unexpected result from the SMT solver:
  Expected: success
  Result: (error "line 162 column 31: select requires 1 arguments, but was provided with 2 arguments" )
```

Dealing with these SMT errors is difficult, because the reported location doesn't match the generated `index.main.smt` file, and I'm not sure exactly where it comes from.  The text of the error itself ends up in the `index.main.smt` file, and is clearly referencing a different file.  What file is it?  (I assume it's something transient during the compilation process, perhaps.)  I'm confused by this because from what I can tell, `select` DOES take 2 arguments, and we always give it two arguments.  I haven't modified any SMT generation code with this PR.  I don't see any obvious issues, so I thought I'd ask for a tip on where to look next.

### DIFF
--- a/docs/src/changelog/index.md
+++ b/docs/src/changelog/index.md
@@ -6,6 +6,8 @@ Versions and changes-within-versions are listed in reverse-chronological order: 
 ## 0.1.12: 2022/09 - present
 
 @{verRC("0.1.12")}
++ 2022/10/25: Added `{!rsh} Tuple.includes`.
++ 2022/10/25: Added `{!rsh} BytesDyn` casting from fixed-length bytes.
 
 @{rcHead("0.1.12-rc.7")}
 + 2022/10/25: Added `{!js} manager` optional field for `{!js} stdlib.launchToken`
--- a/docs/src/changelog/index.md
+++ b/docs/src/changelog/index.md
@@ -6,10 +6,9 @@ Versions and changes-within-versions are listed in reverse-chronological order: 
 ## 0.1.12: 2022/09 - present
 
 @{verRC("0.1.12")}
-+ 2022/10/25: Added `{!rsh} Tuple.includes`.
-+ 2022/10/25: Added `{!rsh} BytesDyn` casting from fixed-length bytes.
 
 @{rcHead("0.1.12-rc.7")}
++ 2022/10/25: Added `{!rsh} Tuple.includes`.
 + 2022/10/25: Added `{!js} manager` optional field for `{!js} stdlib.launchToken`
 + 2022/10/25: Added `{!rsh} BytesDyn` casting from fixed-length bytes.
 
--- a/hs/src/Reach/Eval/Core.hs
+++ b/hs/src/Reach/Eval/Core.hs
@@ -2155,6 +2155,18 @@ evalPolyEq lvl x y =
       at <- withAt id
       return $ (lvl, SLV_Bool at v)
 
+evalOr :: SLVal -> SLVal -> App SLVal
+evalOr l r =
+  case (l, r) of
+    (SLV_Bool bAt lb, SLV_Bool _ rb) ->
+      return $ SLV_Bool bAt $ lb || rb
+    (SLV_DLVar {}, SLV_DLVar {}) ->
+      snd <$> evalPrimOp S_IF_THEN_ELSE [(Public, l), public $ SLV_Bool sb True, (Public, r)]
+    (SLV_Bool bAt True, _) -> return $ SLV_Bool bAt True
+    (SLV_Bool _ False, SLV_DLVar {}) -> return $ r
+    (SLV_DLVar _, SLV_Bool {}) -> evalOr r l
+    _ -> impossible $ "evalOr expecting SLV_Bool or SLV_DLVar"
+
 evalAnd :: SLVal -> SLVal -> App SLVal
 evalAnd l r =
   case (l, r) of
@@ -2166,6 +2178,14 @@ evalAnd l r =
     (SLV_Bool _ True, SLV_DLVar {}) -> return $ r
     (SLV_DLVar _, SLV_Bool {}) -> evalAnd r l
     _ -> impossible $ "evalAnd expecting SLV_Bool or SLV_DLVar"
+
+evalOrMap :: (SLVal -> App SLVal) -> [SLVal] -> App SLVal
+evalOrMap f = \case
+  [] -> return $ SLV_Bool sb False
+  h : t -> do
+    h' <- f h
+    t' <- evalOrMap f t
+    evalOr h' t'
 
 evalAndMap :: (SLVal -> App SLVal) -> [SLVal] -> App SLVal
 evalAndMap f = \case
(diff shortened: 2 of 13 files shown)


hs/src/Reach/Eval/Core.hs
```diff
@@ -2965,13 +2985,8 @@ evalPrim p sargs =
...
       case tup of
         SLV_Tuple _ vs -> do
-          let typeMatch v = do
-                mTA <- typeOfM v
-                case mTA of
-                  (Just (t, _a)) -> return $ t == nt
-                  _ -> return $ False
-          vsMatched <- filterM typeMatch vs
-          evalApplyVals' foldable_includes [(lvl, (SLV_Array at nt vsMatched)), (lvl, needle)]
+          v_ret <- evalOrMap (\v -> snd <$> evalPolyEq Public v needle) vs
```
jeapostrophe, line 2988: I want this to be the only case and get `vs` from `explodeTupleLike`

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
willghatch: IE you want to remove the working dynamic case and make this a static-only feature?

jeapostrophe: `explodeTupleLike` works exactly like your DLVar case: https://github.com/reach-sh/reach-lang/blob/82d3f7c58a444ebcf9f7a4be50fe8b41f38ffa6f/hs/src/Reach/Eval/Core.hs#L2441-L2442

willghatch: Oh, somehow I missed that.

hs/src/Reach/Eval/Core.hs
```diff
@@ -1551,11 +1552,23 @@ evalAsEnvM sv@(lvl, obj) = case obj of
...
+             -- I would just do `delayCall SLPrim_tuple_includes`, but it fails
+             -- if tested by `SLPrim_is` as a function, which means that something
+             -- like `mytuple.includes` can't be set as a view function.
+             -- Fixing the `SLPrim_is` implementation seems a lot harder than this.
+             foldable_includes <- lookStdlib "Foldable_includes"
+             return $ (lvl, jsClo (srclocOf tupSlv) "tuple_includes"
+               "(v) => f(tup, v)"
+               (M.fromList [("f", SLV_Prim SLPrim_tuple_includes),
+                            ("tup", tupSlv),
+                            ("Foldable_includes", foldable_includes)])))
```
jeapostrophe, line 1571: Not needed

hs/src/Reach/Eval/Core.hs
```diff
@@ -1551,11 +1552,23 @@ evalAsEnvM sv@(lvl, obj) = case obj of
...
       M.fromList
         [ ("set", delayCall SLPrim_tuple_set)
         , ("length", doCall SLPrim_tuple_length)
+        , ("includes",
+           do
+             -- I would just do `delayCall SLPrim_tuple_includes`, but it fails
+             -- if tested by `SLPrim_is` as a function, which means that something
+             -- like `mytuple.includes` can't be set as a view function.
+             -- Fixing the `SLPrim_is` implementation seems a lot harder than this.
+             foldable_includes <- lookStdlib "Foldable_includes"
```
jeapostrophe, line 1566: Ditto

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
willghatch: Can you be a bit more specific?  The comment explains why I didn't do the more obvious thing.  So I'm not sure what to do if this closure stuff is not needed.

jeapostrophe: The closure doesn't actually use `Foldable_includes` anymore

hs/src/Reach/Eval/Core.hs
jeapostrophe: Thus the binding isn't needed

hs/src/Reach/Eval/Core.hs
willghatch: Oh, I see now that your comment was about a single line... that makes more sense now...

willghatch: I thought the comment was about the whole block displayed there, which I now realize is mostly for context, but doesn't make it clear which line you are referring to.

willghatch:
Hmm, I find these binding failures due to using that closure form troubling.  I haven't looked at the inner workings of `JsClo`, but I would expect it not to affect the visibility of bindings on its arguments.  I'll look at this more later.  But if you have any helpful words of wisdom with respect to `JsClo`'s design and intentions that might be useful.

jeapostrophe:
It isn't really `jsClo` that is at fault. It is `lookupStdlib`. Basically... when a normal closure gets created, it grabs the environment, which includes the standard library, and some code (like `polyEq`) relies on using standard library functions (like `digest`). Since Reach doesn't allow shadowing ever, we know that those stdlib functions can't be overridden, so they will definitely refer to the ones we want.

jeapostrophe (changes requested):
You're so clever... this should be a variant of `jsClo` that just extends the env

willghatch:
Fair enough.  Sometimes I just can't resist making the minimum change.

willghatch:
I'm not sure in this case whether “clever” is a compliment or an insult.  This patch has a couple of hacks, but in both cases I'm not sure how to do “the right thing”.  IE making `lookStdlib` actually look at the stdlib env instead of the current env (because that is generated in at least Module.hs or something that would have some dependency issues) or making `SLPrim_is` return the right answer when given an `SLPrim_tuple_includes` that may need 1 or 2 args (I'm not sure how to tell at that point the difference between a value from `Tuple.includes` or `mytuple.includes`, one of which needs 2 args and one of which needs 1).  I'm happy to hear answers to those mysteries ;)

jeapostrophe (approved):

jeapostrophe:
It says there's a conflict?

willghatch:
Rebased.  It had a conflict in the changelog from my other PR that was recently merged.

docs/src/changelog/index.md
```diff
@@ -6,6 +6,8 @@ Versions and changes-within-versions are listed in reverse-chronological order:
 ## 0.1.12: 2022/09 - present
 
 @{verRC("0.1.12")}
++ 2022/10/25: Added `{!rsh} Tuple.includes`.
++ 2022/10/25: Added `{!rsh} BytesDyn` casting from fixed-length bytes.
```
jeapostrophe, line 10: Bad merge and in wrong spot

jeapostrophe (changes requested):

jeapostrophe (approved):

PR #1495, opened 2022-10-28
CORE-2643: Make `Contract.new` Accept `Reach.App`s

chrisnevers:
Perhaps one provocative implementation detail is when compiling a child program, we get the resulting `ConnectorObject` and do some tweaking to construct the options we need for `Contract.new`. As opposed to tweaking the `ConnectorObject` to output more fields, which are either slightly varied forms of existing fields, or providing the same value under multiple fields


hs/src/Reach/Connector/ALGO.hs
```diff
@@ -3671,6 +3671,7 @@ compile_algo env disp (CPProg {..}) = do
     code "b" [ "apiReturn_noCheck" ]
     label "alloc"
     let ctf f x = do
+          liftIO $ modifyIORef resr $ M.insert (LT.toStrict f) $ AS.Number $ fromIntegral x
```
jeapostrophe, line 3674: What's going on here?

hs/src/Reach/Eval.hs
```diff
@@ -351,6 +354,7 @@ makeEnv cns = do
   e_mape <- makeMapEnv
   e_droppedAsserts <- newCounter 0
   let e_appr = Left $ app_default_opts e_id e_droppedAsserts cns
+  let e_compileProg = const mempty
```
jeapostrophe, line 357: It should error with impossible

hs/src/Reach/Eval.hs
```diff
@@ -400,13 +404,12 @@ prepareDAppCompiles run shared_lifts exe_ex = do
               case sss_val v of
                 SLV_Prim SLPrim_App_Delay {} -> True
                 _ -> False
-  let go getdapp = run $ checkUnusedVars $ do
+  let go compileProg getdapp = run $ checkUnusedVars $ do
         exports <- getExports exe_ex
-        topv <- ensure_public . sss_sls =<< getdapp
```
jeapostrophe, line 405: I don't get why it is okay to remove these parts

hs/src/Reach/Eval.hs
```diff
@@ -400,13 +404,12 @@ prepareDAppCompiles run shared_lifts exe_ex = do
...
-        compileDApp shared_lifts exports topv
+        compileDApp shared_lifts exports compileProg getdapp
   case S.null tops of
     True -> do
-      return (S.singleton "default", const $ go $ return defaultApp)
+      sv <- liftIO mkDefaultApp
+      return (S.singleton "default", const . flip go sv)
     False -> do
-      let go' which = go $ env_lookup LC_CompilerRequired which exe_ex
```
jeapostrophe, line 411: And this one

hs/src/Reach/Eval/Core.hs
```diff
@@ -104,6 +108,7 @@ data Env = Env
   , e_appr :: Either DLOpts (IORef AppInitSt)
   , e_droppedAsserts :: Counter
   , e_infections :: IORef (M.Map Int (SrcLoc, SLVar))
+  , e_compileProg :: String -> SLVal -> IO ConnectorObject
```
jeapostrophe, line 111: Use the type alias

hs/src/Reach/Eval/Core.hs
```diff
@@ -4124,14 +4128,45 @@ evalPrim p sargs =
...
-      (ccv, opts) <-
+      let expectContractCode = \case
+            SLV_ContractCode _ cc -> return cc
+            ow -> expect_t ow $ Err_Expected "ContractCode or Reach.App"
+      (cc, opts) <-
         case args of
-          [x] -> return (x, mempty)
-          [x, y] -> (,) x <$> mustBeObject y
+          [x@(SLV_Prim (SLPrim_App_Delay _ _ _ _ mname))] -> do
+            -- Compile Reach.App and remember result of compilation
```
jeapostrophe, line 4137: Does the box get filled elsewhere?

hs/src/Reach/Eval/Core.hs
```diff
@@ -4124,14 +4128,45 @@ evalPrim p sargs =
...
+                goAlgoOpts = expectJsonObj $ \ o ->
+                            M.fromList [ ("globalUints", getUInt "GlobalNumUint" o)
+                                        , ("globalBytes", getUInt "GlobalNumByteSlice" o)
+                                        , ("localBytes", getUInt "LocalNumByteSlice" o)
+                                        , ("localUints", getUInt "LocalNumUint" o) ]
+            let goEth = expectJsonObj $ mapJsonString (T.drop 2) . get "Bytecode"
+            let co' = flip M.mapWithKey co $ \case
+                        "ALGO" -> goAlgo
+                        _  -> goEth
+            let opts'= M.fromList $ flip map (M.toList co) $ \ (k, v) -> case k of
```
jeapostrophe, line 4163: This is very scandalous and should at least be in ALGO and part of the connector interface like conGen

hs/src/Reach/Compiler.hs
```diff
@@ -71,337 +72,359 @@ compile env (CompilerOpts {..}) = do
...
+                --
+                -- You might think this only looks at the A & B pieces, but it also has
+                -- to look at the C piece, so we can generate client-only View
+                -- implementations, because Algorand doesn't have a good way to call
+                -- read-only methods of contracts.
+                loud $ "running backend js"
+                backend_js woutn crs $ plp_epp pl
+                return $ crs
+          m_compile_prog :: String -> SLVal -> IO ConnectorObject
+          m_compile_prog which = \case
```
jeapostrophe, line 414: I don't like how big this diff is and I feel like I would have done it by making another top level function.

It is weird that the thing called by the cli returns the object rather than that being a sub thing or the object just being set as an effect

It is really weird that this knows about SL values

hs/src/Reach/Connector/ALGO.hs
chrisnevers: This adds `Global(Bytes/UInt)/Local(Bytes/UInt)` to the `ConnectorObject` result, so the info is accessible when evaluating `new Contract`

hs/src/Reach/Eval/Core.hs
chrisnevers: This comment is stale; I'll remove it. The box only needs to be written in `Compiler::m_compile_prog`, this way if you compile a file with

```JavaScript
export const child = Reach.App(() => {});

  // ...
  const ctc = new Contract(child);
});
```

The compiler will compile `child` first (alphabetical order), remember the result since the box is written in `Compiler.hs` and will not recompile inside of `new Contract`.

hs/src/Reach/Compiler.hs
chrisnevers: I believe there needs to exist a compile function that takes the SLVal as opposed to the name in the top env, since it is possible to define a Reach.App in any scope. Is there a simpler way I'm not thinking about?

hs/src/Reach/Connector/ALGO.hs
jeapostrophe: Ah, I understand

hs/src/Reach/Compiler.hs
jeapostrophe: Inside of Compiler.hs, we basically are just waiting for a DLProg and then we can go from there. `compileDApp`, which is really part of `Eval` knows how to turn a name into a `DLProg`.

So, I think that `Compiler.hs` should be giving the `Eval` module the `DLProg -> IO ConnectorInfo` function and then Eval will deal with that. I basically imagine that `prepareDAppCompiles` will receive that function too and then the `compileDApp : String -> DLProg` will be used in a really trivial way in Compiler.hs

hs/src/Reach/Compiler.hs
```diff
@@ -71,337 +71,345 @@ compile env (CompilerOpts {..}) = do
...
+      (run, shared_lifts, exe_ex) <- evalBundle all_connectors djp co_printKeywordInfo
+      when co_printKeywordInfo $ do
+        let exportMap = M.map sss_val exe_ex
+        printBaseKeywordInfo exportMap
+        exitSuccess
+
+      let compile_prog onlyOne which dl = do
+            -- We are now inside the main part of the compiler that actually does
+            -- something for each DApp.
+            unless onlyOne $ do
```
jeapostrophe, line 98: This should just be part of the forM below

hs/src/Reach/Compiler.hs
```diff
@@ -71,337 +71,345 @@ compile env (CompilerOpts {..}) = do
...
+            let woutnMay = outnMay woutn
+            let showp :: Pretty a => T.Text -> a -> IO ()
+                showp l x = do
+                  let x' = pretty x
+                  let x'' = render x'
+                  loud $ "showp " <> show l
+                  interOut woutn l x''
+            let DLProg { dlp_opts = DLOpts {..} } = dl
+            showp "dl" dl
+            case co_stopAfterEval of
```
jeapostrophe, line 110: This should not be available in this function

hs/src/Reach/Compiler.hs
```diff
@@ -71,337 +71,345 @@ compile env (CompilerOpts {..}) = do
...
+            unless onlyOne $ do
+              putStrLn $ "Compiling " <> show which <> "..."
+            let woutn = co_output . ((T.pack which <> ".") <>)
+            let woutnMay = outnMay woutn
+            let showp :: Pretty a => T.Text -> a -> IO ()
+                showp l x = do
+                  let x' = pretty x
+                  let x'' = render x'
+                  loud $ "showp " <> show l
+                  interOut woutn l x''
```
jeapostrophe, line 107: These should be the one printing arg

jeapostrophe (changes requested):

hs/src/Reach/Eval/Core.hs
```diff
@@ -4124,14 +4130,44 @@ evalPrim p sargs =
...
+            SLV_ContractCode _ cc -> return cc
+            ow -> expect_t ow $ Err_Expected "ContractCode or Reach.App"
+      (cc, opts) <-
         case args of
-          [x] -> return (x, mempty)
-          [x, y] -> (,) x <$> mustBeObject y
+          [x@(SLV_Prim (SLPrim_App_Delay _ _ _ _ mname))] -> do
+            compileProg <- asks e_compileProg
+            name <- maybe (("Reach.App" <>) . show <$> ctxt_alloc) return mname
+            co <- compileProg name x
```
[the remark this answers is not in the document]
chrisnevers: I understand comment 2. I'm not sure I understand comment 3, do you mean something like this:

```
const name = case mname of 
  Just n -> originalWhich <> n 
  Nothing -> originalWhich <> ctr
```
?

examples/view-exns/index.txt
```diff
@@ -10,9 +10,3 @@ Verifying for generic connector
   Verifying when ALL participants are honest
   Verifying when NO participants are honest
 Checked 5 theorems; No failures!
-Compiling "main"...
```
jeapostrophe, line 13: Why did this get removed?

hs/src/Reach/Eval/Core.hs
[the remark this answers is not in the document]
jeapostrophe: I guess what I mean is that I want to see something like

"Compiling coolKid for main" when there's a name, but the files are still named `${file}.coolKid.mjs`

But when there's no name it would be

"Compiled internal contract #2 for main" and the file is `${file}.${originalWhich}${counter}.mjs`

examples/view-exns/index.txt
chrisnevers: `main = dapp_plus` which was already compiled, but we probably want to re-write the output files as `main` too

hs/src/Reach/Eval/Core.hs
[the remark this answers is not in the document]
chrisnevers: sounds good

hs/src/Reach/Eval/Core.hs
```diff
@@ -4131,10 +4131,10 @@ evalPrim p sargs =
       cns <- readDlo dlo_connectors
       (cc, con_x_opts) <-
         case args of
-          [x@(SLV_Prim (SLPrim_App_Delay _ _ _ _ mname))] -> do
+          [x@(SLV_Prim (SLPrim_App_Delay {}))] -> do
             compileProg <- asks e_compileProg
-            name <- maybe (("Reach.App" <>) . show <$> ctxt_alloc) return mname
-            co <- compileProg name x
+            -- name <- maybe (("Reach.App" <>) . show <$> ctxt_alloc) return mname
```
jeapostrophe, line 4136: rm

hs/t/y/child_reach_app.rsh
```diff
@@ -0,0 +1,49 @@
...
+  init();
+  A.publish();
+  const ctor1 = new Contract(t);
+  const ctor2 = new Contract(Reach.App(() => {
+    // How should scope be handled?
+    // We want to evaluate locally so that
+    // the state says we haven't initialized the app
+    // and we want to be in the correct state.
+    // but do we want to access outer values
+    // like closures can?
```
jeapostrophe, line 40: We should be able to access static things (like numbers) and not dynamic things (like A, or variables that are published). So, you could do...
```
A.publish(x);
const y = 5;
const z = x + y;
```
and demonstrate that `ctor2` can look at `y` but not `x` or `z`.

jeapostrophe: I think that you should change this test case so that `a` is `(y) => same-code-as-right-here` and do `a(y)`, then make it so that `t = a(5)`. Right now, we write lots of programs like `t`, but this is a new test to make sure we support things like `a(y)` and embedding.

jeapostrophe: As far as actually implementing it...

Option 1: Scrub the environment of things that aren't allowed
Pro: Simple
Con: You might miss something and you'll let the name be used again

Option 2: Add something to the environment that says what "universe" you are in and then error if you ever look at something from the wrong universe
Pro: Simple conceptually
Con: Big code change:
- Must change the `SLEnv` data type to add another field for the universe of each entry
- Add something to the overall env that has the current universe
- Must include a counter that gets incremented each time you start a new `Reach.App` scope.

I think you should do Option 2

jeapostrophe (commented):
I like this structure

hs/src/Reach/AST/SL.hs
```diff
@@ -177,6 +177,34 @@ data SLVal
...
+  SLV_Bool {} -> False
+  SLV_Int {} -> False
+  SLV_Bytes {} -> False
+  SLV_BytesDyn {} -> False
+  SLV_String {} -> False
+  SLV_DLC _ -> False
+  SLV_Type {} -> False
+  SLV_Connector {} -> False
+  SLV_Prim {} -> False
+  SLV_Form {} -> False
```
jeapostrophe, line 192: These can contain more values, so we have to disallow them

hs/src/Reach/AST/SL.hs
```diff
@@ -177,6 +177,34 @@ data SLVal
...
+  SLV_Null {} -> False
+  SLV_Bool {} -> False
+  SLV_Int {} -> False
+  SLV_Bytes {} -> False
+  SLV_BytesDyn {} -> False
+  SLV_String {} -> False
+  SLV_DLC _ -> False
+  SLV_Type {} -> False
+  SLV_Connector {} -> False
+  SLV_Prim {} -> False
```
jeapostrophe, line 191: Same with these

hs/src/Reach/Eval.hs
```diff
@@ -353,6 +353,7 @@ makeEnv cns = do
   e_droppedAsserts <- newCounter 0
   let e_appr = Left $ app_default_opts e_id e_droppedAsserts cns
   let e_compileProg = const $ impossible "compileProg"
+  e_universe <- newCounter uni
```
jeapostrophe, line 356: It is important that this is a global counter. I think that `makeEnv` should take a `uniC :: Counter` and you should do `e_universe <- readCounter uniC`

jeapostrophe (changes requested):

hs/src/Reach/AST/SL.hs
chrisnevers: Currently, blanket disallowing form and prim will make it so no primitives etc can not be used in nested apps because they are in different universes. I think this can be handled by making `isDynamicPrim` etc. and searching within each prim for dynamic values or changing `sss_universe` to be a `Maybe` and have primitive/form values be `Nothing` to distinguish them. Any preference?

jeapostrophe: I think recurring will have to be the way to do it.

jeapostrophe:
I rebased and resolved the conflicts for you

jeapostrophe (commented):
looked at this version

jeapostrophe (approved):

PR #1500, opened 2022-11-01
invariant example

nstanford5:
This example tracks a large supply of non-network tokens in parallelReduce consisting of a more advanced invariant example.


examples/ticket-sales/index.rsh
```diff
@@ -0,0 +1,40 @@
...
+    .while(ticketsSold < supply)
+    .api_(B.buyTicket, () => {
+      check(ticketsSold != supply, "sorry, out of tickets");
+      return[cost, (ret) => {
+        transfer(1, tok).to(this);
+        ret(null);
+        return [ticketsSold + 1];
+      }];
+    });
+  transfer(balance()).to(A);
```
jeapostrophe, line 37: It is bad style to use `balance()` at runtime. You should use `cost * ticketsSold`

jeapostrophe (changes requested):

examples/ticket-sales/index.mjs
```diff
@@ -0,0 +1,38 @@
...
+await Promise.all([
+  backend.Admin(ctcA, {
+    params: {
+      tok: tickets.id,
+      cost: stdlib.parseCurrency(10),
+      supply: MAX,
+    },
+    launched: (contract) => {
+      console.log(`Ready at contract: ${contract}`);
+      startBuyers();
```
jeapostrophe, line 34:
```suggestion
    launched: async (contract) => {
      console.log(`Ready at contract: ${contract}`);
      await startBuyers();
```

jeapostrophe: But, I would just stick the code right here

examples/ticket-sales/index.mjs
```diff
@@ -0,0 +1,38 @@
...
+const MAX = 50;
+const accA = await stdlib.newTestAccount(stdlib.parseCurrency(5000));
+const ctcA = accA.contract(backend);
+const tickets = await stdlib.launchToken(accA, "Tickets", "TIX", {supply: MAX});
+
+console.log('Welcome to the ticket distributor\nLets get you a ticket');
+
+const startBuyers = async () => {
+  await stdlib.wait(10);// need this to work on ETH
+  const runBuyers = async (i) => {
```
jeapostrophe, line 13:
```suggestion
  const runBuyer = async (i) => {
```

examples/ticket-sales/index.mjs
```diff
@@ -0,0 +1,38 @@
...
+    const acc = await stdlib.newTestAccount(stdlib.parseCurrency(100));
+    const ctc = acc.contract(backend, ctcA.getInfo());
+    await acc.tokenAccept(tickets.id);
+    await ctc.apis.Buyer.buyTicket();
+    console.log(`Tickets sold: ${i}`);
+  }// end of runBuyers
+  for(let i = 0; i < MAX; i++){
+    await runBuyers(i);
```
jeapostrophe, line 21:
```suggestion
    await runBuyer(i);
```

This runs them one at a time. If you want them in parallel, then you'd use `Promise.all`

examples/ticket-sales/index.mjs
```diff
@@ -0,0 +1,38 @@
...
+const stdlib = loadStdlib({REACH_NO_WARN: 'Y'});
+const MAX = 50;
+const accA = await stdlib.newTestAccount(stdlib.parseCurrency(5000));
+const ctcA = accA.contract(backend);
+const tickets = await stdlib.launchToken(accA, "Tickets", "TIX", {supply: MAX});
+
+console.log('Welcome to the ticket distributor\nLets get you a ticket');
+
+const startBuyers = async () => {
+  await stdlib.wait(10);// need this to work on ETH
```
jeapostrophe, line 12: I bet that is because the `interact.ready` was in the wrong spot

jeapostrophe (changes requested):

examples/ticket-sales/index.rsh
```diff
@@ -0,0 +1,40 @@
...
+    buyTicket: Fun([], Null),
+  });
+  init();
+
+  A.only(() => {
+    const {cost, tok, supply} = declassify(interact.params);
+  });
+  A.publish(cost, tok, supply);
+  commit();
+  A.interact.launched(getContract());
```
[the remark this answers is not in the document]
nstanford5: Curious why it doesn't work here?

The first `publish` launches the contract and after the `commit` I assume its safe to call `getContract`

Does the introduction of the second consensus operation `pay` just continue the first consensus step? If this is the case, does the `commit` here just give us access to `pay` the `tok` and use `supply` value, while not moving us back to "step"?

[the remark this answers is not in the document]
jeapostrophe: It is safe to call. The problem is that you intend this to mean "you can start making API calls". That is only possible once you're in the consensus step of the parallelreduce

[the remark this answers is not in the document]
nstanford5: Got it -- thank you

examples/ticket-sales/index.txt
```diff
@@ -0,0 +1,5 @@
+Verifying knowledge assertions
+Verifying for generic connector
+  Verifying when ALL participants are honest
+  Verifying when NO participants are honest
+Checked 27 theorems; No failures!
```
jeapostrophe, line 5: I think you copy and pasted this file rather than using the gen-test-output script

jeapostrophe:
Isn't the same example as `point-of-sale`?

examples/ticket-sales/index.txt
```diff
@@ -0,0 +1,5 @@
+Verifying knowledge assertions
```
jeapostrophe, line 1:
```suggestion
Compiling `main`...
Verifying knowledge assertions
```

jeapostrophe (changes requested):

PR #1508, opened 2022-11-06
Add Algorand Wallet Fallback support for AlgoSigner and Exodus Web Wallet

xarmian:
## Summary

This is a wallet fallback wrapper function to be used with the AlgoSigner global exposed by the AlgoSigner browser extension. It works similarly to MakePeraConnect and allows AlgoSigner to be used as a wallet fallback.

This was also created to add support for the Exodus Web Wallet, a newer browser extension wallet for Chrome/Brave. Exodus Web Wallet exposes an AlgoSigner object for compatibility with existing dApps, and only supports MainNet.

Some of the specifications for this PR were:
1. Add compatibility for both AlgoSigner and Exodus Web Wallet browser extensions
2. Work with existing wallerFallback function without modifying the core standard library
3. Require implementer to specify AlgoSigner global to help avoid potential future compatibility issues

The same functionality could be implemented with slight modifications to ALGO.ts, however I went this route to maintain consistency with existing walletFallback routes.

It works the same way as the MakePeraConnect function by exporting a function that returns a class that can be used with `stdlib.walletFallback()`. The first argument to the function is the AlgoSigner global exposed by the AlgoSigner browser extension, and the second argument is the network (TestNet/MainNet) which is optional and defaults to TestNet.

The object returned from this function is used in place of MyAlgoConnect as follows:

```
import { ALGO_MakeAlgoSignerConnect as MakeAlgoSignerConnect } from '@reach-sh/stdlib';
const wf = stdlib.walletFallback({providerEnv: 'TestNet', MyAlgoConnect: MakeAlgoSignerConnect(AlgoSigner)});
```

or

```
import { ALGO_MakeAlgoSignerConnect as MakeAlgoSignerConnect } from '@reach-sh/stdlib';
const wf = stdlib.walletFallback({providerEnv: 'MainNet', MyAlgoConnect: MakeAlgoSignerConnect(AlgoSigner,'MainNet')});
```

This PR includes an update to [examples/algo-wallet-demo](/reach-sh/reach-lang/tree/master/examples/algo-wallet-demo) which demonstrates the walletFallback `connect()` function. Transaction signing was tested using API transactions with my own contracts deployed on TestNet and MainNet, but may require further testing for other potential scenarios (deploy contract, transfer, etc)

Yes -- in the walletFallback section of the Frontend category [here](https://docs.reach.sh/frontend/#p_33), and I imagine a blurb in the changelog

### DIFF
--- a/examples/algo-wallet-demo/index.js
+++ b/examples/algo-wallet-demo/index.js
@@ -7,6 +7,7 @@ import { loadStdlib } from '@reach-sh/stdlib';
 import { ALGO_MyAlgoConnect as MyAlgoConnect } from '@reach-sh/stdlib';
 import { ALGO_WalletConnect as WalletConnect } from '@reach-sh/stdlib';
 import { ALGO_MakePeraConnect as MakePeraConnect } from '@reach-sh/stdlib';
+import { ALGO_MakeAlgoSignerConnect as MakeAlgoSignerConnect } from '@reach-sh/stdlib';
 import { PeraWalletConnect } from "@perawallet/connect";
 
 // This program is intended to be used on the Algorand TestNet.
@@ -101,6 +102,15 @@ class App extends React.Component {
     this.appendMsg('Using PeraConnect');
   }
 
+  async useAlgoSignerConnect() {
+    delete window.algorand;
+    stdlib = loadStdlib(process.env);
+    stdlib.setWalletFallback(reach.walletFallback({
+      // ...we use a different fallback here:
+      providerEnv, MyAlgoConnect: MakeAlgoSignerConnect(AlgoSigner,'TestNet') }));
+    this.appendMsg('Using AlgoSignerConnect');
+  }
+
   // Here we save and restore WalletConnect sessions
   async saveWalletConnect() {
     walletConnect_session = window.algorand.wc.wc;
@@ -150,6 +160,8 @@ class App extends React.Component {
           >Use WalletConnect</button>
           <button onClick={() => parent.usePeraConnect()}
           >Use PeraConnect</button>
+          <button onClick={() => parent.useAlgoSignerConnect()}
+          >Use AlgoSignerConnect</button>
           <button onClick={() => parent.attachWallet()}
           >Attach Wallet</button>
           <button onClick={() => parent.queryAccount()}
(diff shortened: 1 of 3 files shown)


examples/algo-wallet-demo/index.js
```diff
@@ -101,6 +102,15 @@ class App extends React.Component {
     this.appendMsg('Using PeraConnect');
   }
 
+  async useAlgoSignerConnect() {
+    delete window.algorand;
+    stdlib = loadStdlib(process.env);
+    stdlib.setWalletFallback(reach.walletFallback({
+      // ...we use a different fallback here:
+      providerEnv, MyAlgoConnect: MakeAlgoSignerConnect(AlgoSigner,'TestNet') }));
```
jeapostrophe, line 110: I am ignorant... will this error during loading on browsers that don't have AlgoSigner, or will it not because it won't actually evaluate this line?

xarmian: Those lines wouldn't be evaluated during page load, but it would trigger an error if a user clicks the "Use AlgoSignerConnect" button and there is no AlgoSigner object.

If we want to ensure that walletFallback doesn't fail (and leave the failing to a function like .getDefautAccount) we could have the AlgoSignerConnect class assume the presence of the AlgoSigner object without passing it to the `MakeAlgoSignerConnect` function as an argument.

So either this:

```
export default function ALGO_MakeAlgoSignerConnect( provider: string = 'TestNet' ) {
    return class AlgoSignerConnect {
        constructor() {
        }

        async connect() {
            //@ts-ignore
            await AlgoSigner.connect();
            //@ts-ignore
            return await AlgoSigner.accounts({ledger:provider});
        }
// ...
```

Or using the class' constructor to grab the AlgoSigner object instead, i.e.:

```
export default function ALGO_MakeAlgoSignerConnect( provider: string = 'TestNet' ) {
    return class AlgoSignerConnect {
        AlgoSigner: any = null;
        
        constructor() {
            //@ts-ignore
            if (typeof AlgoSigner != 'undefined') this.AlgoSigner = AlgoSigner;
        }
// ... and change all other references from AlgoSigner to this.AlgoSigner
```

In either case, walletFallback would change to the following:

```
stdlib.setWalletFallback(reach.walletFallback({
      providerEnv, MyAlgoConnect: MakeAlgoSignerConnect('TestNet') }));
```

Let me know if that's a preferable approach.. or if you think a different angle would be better.

jeapostrophe:
I'm rebasing, then I'm going to add the changelog/docs changes and merge it.

PR #1517, opened 2022-11-07
point-of-sale example

nstanford5:
This is a basic point-of-sale DApp with `purchase` and `refund` functions in a `parallelReduce`. This examples covers: tracking a supply of non-network tokens, taking in varying amount of network tokens, more advanced invariants/checks, refunding based on varied purchase amount, `paySpec`

### DIFF
--- a/examples/point-of-sale/index.mjs
+++ b/examples/point-of-sale/index.mjs
@@ -12,20 +12,24 @@ const loyalTok = await stdlib.launchToken(accA, "Loyalty", "LYL", {supply: USERS
 
 console.log(`Welcome to the point-of-sale machine. This machine processes
 payments of varying amounts and rewards each purchase with a loyalty token.
-The pos will also process refunds, which return non-network tokens to the customer,
+The pos will also process refunds, which return network tokens to the customer,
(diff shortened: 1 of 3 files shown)


chrisnevers (commented):
LGTM, just a couple of nit picks

chrisnevers (dismissed):

examples/point-of-sale/index.mjs
```diff
@@ -0,0 +1,59 @@
...
+
+const startBuyers = async () => {
+  const runBuyer = async (i) => {
+    const acc = await stdlib.newTestAccount(stdlib.parseCurrency(100));
+    const ctc = acc.contract(backend, ctcA.getInfo());
+    const cost = getPurchasePrice(i);
+    await acc.tokenAccept(loyalTok.id);
+    try{
+      await ctc.apis.Buyer.purchase(cost);
+      txns++;
```
jeapostrophe, line 31: This is not atomic in JS and it is possible to get the same number multiple times. You could have the contract keep track of a counter and return it, then it would be accurate.

jeapostrophe (approved):

PR #1556, opened 2023-02-10
add arc-72 example

willghatch:
This is a pretty generic ARC-72 implementation to serve as an example.


jeapostrophe (approved):
This looks good. I re-run the Circle job, because it died for a spurious reason

PR #1559, opened 2023-02-17
add nextUpToTime to event stream interface

willghatch:
As discussed previously, this allows for users to get all events up to a particular time and batch them for purposes like atomic database transactions that keep the database consistently in sync with a particular block time.  Because typical contracts with events may emit zero to many events per block, the `next` API does not give users a clean way to see if there are more events for a block time (including the latest) without waiting potentially indefinitely for a future event that may never come.


jeapostrophe (commented):
This looks good, but the CI doesn't pass.

jeapostrophe (approved):

Issue #1560, opened 2023-02-18
Unable to import stdlib in Vite


Gungy2:
**Describe the error**

If I try importing anything from `@reach-sh/stdlib` in a Vite browser project, I get the following error: 
```
ReferenceError: process is not defined
    at node_modules/util/util.js (util.js:109:1)
    at __require2 (chunk-JZQ37OGZ.js?v=63656b75:18:50)
    at shared_impl.ts:12:18
```
`stdlib` appears to be using the Node API, which should not be the case, since it is meant to be run on clients.

**To Reproduce**

Import `stdlib` from Vite in the browser.

**Expected behavior**

I expect to get no errors.

**Extra information**

I am using version 0.1.13 of Reach.
Thanks for the help!:)

jeapostrophe:
You should set up your vite to use fallbacks, like this:

https://github.com/reach-sh/reach-lang/blob/master/js/stdlib/webpack.config.js#L20-L26

Issue #1316, opened 2022-08-04
Vitejs (build tool) incompatibility (polyfill issue)


leakim34:
**Describe the error**

Vitejs cannot be used with reach stdlib, since it uses some node built in modules that are not polyfilled.
You'll start having multiple errors on each built in module used by reach-sh/stdlib.
Starting with one of them below:

```
DependencyNotFoundError
Could not find dependency: 'http2' relative to '/node_modules/@reach-sh/stdlib/dist/esm/rpc_server.js'
```
It will do this for each built in modules used by reach. 

**To Reproduce**

Setup a new environment with vite js using any framework you want (svelte/vue/react/..) and install reach stdlib dep.

```
npm create vite@latest my-vue-app -- --template vue
npm i @reach-sh/stdlib url-parse readline
npm run dev 
```
Import reach-sh/stdlib and instantiate it.

```
import { loadStdlib } from "@reach-sh/stdlib";
import MyAlgoConnect from "@reach-sh/stdlib/ALGO_MyAlgoConnect";
const reach = loadStdlib(process.env);
```
OR 

[Codesandbox test](https://codesandbox.io/s/nifty-wind-bcyv4t)

**Expected behavior**

Being able to instantiate the Stdlib and use reach on vitejs. 
Many libraries/frameworks are using vitejs lately for its speed and ease of use instead of the heavy webpack solution.

**Extra information**

It would help and bring many developers that use vitejs lately, it's easy to setup & run in few seconds ! 
Vitejs does not plan to polyfill this in the future since it will decrease their performance and seems to be better way to handle built in module nowadays.

CryptoRUSHGav:
I second this request. Vite is getting harder to bypass and several frameworks (VueJS, Svelte come to mind) have started to bundle it as default.

jeapostrophe:
We will work on this.

I am totally ignorant about these things, so if either of you know, please help.

My assumption #1: People that use Vite know that it is "weird" with respect to dependencies, because it changes when and where things are used.

My assumption #2: Given that, they are used to "configuring Vite to deal with it". I put that in quotes, because I have no idea how that is actually done.

I am working with analogical-thinking that Vite is like Webpack and you need to manually rewrite things --- https://github.com/reach-sh/reach-lang/blob/master/js/stdlib/webpack.config.js

Is that correct? If so, does that mean that we/Reach tell you/leakim34 what to change in your Vite configuration?

[...]

Issue #152, opened 2021-05-25
Webpack 5: Unable to import @reach-sh/stdlib


victorsh:
The error seems to be cause by Webpack 5 not being able to resolve the imports within a .mjs file.

Related issue: https://github.com/webpack/webpack/issues/11467

Attempted suggested solution with no success: https://github.com/webpack/webpack/issues/11467#issuecomment-691982478

Webpack Config File
```
/* eslint-disable */
const path = require('path')
const HtmlWebpackPlugin = require('html-webpack-plugin')
const MiniCssExtractPlugin = require('mini-css-extract-plugin')
require('dotenv').config()

module.exports = {
  mode: 'development',
  entry: {
    index: './src/index.js',
  },
  output: {
    filename: '[name].bundle.js',
    path: path.resolve(__dirname, 'dist'),
    clean: true,
    publicPath: '/'
  },
  devServer: {
    port: process.env.DEV_PORT,
    watchContentBase: true
  },
  resolve: {
    extensions: ['.js', '.ts', '.mjs', '.json', '.wasm']
  },
  module: {
    rules: [
      {
        test: /\.(js|jsx)$/,
        exclude: /node_modules/,
        use: ['babel-loader', 'eslint-loader']
      },
      {
        test: /\.m?js/,
        resolve: {
          fullySpecified: false
        }
      },
      {
        test: /\.css$/,
        use: [MiniCssExtractPlugin.loader, 'css-loader']
      },
      {
        test: /\.(png|svg|jpg|jpeg|gif)$/i,
        type: 'asset/resource',
      },
    ]
  },
  plugins: [
    new HtmlWebpackPlugin({ template: './src/index.html' }),
    new MiniCssExtractPlugin()
  ],
}
```

Error message:
```
vsh in ~/Documents/github/crypto/algorand/reach/battleship/client(ruby-3.0.0) on master λ npm run nodeserve

> client@1.0.0 nodeserve
> node server.js

Battleship client development server running on port 3000...
✖ ｢wdm｣: assets by chunk 3.39 MiB (name: index)
  asset index.bundle.js 3.39 MiB [emitted] (name: index)
  asset index.css 21 bytes [emitted] (name: index)
asset 10fe4030c4f3a48a7c95.png 6.42 KiB [emitted] [immutable] [from: src/assets/battleship.png] (auxiliary name: index)
asset index.html 376 bytes [emitted]
[...]

PR #1561, opened 2023-02-20
Add `exports` field to package.json in stdlib

Gungy2:
By adding a separate entrypoint `/browser` for the library, we make it much easier for the users to import the library in a browser, like so: `@reach-sh/stdlib/browser`.

## Summary
This would address #152, #1316 and #1560, by using the library like so: `import { ... } from "@reach-sh/stdlib/browser`. In the current implementation there is the field "browser" that points to the right entrypoint, but that is not usually supported by default in most bundlers. Moreover, it does not work at all if used with Typescript, since the right types are not exposed.

Probably, documentation will need to be updated, I am also willing to do that if pointed in the right direction. Thank you  for taking the time to review this!

### DIFF
--- a/docs/src/frontend/index.md
+++ b/docs/src/frontend/index.md
@@ -32,6 +32,13 @@ $ npm install @reach-sh/stdlib
 
 You only need to install the package directly if you are running your frontend without `reach` or using a tool like [webpack](https://webpack.js.org/) for deployment.
 
+If you want to install the standard library for a browser application or library and you are using a bundler **without** _Node polyfills_ (e.g. [Webpack 5](https://webpack.js.org/), [Vite](https://vitejs.dev/)), you can import the library by writing:
+
+```js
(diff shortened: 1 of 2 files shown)


jeapostrophe:
Thanks. I'd add the docs here:

https://github.com/reach-sh/reach-lang/blob/master/docs/src/frontend/index.md?plain=1#L23

In something like

```
:::note
If you have problems with JavaScript bundlers in your browser, you can instead require from `@reach-sh/stdlib/browser`.
:::

jeapostrophe:
I don't really have a good way of testing... I'm assuming it is working for you?

Gungy2:
> I don't really have a good way of testing... I'm assuming it is working for you?

Yes, I tested it locally. It is as easy as creating a skeleton Webpack 5 or Vite project, importing the library and logging the returned object. If everything is correct, it should build with no issues.

jeapostrophe:
Thank you so much!

PR #1585, opened 2023-05-12
Fix isqrt function

Gungy2:
## Summary

The `isqrt` function seems to be broken. It throws an exception for n = 0 and n = 1 and just hangs for n = 3. I changed it according with the Wikipedia implementation


jeapostrophe (approved):

jeapostrophe:
Thanks!

Issue #1586, opened 2023-06-01
Sqrt fails formal verification


Gungy2:
**Describe the error**

There seems to be a bug in sqrt for formal verification. The following example fails verification (irrespective of the value of `tokensToReturn`.

**To Reproduce**
```js
const tokensToReturn = (depositInvariant * stableTokenBalance) / slbTokenBalance;
require(sqrt(tokensToReturn) * sqrt(tokensToReturn) <= tokensToReturn, "Sqrt works");
```

**Extra information**

I am using the latest version of Reach. Also, may I ask you to prioritize this, if at all possible? My Master thesis is due in about 2 weeks, and my project has been blocked by this bug for about 3 weeks now. Thank you so much!

Gungy2:
![image](https://github.com/reach-sh/reach-lang/assets/64141526/2b371fb4-b1f8-4e80-b357-4cc767b2178d)

Gungy2:
I think the [conversation on Discord](https://discord.com/channels/628402598663290882/1106553538148057150) is also relevant.

Gungy2:
Alternatively, even easier:
```js
export const main =
  Reach.App( () => {
    const A = Participant('A', { x: UInt });
    init();
      A.only(() => {
        const x = declassify(interact.x);
        assume(x > 0, "x too low");
      });
      A.publish(x);
      commit();
      assert(sqrt(x) * sqrt(x) <= x);
    }
  );
```
This gives:
```
  // Violation Witness

  const UInt.max = 4;

  const x/43 = "A".interact.x;
  //    ^ could = 3
  //      from: ./t/y/sqrt.rsh:5:34:property binding

  // Theorem Formalization

  const v51 = sqrtx/43;
  //    ^ would be 2
  const v54 = (v51 * v51) <= x/43;
  //    ^ would be false
  assert(v54);

  Verifying when NO participants are honest
Checked 5 theorems; 2 failures (and 1 omitted repeats) :'(
```

jeapostrophe:
If you are not able to test or use the new SMT assertion, then you can change your code from using a `require` to using an `enforce`, which will make it happen at run-time.

PR #1587, opened 2023-06-06
Add additional assertion to UInt_sqrt

Gungy2:
## Summary
This PR fixes issue #1586 by adding an assertion for the `UInt_sqrt` function.

### DIFF
--- a/hs/smt2/runtime.smt2
+++ b/hs/smt2/runtime.smt2
@@ -15,6 +15,19 @@
 (declare-fun UInt_bior (UInt UInt) UInt)
 (declare-fun UInt_bxor (UInt UInt) UInt)
 (declare-fun UInt_sqrt (UInt) UInt)
+(assert
+  (forall ((x UInt))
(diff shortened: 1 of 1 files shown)


jeapostrophe:
Were you able to test this? Does it make your code work?

Gungy2:
Yes, it makes my code work. The assertion should only work for the right integer square root, if I am not mistaken.

Gungy2:
The only issue with this assertion is that, while it is _perfect_ (meaning that it passes just for the exact number) it tends to overwhelm the SMT solver (for my program, which is relatively complex, it times out multiple times). So I am not sure what is the right approach here to strike the balance between correctness (usability) and performance. Alternatively, to get around this, `enforce` can be used, as pointed out in #1586.

jeapostrophe:
Yea, that matches my priors too. One of the reasons Reach verification can be so fast is that we never use quantification so the theory is really small.

PR #1612, opened 2026-07-23
Fix/no docker deps for local mode

KBryan:


jeapostrophe (changes requested):
This is too big of a change to evaluate, because it does a huge reformatting.

